PR #1, opened 2019-05-09
Node/yarn app for code generation

jkaster:
- reads `looker.ini`for config
- fetches versioned API spec
- upgrades Swagger 2 to OpenApi 3
- lint checks the upgrade
- generates the configured languages to `api\version\language`

PR #2, opened 2019-05-14
Node version that generates multiple API version and languages

lukebowerman:
Both 3.0 and 3.1 APIs bindings can be generated at the same time.
Add lint step, better sync output
Fetch API spec, upgrade, lint, generate
Add TSConfig
Added license statements to source files


jkaster:
This is the cleaned up PR for #1

PR #8, opened 2019-06-20
WIP type resolver

dthorpe:
Still very much WIP, but type resolution is falling into place - Type refs / named types, intrinsic types, and array types are now handled. No linkage to method generation code yet.

Take a look at IProperty / Property class as a pattern for simplifying the swarm of OAS optional bits - if/when we can safely/sanely pick a default value for optional artifacts, we can internalize that with property getters so that consumption in code formatters doesn't need to care.

There may still be situations where knowing something is absent is significant, but I can't think of one at the moment.

I'm not entirely satisfied with how the IType.elementType turned out. It feels dirty to have that polluting the IType base info. I'd prefer a type guard function to allow narrowing IType to IArrayType, and elementType would only exist on IArrayType, but it doesn't look like it's possible to declare a type guard function as an instance method returning this. I'll keep fiddling with it. if (type.isArray) { type.elementType.name....}

Intrinsic types (string, integer, float, etc) are now supported, predefined as empty types. I'm assuming code formatters will recognize intrinsics by their name (and schema == null) and do some sort of mapping to output whatever the language-specific equivalent name or syntax is for each intrinsic. Same for array types. For example, int64 should be expressed in C# as a long (Int64 does exist in .NET as a type, but it's a wrapper class around long. Best to use long)

### DIFF
--- a/codeFormatter.ts
+++ b/codeFormatter.ts
@@ -1,0 +1,74 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2019 Looker Data Sciences, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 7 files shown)


jkaster:
Thanks for the comments. I do have a language/platform specific type mapper prototype that can be adopted for intrinsic type conversion.

dthorpe:
up next: merge existing params stuff into ApiModel.methods

PR #9, opened 2019-07-24
Start typing transport.py based on transport.ts

joeldodge79:
This is just the beginning of an exploration into python3 typing for
the transport "interface". I found this PEP interesting:
https://www.python.org/dev/peps/pep-3119/#abcs-vs-interfaces
not sure if abc is really what we want but I'd rather not use
zope.interface (I think we should try to keep dependancies small).

I've added a "configure" method to the transport interface as a low cost
implementation of configuring dependancy injection.

TODO:
- determine if we use dataclasses versus attrs
- flesh out transport.py interface/types
- actually implement and test requests_transport.py

### DIFF
--- a/python/tests/rtl/test_requests_transport.py
+++ b/python/tests/rtl/test_requests_transport.py
@@ -1,0 +1,15 @@
+"""Test the requests transport.
+"""
+
+from looker.rtl import transport as tp
+from looker.rtl import requests_transport as rtp
+
+
+def test_request_minimal():
+    """Test basic successful round trip
+    """
+    settings = tp.TransportSettings(base_url='/some/path', api_version='3.1')
+    test = rtp.RequestsTransport.configure(settings)
+    resp = test.request(tp.HttpMethod.GET, '/some/path')
+    assert isinstance(resp, tp.SDKSuccessResponse)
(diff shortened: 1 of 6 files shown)


joeldodge79:
This is still a work in progress. I spent time talking with @josephaxisa and kalen and the proposal is to continue using typing where it makes sense (e.g. for pluggable transport for sure, but might not be necessary in other parts of the python rtl).

python/tests/rtl/test_requests_transport.py
```diff
@@ -0,0 +1,15 @@
...
+
+from looker.rtl import transport as tp
+from looker.rtl import requests_transport as rtp
+
+
+def test_request_minimal():
+    """Test basic successful round trip
+    """
+    settings = tp.TransportSettings(base_url='/some/path', api_version='3.1')
+    test = rtp.RequestsTransport.configure(settings)
```
joeldodge79: Next step for this test is to create a fake "requests" module (perhaps a pytest fixture) and then instantiate `RequestsTransport` directly passing in the fake:

```
test = rtp.RequestsTransport(settings, fake_requests)
```

PR #11, opened 2019-07-26
Improve UserSession

josephaxisa:
A few changes to `user_session.py` following Joel's feedback as well as some refactoring due to other RTL changes

### DIFF
--- a/python/looker/rtl/user_session.py
+++ b/python/looker/rtl/user_session.py
@@ -1,46 +1,37 @@
 import abc
-from typing import Optional, Union
+from typing import Optional, Union, Dict
+
 from looker.rtl import requests_transport as rtp
 from looker.rtl import auth_token as at
 from looker.rtl import transport as tp
-
-
-# TODO: confirm about name mangling 
+from looker.rtl import sdk_error as se
 
 class BaseUserSession(abc.ABC):
     """UserSession base class"""
-
     @property
     @abc.abstractmethod
     def is_impersonating(self) -> bool:
-        raise NotImplementedError
-
+        """Determines if user is impersonating another user"""
     @property
     @abc.abstractmethod
     def is_authenticated(self) -> bool:
-        raise NotImplementedError
-
+        """Determines if current token is active"""
     @abc.abstractmethod
     def get_token(self) -> at.AuthToken:
-        raise NotImplementedError
-
+        """Returns an active token"""
     @abc.abstractmethod
     def login(self, user_id: Optional[str] = None) -> at.AuthToken:
-        raise NotImplementedError
-
+        """Logs in a user or allows an already logged in user to sudo as another user"""
     @abc.abstractmethod
-    def authenticate(self):
-        # TODO: still to be built based on request
-        raise NotImplementedError
-
+    def authenticate(self) -> Dict[str, str]:
+        """Create a dictionary with the current token to be used as a header"""
     @abc.abstractmethod
     def logout(self) -> bool:
-        raise NotImplementedError
+        """Logs out current user"""
 
 
 class UserSession(BaseUserSession):
-
-    def __init__(self, settings, transport: Optional[transport.Transport]):
+    def __init__(self, settings, transport: Optional[tp.Transport]):
         self._token: at.AuthToken = at.AuthToken()
         self.user_id: str = ''
         self.settings = settings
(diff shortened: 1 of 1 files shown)


python/looker/rtl/user_session.py
```diff
@@ -61,48 +52,58 @@ def get_token(self):
...
+    def _logout(self) -> tp.Response.value:
         token = self._token
-        # TODO: this is still to be built based on authenticator
+        result = self._ok(
+            self.transport.request(
+                tp.HttpMethod.DELETE,
+                '/logout',
+                authenticator=(lambda: {
+                    'Authorization': f'token {token.access_token}'
+                }) if token else None))
```
joeldodge79: this is a little tricky to read ... can the `lambda` here return `None`? Looks like `_logout` is gated by an `is_authenticated` check so we should be sure of having a `token`. So think we can get rid of the ternary here. Also, I believe lambda's in python are more like arrow functions in js and so `self` references the enclosing instance (e.g. I think you can do ...`f'token {self._token.access_token}` and avoid L92)

python/looker/rtl/user_session.py
```diff
@@ -61,48 +52,58 @@ def get_token(self):
...
+                                   }))
         return result
 
     def logout(self) -> bool:
-        result = False
         if self.is_authenticated:
-            result = self.__logout()
-        return result
+            result = self._logout()
+        return bool(result)
```
joeldodge79: `result` won't exist if the `self.is_authenticated` condition is false. I think you should keep the removed L92. I wonder what meaning the return value of `logout()` should have ... `True` - "you were logged in and now you're logged out", `False` - "you weren't logged in dummy" | "we failed to log you out" Might be simplest for now to leave the return value as `None` and raise an exception if the API or network barfed.

python/looker/rtl/user_session.py
```diff
@@ -61,48 +52,58 @@ def get_token(self):
...
+        return result
 
-    def __reset(self):
+    def _reset(self):
         self._token = at.AuthToken()
 
-    def __ok(self, response: Union[tp.SDKErrorResponse, tp.SDKSuccessResponse]
-            ) -> Union[tp.SDKErrorResponse, tp.SDKSuccessResponse.value]:
+    def _ok(self,
+            response: tp.Response) -> Union[tp.Response.value, se.SDKError]:
```
joeldodge79: I've been curious about how to annotate types of exceptions raised. I'm not sure there's support for that in annotations. I think this is saying "we'll return either a value or an error" but what you're actually doing is either returning a value or *raising* an error

python/looker/rtl/user_session.py
```diff
@@ -61,48 +52,58 @@ def get_token(self):
             self.login()
         return self._token
 
-    def authenticate(self):
-         # TODO: still to be built based on request
```
joeldodge79: this will be one of many questions to bring up (or maybe we'll resolve it ourselves as we flesh out the implementation more)

joeldodge79:
actually, I'm gonna merge this in now because I'm trying to glue some if this together and I need the updated authenticate method

python/looker/rtl/user_session.py
```diff
@@ -1,46 +1,37 @@
 import abc
-from typing import Optional, Union
+from typing import Optional, Union, Dict
+
 from looker.rtl import requests_transport as rtp
 from looker.rtl import auth_token as at
```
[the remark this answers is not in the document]
josephaxisa: I got the same thing and let @jkaster know last week, it slipped my mind that you'll run into it once you pull `AuthToken`. While writing tests for `AuthToken` I ended up commenting  
out whatever wasn't needed in `looker.sdk.models` as pytest caught it.

python/looker/rtl/user_session.py
josephaxisa: Good point, I agree. Do you think it should just say it returns a `tp.Response.value` then?

python/looker/rtl/user_session.py
josephaxisa: Nice catch, I'll change this.

python/looker/rtl/user_session.py
joeldodge79: yeah, I think the return type should just be `tp.ResponseValue` (see my new alias for that type). The function should still raise `se.SDKError` but I just don't think there's a way using typing/annotations to denote that.

python/looker/rtl/user_session.py
jkaster: I think I intended logout `false` to mean "no logout happened, probably because you weren't logged in." Maybe return `true` either way, meaning "yeah, you're now definitely logged out?"

PR #12, opened 2019-07-27
Serialize response, start on APIMethods

joeldodge79:


joeldodge79:
Created another PR here to encapsulate a few commits of separate work I made. going to just merge this and carry on

PR #13, opened 2019-07-29
UserSession, ApiSettings and tests

josephaxisa:


joeldodge79:
I'd just started to play around with this file too :-) glad I saw your slack! lets have a good coordination meeting this morning

Here's where I was going with this file. I think it can be pretty simple. I haven't tested it and it requires updating `tp.TransportSettings` to not assign a default value of `None` to `headers` which I think is fine. What do you think? (also, I couldn't quite get ConfigParser.read_file to work. the docs seem to say use it first for any required settings then use `read` for optional settings?)

```python
"""Settings for configuring API Access
"""
from __future__ import annotations

import configparser
import dataclasses as dc
from typing import Optional

from looker.rtl import transport as tp


# pylint: disable=too-many-instance-attributes
@dc.dataclass(frozen=True)
class ApiSettings(tp.TransportSettings):
    """API settings uses `looker.ini`."""
    client_id: str
    client_secret: str
    embed_secret: str
    user_id: str

    @classmethod
    def configure(cls, filename: str,
                  section: Optional[str] = None) -> ApiSettings:
        """Read a .ini file and return an ApiSettings instance.
        """
        cfg_parser = configparser.ConfigParser()
        with open(filename) as cfg_file:
            cfg: str = cfg_parser.read_file(cfg_file)

        section = section if section else cfg.sections[0]
        return cls(**cfg[section])
```

python/tests/rtl/test_api_settings.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def test_settings_defaults_to_looker_section(config_file):
+    """ApiSettings should retrieve settings from default (Looker) section 
+    if section is not specified during instantiation."""
+    settings = api_settings.ApiSettings(config_file)
+    assert settings.base_url == 'https://host1.looker.com:19999'
+
+
+@pytest.mark.parametrize("test_section, expected_url",
```
joeldodge79: 👍 for parametrize. though having spent some time in other language testing frameworks I find the configuration of `parametrize` to be a bit kludgey. I wish you could pass in a list of `NamedTuple` (or even better `@dataclass`) instances and use the attribute names as the test func args. The "list as a string" construct here hurts my eyes a little :-)

python/tests/rtl/test_api_settings.py
```diff
@@ -0,0 +1,103 @@
...
+                         [('https://host1.looker.com:19999',
+                           'https://host1.looker.com:19999/api/3.1'),
+                          ('https://host1.looker.com:19999/',
+                           'https://host1.looker.com:19999/api/3.1')])
+def test_versioned_api_url_is_built_properly(config_file, test_url,
+                                             expected_url):
+    """ApiSettings.url should append the api version to the base url"""
+    settings = api_settings.ApiSettings(config_file)
+    settings.base_url = test_url
+    assert settings.url == expected_url
```
joeldodge79: I know you'd brought up the fact that some customer's have whacky API urls that don't conform to  "base_url + api_version". I think this module is the place to handle that for sure and we should update `transport.TransportSettings` as appropriate (removing any logic for determining actual api path in the transport implementation)

python/tests/rtl/test_user_session.py
```diff
@@ -0,0 +1,125 @@
...
+# Set to false if testing locally against self-signed certs. Otherwise leave True
+verify_sl=True
+# leave verbose off by default
+verbose=false
+        """)
+    return filename
+
+
+# pylint: disable=no-method-argument
+class MockTransport():
```
joeldodge79: ah, I see you've followed my pattern :-) I thought you were going to experiment with pytest monkey patching but I'll admit I like this approach better. It might require a little more code but I find it more explicit and readable.

python/tests/rtl/test_user_session.py
```diff
@@ -0,0 +1,125 @@
+# pylint: disable=redefined-outer-name
+
+import pytest
+
+from looker.rtl.user_session import UserSession
+from looker.rtl.api_settings import ApiSettings
```
joeldodge79: even for test code I prefer we'd stick with the "only import modules" convention

python/tests/rtl/test_user_session.py
```diff
@@ -0,0 +1,125 @@
...
+verbose=false
+        """)
+    return filename
+
+
+# pylint: disable=no-method-argument
+class MockTransport():
+    """A mock transport layer used for testing purposes"""
+    def request(method, path, **kwargs):
+        if (method == tp.HttpMethod.POST) and (path.startswith('/login')):
```
joeldodge79: [no parentheses needed](https://google.github.io/styleguide/pyguide.html#33-parentheses)

python/tests/rtl/test_user_session.py
```diff
@@ -0,0 +1,125 @@
...
+            r = type(
+                'Response',
+                (object, ),
+                dict(
+                    ok=True,
+                    # Simulating the deserialization step here. In practice value will be a string.
```
joeldodge79: hm, I'm thinking `UserSession` needs a `serialize.TDeserializeFunc` instance to decode the `AccessToken`. Or if that's overkill I suppose you could just `json.loads()` it

python/tests/rtl/test_user_session.py
```diff
@@ -0,0 +1,125 @@
...
+    return filename
+
+
+# pylint: disable=no-method-argument
+class MockTransport():
+    """A mock transport layer used for testing purposes"""
+    def request(method, path, **kwargs):
+        if (method == tp.HttpMethod.POST) and (path.startswith('/login')):
+            r = type(
+                'Response',
```
joeldodge79: I think you could just directly use a `tp.Response` object once you make the code deserialize the string response from the transport

PR #14, opened 2019-07-30
APIMethods and RequestsTransport stuff

joeldodge79:


python/looker/rtl/user_session.py
```diff
@@ -10,11 +11,16 @@
 class UserSession():
     def __init__(self,
                  settings: st.ApiSettings,
-                 transport: Optional[tp.Transport] = None):
+                 transport: Optional[tp.Transport] = None,
```
joeldodge79: @josephaxisa I wonder if maybe your mypy <=> ide integration isn't working? It also breaks whenever `ml` is regenerated with the current syntax error. If your ide isn't showing type errors, try just running `mypy looker/rtl/user_session.py`

I didn't get to fix all the type stuff in this module, I left some comments in the code. Feel free to take a stab at them

python/looker/rtl/user_session.py
```diff
@@ -100,10 +113,11 @@ def _reset(self):
     def _ok(
             self,
             response: tp.Response,
-            model: sr.SDKModel = None,
-    ) -> Union[tp.TResponseValue, se.SDKError]:
+            structure: Optional[sr.TStructure] = None,
+            ) -> Optional[ml.AccessToken]:
```
joeldodge79: this should actually be `Union[ml.AccessToken, tp.TResponseValue`

PR #15, opened 2019-07-31
Minor changes to UserSession and Transport

josephaxisa:


python/looker/rtl/transport.py
```diff
@@ -30,9 +30,20 @@ class TransportSettings:
     base_url: str
     api_version: str
     headers: Optional[MutableMapping[str, str]] = None
+    """API-versioned base endpoint"""
```
joeldodge79: the [styleguide](http://google.github.io/styleguide/pyguide.html#385-block-and-inline-comments) doesn't address this specifically but I think these should be regular comments rather than docstring-like `str` objects in the code.

python/looker/rtl/transport.py
```diff
@@ -30,9 +30,20 @@ class TransportSettings:
...
+        """Create and return an API-versioned base endpoint."""
+        if not self._url:
+            self._url = "{}{}api/{}".format(
+                self.base_url, "" if self.base_url.endswith("/") else "",
+                self.api_version
+            )
+        return self._url
 
-TResponseValue = Union[str, bytes]
+TResponseValue = Optional[Union[str, bytes]]
```
joeldodge79: 👍

python/looker/rtl/user_session.py
```diff
@@ -1,38 +1,13 @@
 import abc
 from typing import Optional, Union, Dict
 
-from looker.rtl import requests_transport as rtp
-from looker.rtl import auth_token as at
-from looker.rtl import transport as tp
-from looker.rtl import sdk_error as se
-from looker.rtl import api_settings as st
+from looker.rtl import (api_settings as st, auth_token as at,
```
joeldodge79: actually, it looks like the [styleguide](http://google.github.io/styleguide/pyguide.html#313-imports-formatting) indicates how you had this before (one per line rather than parens line continuations). I can't find where it says to do that explicitly but if you look at the bottom of that section the big example shows one per line from the same "package".

It looks repetitive but actually probably easier to read. I'll fix `api_methods.py`. I think the exception again is probably the `typing` imports? I think this:

```python
from typing import (AnyStr, Callable, Dict, IO, MutableMapping, Optional,
                    Union)
```
is easier to read than this:
```python
from typing import AnyStr, Callable, Dict, IO, MutableMapping, Optional
from typing import Union
```

or maybe we should do one per line there too?
```python
from typing import AnyStr
from typing import Callable
.
.
.
from typing import Union
```

I think my vote is option1

josephaxisa: Agreed, option1 is easier to read

PR #18, opened 2019-08-02
Fix issue with cattr not structuring from config file properly

josephaxisa:
- minor formatting changes in line with the google style guide
- added pytest-cov as a test coverage library as a dev dependency

### DIFF
--- a/python/looker/rtl/api_methods.py
+++ b/python/looker/rtl/api_methods.py
@@ -12,6 +12,7 @@ from looker.rtl import user_session as us
 class APIMethods:
     """Functionality for making authenticated API calls
     """
+
     def __init__(self, user_session: us.UserSession,
                  deserialize: sr.TDeserialize, serialize: sr.TSerialize,
                  transport: tp.Transport):
@@ -33,7 +34,7 @@ class APIMethods:
             structure: sr.TStructure,
             path: str,
             query_params: Optional[MutableMapping[str, str]] = None
-            ) -> sr.TDeserializeReturn:
+           ) -> sr.TDeserializeReturn:
         """GET method
         """
         response = self.transport.request(
--- a/python/looker/rtl/api_settings.py
+++ b/python/looker/rtl/api_settings.py
@@ -10,6 +10,17 @@ import cattr
 from looker.rtl import transport as tp
 
 
+def _convert_bool(val: str) -> bool:
+    converted: bool
+    if val.lower() in ('yes', 'y', 'true', 't', '1'):
+        converted = True
+    elif val.lower() in ('no', 'n', 'false', 'f', '0'):
+        converted = False
+    else:
+        raise TypeError
+    return converted
+
+
 @attr.s(auto_attribs=True)
 class ApiSettings(tp.TransportSettings):
     """API Configuration Settings.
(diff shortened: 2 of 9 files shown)


python/looker/rtl/api_methods.py
```diff
@@ -33,7 +34,7 @@ def get(self,
             structure: sr.TStructure,
             path: str,
             query_params: Optional[MutableMapping[str, str]] = None
-            ) -> sr.TDeserializeReturn:
+           ) -> sr.TDeserializeReturn:
```
joeldodge79: funny, our separate usages of yapf produce different results! might be this issue ["YAPF would at times not produce deterministic formatting..."](https://news.ycombinator.com/item?id=17155048) I'm curious to try out [Black](https://github.com/psf/black)

python/looker/rtl/api_settings.py
```diff
@@ -33,5 +44,9 @@ def configure(cls,
         section = section or cfg_parser.sections()[0]
 
         cfg = dict(cfg_parser[section])
-        settings: ApiSettings = cattr.structure(cfg, cls)
+
+        converter = cattr.Converter()
+        converter.register_structure_hook(
+            bool, lambda string, _: _convert_bool(string))
+        settings: ApiSettings = converter.structure(cfg, cls)
```
joeldodge79: 🚀 thanks for figuring this out!

python/looker/rtl/auth_token.py
```diff
@@ -6,6 +6,7 @@
 
 class AuthToken():
     """Used to instantiate or check expiry of an AccessToken object"""
+
```
joeldodge79: again, yapf differences, wonder if we're somehow each using different defaults (I don't see a .style.yapf file anywhere so I presume we're both using the defaults but then why the differences?)

python/looker/rtl/serialize.py
```diff
@@ -7,7 +7,6 @@
 from typing import Callable, Dict, List, Type, Union
 
 import cattr
-
```
joeldodge79: huh, yapf really shouldn't kill this line. there should be a single blank line between groupings of imports (stdlib, 3rd-party, own-lib/app)

python/looker/rtl/transport.py
```diff
@@ -22,24 +21,13 @@ class HttpMethod(enum.Enum):
...
 @attr.s(auto_attribs=True)
 class TransportSettings:
     """Basic transport settings.
     """
     base_url: str
     api_version: str = '3.1'
-    # TODO: this isn't working ... we want "False" to be False
-    # the only way to get false now is leave "verify_ssl" blank
-    verify_ssl: bool = attr.ib(converter=_convert_bool, default=True)
+    verify_ssl: bool = True
```
joeldodge79: 😌 so much cleaner!

python/looker/rtl/user_session.py
```diff
@@ -19,6 +19,7 @@ class UserSessionError(Exception):
 class UserSession():
     """UserSession to provid automatic authentication
     """
+
```
joeldodge79: why doesn't my yapf do this? it's pep8 compliant which I think is the default style

python/looker/rtl/user_session.py
```diff
@@ -70,10 +71,8 @@ def _login(self, user_id: Optional[str]) -> ml.AccessToken:
         self._reset()
 
         serialized = urllib.parse.urlencode({
-            'client_id':
```
joeldodge79: ok, I officially hate my `yapf`

python/looker/rtl/auth_token.py
josephaxisa: Interesting. I use `yapf --style=google --in-place`, is that what you are doing?

PR #19, opened 2019-08-02
Code formatting and new home for base Model

joeldodge79:


python/.flake8
```diff
@@ -0,0 +1,5 @@
+[flake8]
+ignore = E203, E266, E501, W503
+max-line-length = 80
+max-complexity = 18
+select = B,C,E,F,W,T4,B9
```
josephaxisa: Is this intentional? Are we moving away from Pylint to flake8?

PR #21, opened 2019-08-09
Python formatter updates

joeldodge79:
- handle mapping reserved python words
- cattr can't handle `ForwardRef()` types
- all sdk model fields and method "in query" args are Optional[] = None

### DIFF
--- a/python.fmt.ts
+++ b/python.fmt.ts
@@ -42,35 +42,107 @@ export class PythonFormatter extends CodeFormatter {
   indentStr = '    '
   endTypeStr = ''
 
+  // keyword.kwlist
+  pythonKeywords = [
+    'False',
+    'None',
+    'True',
+    'and',
+    'as',
+    'assert',
+    'async',
+    'await',
+    'break',
+    'class',
+    'continue',
+    'def',
+    'del',
+    'elif',
+    'else',
+    'except',
+    'finally',
+    'for',
+    'from',
+    'global',
+    'if',
+    'import',
+    'in',
+    'is',
+    'lambda',
+    'nonlocal',
+    'not',
+    'or',
+    'pass',
+    'raise',
+    'return',
+    'try',
+    'while',
+    'with',
+    'yield'
+  ]
+  pythonTypes: Record<string, IMappedType> = {
+    'number': { name: 'float', default: this.nullStr },
+    'double': { name: 'float', default: this.nullStr },
+    'integer': { name: 'int', default: this.nullStr },
+    'int32': { name: 'int', default: this.nullStr },
+    'int64': { name: 'int', default: this.nullStr },
+    'string': { name: 'str', default: this.nullStr },
+    'password': {name: 'str', default: this.nullStr },
+    'byte': {name: 'bytes', default: this.nullStr },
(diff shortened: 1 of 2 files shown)


python.fmt.ts
```diff
@@ -42,35 +42,107 @@ export class PythonFormatter extends CodeFormatter {
...
   // @ts-ignore
   argGroup(indent: string, args: Arg[]) {
     if ((!args) || args.length === 0) return this.nullStr
     let hash: string[] = []
     for (let arg of args) {
-      hash.push(`'${arg}': ${arg}`)
+      hash.push(`"${arg}": ${arg}`)
     }
-    return `\n${indent}{${hash.join(this.argDelimiter)}}`
+    return `{${hash.join(this.argDelimiter)}}`
```
jkaster: is removal of `indent` because we never expect this to be broken up by line?

python.fmt.ts
```diff
@@ -86,29 +158,35 @@ from typing import *
...
+    const propTypeNameAndDefault = `Optional[${propType.name}] = ${propType.default}`
+    const propDef = `${indent}${propName}: ${propTypeNameAndDefault}`
+    return this.commentHeader(indent, property.description) + propDef
   }
 
   // because Python has named default parameters
   methodSignature(indent: string, method: IMethod) {
-    const type = this.typeMap(method.type)
+    const type = this.typeMapMethods(method.type)
     let bump = indent + this.indentStr
```
jkaster:
```suggestion
    const bump = this.bumper(indent)
```

python.fmt.ts
joeldodge79: I forgot to mention that we'd like to add a final step to code generation. We've settled on [black](https://black.readthedocs.io/en/stable/) for code formatting. So, rather than break things up here, I wanted to write it all one one line and then run `black python/looker/sdk/` which will make the code look as pretty as possible :-)

thoughts on how to do this? hook into `IGeneratorSpec.options` somehow?

joeldodge79:
Regarding the esoteric use of attr/cattr - I think it's still ok. It looks kind of ugly at the bottom of the file but the class definitions themselves look perfectly clean and I'm finally playing around pretending to write example code in my IDE and boy is the auto-completion/context/type-error-catching nice!! So... I think the sausage making at the end of the file is gross but the product is delicious :-)

also, at some point cattrs will resolve https://github.com/Tinche/cattrs/pull/42 at some point and then we can ditch all the `forwardRefHooks` lines at the bottom (but we'll always need the `reservedKeywordHooks` which are only generated for offending classes ... which is only 1 at this point)

python.fmt.ts
jkaster: Yeah, I've been thinking about pretty-printing/post-processing (hey, `p`-`p` FTW!) since we started this project. I think we should make a post-process hook as part of the standard code formatter process. We _might_ want to make the post-processor method part of [targetLanguages.ts](targetLanguages.ts), or just make it part of the last process for the language-specific code generators to perform, since they'll have intimate knowledge of the files they write.

jkaster:
Still looks good to merge to me. Nice work!

PR #23, opened 2019-08-12
Python: build a package take 1

joeldodge79:


joeldodge79:
I was able to get this installed with Pipfile using the following magic (since it's only on test.pypi.org right now):

```
› cat Pipfile
[[source]]
url = "https://test.pypi.org/simple/"
verify_ssl = true
name = "test"

[[source]]
url = "https://pypi.python.org/simple"
verify_ssl = true
name = "pypi"

[dev-packages]
mypy = "*"

[packages]
cattrs = {version = "*",index = "pypi"}
LookerSDK = {version = "*",index = "test"}
```

PR #24, opened 2019-08-13
Integration tests for generated methods.py

josephaxisa:
There's also some formatting changes to the unit tests previously created.

TODOs:
- Divide the integration tests into multiple files so that fixtures are not setup/thorn down more often than needed. `get_test_data` is an example of this. Another solution is to change it to a session scoped fixture.
- mypy is complaining about typing issues in fixture `create_users`
- mypy is also complaining about untyped decorated functions (Example: `@pytest.fixture`). There's a flag to turn this off but I want to find a way to turn it off for a particular directory.
- We are hitting this https://github.com/yaml/pyyaml/issues/202 which has been solved in newer `pyyaml` versions but those versions don't seem to work properly with `pipenv`. I will investigate further.

### DIFF
--- a/python/tests/conftest.py
+++ b/python/tests/conftest.py
@@ -2,7 +2,7 @@ import os.path
 import pytest
 import sys
 import yaml
-from typing import Dict, List, Union
+from typing import cast, Dict, List, Union
 
 sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 
@@ -28,9 +28,8 @@ def get_test_data() -> Dict[str, Union[List[Dict[str, str]], str]]:
 def create_users(
     client: mtds.LookerSDK, test_data: Dict[str, Union[List[Dict[str, str]], str]]
 ) -> None:
-    # TODO: fix mypy type error on 'u'
     # Create some users
-    users = test_data["users"]
+    users = cast(List[Dict[str, str]], test_data["users"])
     email_domain = test_data["email_domain"]
     user_ids: List[int] = []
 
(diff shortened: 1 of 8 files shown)


python/tests/conftest.py
```diff
@@ -1,4 +1,53 @@
...
+    return {"users": test_data["users"], "email_domain": "@testfoo.com"}
+
+
+@pytest.fixture()
+def create_users(
+    client: mtds.LookerSDK, test_data: Dict[str, Union[List[Dict[str, str]], str]]
+) -> None:
+    # TODO: fix mypy type error on 'u'
+    # Create some users
+    users = test_data["users"]
```
joeldodge79: maybe something with https://mypy.readthedocs.io/en/latest/casts.html like

```python
users = cast(List[Dict[str, str]], test_data["users"])
```

I love this:

> give the type checker a little help when it can’t quite understand what is going on.

python/tests/integration/test_methods.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def test_me_returns_correct_result(client: mtds.LookerSDK):
+    """me() should return the right user
+    """
+    me = client.me()
+    assert isinstance(me, ml.User)
+    assert isinstance(me.credentials_api3, list)
+    assert isinstance(me.credentials_api3[0], ml.CredentialsApi3)
+    assert len(me.credentials_api3) > 0
```
joeldodge79: I would swap L53/54 since a test failure where `me.credentials_api3` is empty will currently fail with an `IndexError`

josephaxisa: Good catch

python/tests/conftest.py
josephaxisa: This is great, nice find!

PR #26, opened 2019-08-15
APIMethods tests

josephaxisa:


python/tests/rtl/test_api_methods.py
```diff
@@ -0,0 +1,119 @@
...
+            ),
+            b'{"looker_release_version": "6.18", "current_version": {"full_version": "6.18.4", "status": "fully functional"}}',  # noqa: B950
+        ),
+        (
+            ml.ApiVersion(
+                looker_release_version="6.18",
+                current_version=ml.ApiVersionElement(
+                    full_version="6.18.4", status="fully functional"
+                ),
+                supported_versions=model.EXPLICIT_NULL,
```
joeldodge79: the intention is for the end user to use
```suggestion
                supported_versions=ml.EXPLICIT_NULL,
```
so they won't have to import any rtl code ditto for the other places and then you can remove the model import above

josephaxisa:
Will add more tests that address the types defined in TBody later on

PR #34, opened 2019-08-22
Python: formatter handles HashType

joeldodge79:
- Using MutableMapping[str, T]
- refactored httpCall to be cleaner
- updated Python RTL typing to accept new types
- tests


python.fmt.spec.ts
```diff
@@ -150,7 +150,7 @@ describe('python formatter', () => {
...
+      expect(actual).toEqual('\n@attr.s(auto_attribs=True, kw_only=True)\nclass Workspace(model.Model):\n    """\n    Attributes:\n        id : The unique id of this user workspace. Predefined workspace ids include "production" and "dev"\n        projects : The local state of each project in the workspace\n        can : Operations the current user is able to perform on this object\n    """\n    # The unique id of this user workspace. Predefined workspace ids include "production" and "dev"\n    id: Optional[str] = None\n    # The local state of each project in the workspace\n    projects: Optional[Sequence["Project"]] = None\n    # Operations the current user is able to perform on this object\n    can: Optional[MutableMapping[str, bool]] = None')
```
jkaster: fwiw, it may be easier for you to use `''` as your indent. and, you can also include line breaks in typescript string templates like:
```
const expected = `line 1
line 2
  indented line 3`
```

joeldodge79: thanks! ed02689

PR #35, opened 2019-08-24
Codegen: set required attribute of Property

joeldodge79:
Per the "Required Properties" section of
https://swagger.io/docs/specification/data-models/data-types/
we will inspect the Type.Schema.required list to determine if
a property is required

Refactored makeWriteableType to push more creation logic into
WriteType in order to be able to centralize the decision of
whether a property is required or not inside the Property class.

### DIFF
--- a/sdkModels.spec.ts
+++ b/sdkModels.spec.ts
@@ -26,6 +26,26 @@ describe('sdkModels', () => {
 
   })
 
+  describe('required properties', () => {
+
+    it('CreateQueryTask', () => {
+      const type = apiModel.types['CreateQueryTask']
+      const actual = apiModel.getWriteableType(type)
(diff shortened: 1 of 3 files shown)


joeldodge79:
Here's an example of using this info in the python formatter: https://github.com/looker-open-source/sdk_codegen/compare/jd-codegen-required-properties...jd-python-fmt?expand=1#diff-4bb035f21e6587f3cdea3e75273214eeR252

you can see the syntax in the test (note both the class level property declaration and the `__init__` signature for `query_id` and `result_format`) https://github.com/looker-open-source/sdk_codegen/compare/jd-codegen-required-properties...jd-python-fmt?expand=1#diff-27512a037bd7b230c89789ce3087b517R220

PR #40, opened 2019-08-26
Read credentials from env variables

josephaxisa:
Allow users to pass LOOKER_BASE_URL, LOOKER_CLIENT_ID, LOOKER_CLIENT_SECRET and LOOKER_EMBED_SECRET as env variables.

If settings are defined in both the config file and env variables, then the latter will override the config file settings.

### DIFF
--- a/python/looker_sdk/rtl/api_settings.py
+++ b/python/looker_sdk/rtl/api_settings.py
@@ -49,15 +49,18 @@ class ApiSettings(transport.TransportSettings):
         cfg = dict(cfg_parser[section])
 
         env_base_url = cast(str, os.getenv("LOOKER_BASE_URL"))
-        env_client_id = cast(str, os.getenv("LOOKER_CLIENT_ID"))
-        env_client_secret = cast(str, os.getenv("LOOKER_CLIENT_SECRET"))
-        env_embed_secret = cast(str, os.getenv("LOOKER_EMBED_SECRET"))
         if env_base_url:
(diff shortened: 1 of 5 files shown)


python/looker_sdk/rtl/api_settings.py
```diff
@@ -21,17 +22,17 @@ def _convert_bool(val: str, _: bool) -> bool:
...
+    client_secret: str
+    embed_secret: str = ""
     # User ID to impersonate (optional)
-    user_id: str = attr.ib(default="", kw_only=True)
-    verbose: bool = attr.ib(default=False, kw_only=True)
+    user_id: str = ""
+    verbose: bool = False
 
     @classmethod
     def configure(
```
joeldodge79: We should make having a "looker.ini" file totally optional. maybe a `try/except` around L43~49 (initializing `cfg` to an empty dict if no config file)

python/looker_sdk/rtl/api_settings.py
```diff
@@ -47,6 +48,28 @@ def configure(
 
         cfg = dict(cfg_parser[section])
 
+        env_base_url = cast(str, os.getenv("LOOKER_BASE_URL"))
```
jkaster: I don't see "LOOKER_API_VERSION" configuration support

joeldodge79: added

PR #42, opened 2019-09-06
WIP: make client_id and client_secret ephemeral

josephaxisa:


python/looker_sdk/rtl/api_settings.py
```diff
@@ -49,49 +49,48 @@ def configure(
...
-            config_data["client_secret"] = env_client_secret
-
-        env_embed_secret = cast(str, os.getenv("LOOKER_EMBED_SECRET"))
-        if env_embed_secret:
-            config_data["embed_secret"] = env_embed_secret
+        config_data = cls.read_ini(filename, section)
 
         api_version = cast(str, os.getenv("LOOKER_API_VERSION"))
         if api_version:
             config_data["api_version"] = api_version
```
jkaster: Is there an immediate if/ternary operator we could use for cascading the defaults so this is one statement? e.g.

```python
api_version = cast(str, os.getenv("LOOKER_API_VERSION") or "3.1"
```
for example

python/looker_sdk/rtl/api_settings.py
```diff
@@ -49,49 +49,48 @@ def configure(
...
-        # Remove required params from config dictionary if they are empty strings
-        required_params = ["base_url", "client_id", "client_secret"]
-        for key, val in list(config_data.items()):
-            if key in required_params and not val:
-                config_data.pop(key)
+        env_base_url = cast(str, os.getenv("LOOKER_BASE_URL"))
+        if env_base_url:
+            config_data["base_url"] = env_base_url
+
+        required = ["base_url", "client_id", "client_secret"]
```
jkaster:
```suggestion
        required = ["base_url", "api_version"]
```

python/looker_sdk/rtl/api_settings.py
```diff
@@ -8,6 +8,7 @@
 import attr
 import cattr
 
+from looker_sdk import error
```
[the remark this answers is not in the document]
josephaxisa: It's not a reserved keyword and we have used it exactly like this in other places in the Python RTL. I'm open to aliasing it if you think that's best

python/looker_sdk/rtl/api_settings.py
josephaxisa: `api_version = cast(str, os.getenv("LOOKER_API_VERSION") or config_data.get("api_version")` works actually. I think the main reason I did it this way was because `black` still splits the expression into multiple lines because the max line length gets exceeded

python/looker_sdk/rtl/api_settings.py
josephaxisa: `api_version` is not required because the parent class of `ApiSettings`, `transport.TransportSettings` has an `api_version` with a default of `"3.1"`. 

The reason `client_id` and `client_secret` are set to required is so that `ApiSettings` throws an error during the configuration phase if they're not found. But they're not used to instantiate the class since we don't want to store them in memory longer than we need.

python/looker_sdk/rtl/api_settings.py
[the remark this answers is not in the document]
jkaster: No reason to change it then. Looks good.

python/looker_sdk/rtl/api_settings.py
jkaster: gotcha. If it's a battle with the code formatting tools, never mind :)

python/looker_sdk/rtl/api_settings.py
jkaster: Gotcha on `api_version`. I didn't check to see how much you mirrored the TS design. 

I'm not sure I'm following how `client_id` and `client_secret` get established here for the required check when they're not properties of the class. I'll look at the updated test code some more to see if I can understand it.

python/tests/rtl/test_api_settings.py
```diff
@@ -138,26 +142,25 @@ def test_credentials_from_env_variables_override_config_file(
     monkeypatch.setenv("LOOKER_BASE_URL", "https://host1.looker.com:19999")
     monkeypatch.setenv("LOOKER_CLIENT_ID", "id123")
```
jkaster: I'm not clear on how the environment variable credentials are tested any more

joeldodge79: @josephaxisa yeah, we should remove `LOOKER_CLIENT_ID` and `LOOKER_CLIENT_SECRET` from these tests and do some testing in `test_user_session.py` (which should be renamed to `test_auth_session.py`) to confirm the behavior when those settings are present via config file, env variable, or absent.

python/looker_sdk/rtl/auth_session.py
```diff
@@ -95,10 +96,21 @@ def login_user(self, sudo_id: int) -> None:
...
+        config_data = self.settings.read_ini(
+            self.settings._filename, self.settings._section
+        )
+        client_id = os.getenv("LOOKER_CLIENT_ID") or config_data.get("client_id")
+        client_secret = os.getenv("LOOKER_CLIENT_SECRET") or config_data.get(
+            "client_secret"
+        )
+
+        if not (client_id or client_secret):
+            raise error.SDKError(f"Required auth credentials not found.")
```
joeldodge79: don't need an `f` string here

josephaxisa: I only left `LOOKER_CLIENT_ID` and `LOOKER_CLIENT_SECRET` to ensure that if the credentials are passed either through the config file or through env variables, `cattr.structure` works as expected and the credentials are not stored as attributes in the ApiSettings once instantiated

joeldodge79: ah, that makes sense

PR #46, opened 2019-09-12
Python: stamp looker/api version for user-agent

joeldodge79:
For the codegen this is currently almost an exact copy of the typescript
generator with variable names snake_cased for python style. Might be
worth abstracting those variables out into config and moving the method
implementation into the base class.

For the python rtl - added User-Agent header like "PY-SDK 3.1.6.19". We
might want to also consider the same prefixing strategy we plan for
environment variable config here as well. (e.g. we'd be "LOOKER-PY-SDK
3.1.6.19")

### DIFF
--- a/src/typescript.gen.ts
+++ b/src/typescript.gen.ts
@@ -276,7 +276,7 @@ export interface IDictionary<T> {
       let content = fs.readFileSync(stampFile, utf8)
       const lookerPattern = /lookerVersion = ['"]\d+\.\d+['"]/i
       const apiPattern = /apiVersion = ['"]\d+\.\d+['"]/i
-      const envPattern = /environmentPrefix = ['"]\d+\.\d+['"]/i
+      const envPattern = /environmentPrefix = ['"]\w+['"]/i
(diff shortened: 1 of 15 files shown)


src/typescript.gen.ts
```diff
@@ -276,7 +276,7 @@ export interface IDictionary<T> {
       let content = fs.readFileSync(stampFile, utf8)
       const lookerPattern = /lookerVersion = ['"]\d+\.\d+['"]/i
       const apiPattern = /apiVersion = ['"]\d+\.\d+['"]/i
-      const envPattern = /environmentPrefix = ['"]\d+\.\d+['"]/i
+      const envPattern = /environmentPrefix = ['"]\w+['"]/i
```
jkaster: Doh. Good catch on my copy pasta mistake.

joeldodge79:
> LGTM. After you merge this in I'll fix the `LOOKER_` hard-coded prefixes in NodeSession.ts

@jkaster I went ahead and updated `nodeSessin.ts`. I also made everything use `environmentPrefix` rather than hardcoding `LOOKERSDK`

PR #48, opened 2019-09-18
Python: implement DelimSequence

joeldodge79:
This handles the "CSV" collection format

### DIFF
--- a/python/looker_sdk/rtl/api_methods.py
+++ b/python/looker_sdk/rtl/api_methods.py
@@ -95,6 +95,8 @@ class APIMethods:
                 params[k] = f'{v.isoformat(timespec="minutes")}Z'
             elif isinstance(v, str):
                 params[k] = v
+            elif isinstance(v, model.DelimSequence):
+                params[k] = str(v)
             else:
                 params[k] = json.dumps(v)
         return params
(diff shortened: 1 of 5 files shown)


joeldodge79:
example usage in generated methods:

```python

# PUT /connections/{connection_name}/test -> Sequence[models.DBConnectionTestResult]
def test_connection(
    self,
    # Name of connection
    connection_name: str,
    # Array of names of tests to run
    tests: Optional[model.DelimSequence[str]] = None,
) -> Sequence[models.DBConnectionTestResult]:
    """Test Connection"""
    response = self.put(
        f"/connections/{connection_name}/test",
        Sequence[models.DBConnectionTestResult],
        query_params={"tests": tests},
    )
    assert isinstance(response, list)
    return response
```

joeldodge79:
Gonna wait for @jkaster's changes on the typescript side and then update the python code generator. also need to write a test in `tests/integration/test_methods.py`

jkaster:
Added a Typescript test for `all_users` that used a DelimArray of user ids if you want to adopt that pattern

joeldodge79:
added a similar `all_users` test and aliased the `looker_sdk.rtl.model.DelimSequence` as `looker_sdk.sdk.models.DelimSequence` for easier client access

PR #53, opened 2019-09-28
Implemented binary mode response handling

jkaster:
Implemented trinary binary response mode options:
- binary only
- string only
- mixed

Tweaked comment generation for Typescript

Updated the rationale doc for Delim types and binary mode handling


rationale.md
```diff
@@ -383,42 +427,192 @@ async create_query_task(
...
+
+- `suffix`, which defaults to an empty string.
+
+### Binary response handling
+
+The popular request libraries currently used by the RTLs for both Python and the Node-based Typescript SDK treat response content as string by default, with binary mode as an opt-in setting. This results in the download of binary content like images or PDFs from SDK methods to be corrupted.
+
+Until we've [resolved this issue](#binary-response-detection) with replacement HTTP request methods, the code generator will assist with binary responses in the following ways (the provided examples in Typescript).
+
+#### Response can be either binary and string
```
josephaxisa: What are your thoughts about also showing the python equivalent of the snippets below? We won't have to do it for all languages we support but given that Python is the most popular language amongst our customers it might be worth considering (once we implement this support for the Python SDK at least.)

rationale.md
```diff
@@ -383,42 +427,192 @@ async create_query_task(
...
-### Missing features
+#### Methods with delimited list types
+
+As indicated in the [style values section of the OpenAPI specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#style-values), a list of values can be provided to a REST call in a variety of delimited formats. The Looker API uses "csv" formatting for some requests that receive simple lists, such as IDs.
+
+In Typescript, calling `toString()` on an array of numbers such as `[1,2,3]` would return `1, 2, 3`, which is fine for the Looker API methods expecting "csv-formatted" lists. However, in Python, the equivalent function returns `[1, 2, 3]` which would cause a parsing error by the Looker API endpoint.
+
+As indicated in the OpenAPI specification style values, other delimiters such as pipe (`|`), space, or tab characters could be expected by an API endpoint.
+
+To guarantee consistency in the serialization of "csv-formatted" array values, the code generator uses a **Delim** type for parameters that are specified as having a special delimiter. In Typescript, this type is called `DelimArray`. In Python, it's called `DelimSequence`.
```
dthorpe: Any chance of coalescing these two terms for the same thing?  `DelimSequence` would work in TypeScript, yes?

rationale.md
```diff
@@ -383,42 +427,192 @@ async create_query_task(
...
+A **mixed mode** response that can be either binary or string (e.g. text or json) has a note placed into the method header.
+
+```typescript
+/**
+  * POST /queries/run/{result_format} -> string}
+  *
+  * **Note**: Binary content may be returned by this method. Add `binaryMode` to the `options` parameter to correctly receive binary content
+  */
+async run_inline_query(
+  request: Partial<IRequestrun_inline_query>,
```
dthorpe: Can we transform snake case to camel case for generated type names?  For another PR.

rationale.md
```diff
@@ -383,42 +427,192 @@ async create_query_task(
...
+    },
+    request.body,
+    options,
+  )
+}
+```
+
+#### Binary responses
+
+A method that only has binary responses will have the request configured for binary inside the generated SDK method. Images and PDFs are considered binary.
```
dthorpe: delete 'considered'

rationale.md
```diff
@@ -383,42 +427,192 @@ async create_query_task(
...
+  */
+async render_task_results(
+  /**
+    * @param {string} render_task_id Id of render task
+    */
+  render_task_id: string,
+  options?: Partial<ITransportSettings>,
+) {
+  options = { options, ...binaryMode }
+  return this.get<string, IError>(
```
dthorpe: Looks like this is returning a string?  Shouldn't binary be returned as an array of byte?

rationale.md
```diff
@@ -383,42 +427,192 @@ async create_query_task(
...
+}
+```
+
+### Future enhancements
+
+While the new SDKs are easier to use than the previous options, there are still some additional improvements we are working on.
+
+#### Streaming support
+
+Looker intends to provide streaming (http chunking) methods for all streamable API endpoints, in addition to the current non-streaming methods.
```
dthorpe: If we're going to mention it, perhaps a little more explanation of what streaming is good for?

> Assigning an API response to a local variable is simple and easy, but if the response is extremely large (gigabytes of data), deserializing into client app memory will probably cause the client app to crash with an out-of-memory exception. Streaming large responses chunk by chunk can achieve constant memory use for arbitrarily large responses, at a cost of additional code complexity.

rationale.md
```diff
@@ -383,42 +427,192 @@ async create_query_task(
...
+  return this.get<IRole, IError>(
+    encodeURI(`/roles/${role_id}`),
+    null,
+    null,
+    options,
+  )
+}
+```
+
+### Future enhancements
```
dthorpe: Future Research?
to soften the implied commitment

rationale.md
jkaster: After discussing with @joeldodge79 and @dthorpe I think we can completely hide the `BinaryMode` flag requirement from the user. Still working on rewriting what I did this weekend for that. The comments in the method header will still be worth producing.

rationale.md
jkaster: It was easy enough to do and only changes code in one place, so I've done it in this PR.

rationale.md
jkaster: "A string is a series of bytes" https://kevin.burke.dev/kevin/node-js-string-encoding/

rationale.md
jkaster: good idea

rationale.md
jkaster: added

rationale.md
jkaster: gone!

typescript/looker/rtl/transport.ts
```diff
@@ -217,6 +226,32 @@ export interface ITransportSettings {
...
+/**
+ * Is the content type binary or "string"?
+ * @param {string} contentType
+ * @returns {ResponseMode.binary | ResponseMode.string}
+ */
+export function responseMode(contentType: string) {
+  if (contentType.match(contentPatternBinary)) {
+    return ResponseMode.string
+  } else if (contentType.match(contentPatternString)) {
+    return ResponseMode.binary
```
joeldodge79: Why does `contentPatternBinary` contain the pattern for string and return string? same question in reverse for `contentPatternString`

jkaster: Whoops! Fixed and tightened up the patterns a bit.

typescript/looker/rtl/nodeTransport.ts
```diff
@@ -172,23 +174,26 @@ export class NodeTransport implements ITransport {
...
-      return result instanceof Object ? result : JSON.parse(result)
-    } catch (error) {
-      console.log(res.body)
-      return Promise.reject(error)
+  const mode = responseMode(contentType)
+  let result = await res.body
+  if (mode === ResponseMode.string) {
+    if (!isUtf8(contentType)) {
+      // always convert to UTF-8 from whatever it was
+      result = Buffer.from(result.toString(), 'utf8')
```
joeldodge79: I may be mis-reading the `Buffer` docs but it seems like `Buffer.toString()` defaults to output using the utf8 encoding so is this necessary?

rationale.md
```diff
@@ -511,22 +511,22 @@ Both `DelimArray` and `DelimSequence` accept optional overrides for:
...
 
 ```typescript
 /**
-  * POST /queries/run/{result_format} -> string}
-  *
-  * **Note**: Binary content may be returned by this method. Add `binaryMode` to the `options` parameter to correctly receive binary content
-  */
+ * POST /queries/run/{result_format} -> string
+ *
+ * **Note**: Binary content may be returned by this method. Add `binaryMode` to the `options` parameter to correctly receive binary content
```
dthorpe: Still relevant?

rationale.md
```diff
@@ -553,7 +553,7 @@ async run_inline_query(
 
 #### Binary responses
 
-A method that only has binary responses will have the request configured for binary inside the generated SDK method. Images and PDFs are considered binary.
+A method that only has binary responses will have a note that the response is binary.
```
dthorpe: Still relevant?

rationale.md
jkaster: Removed the "Add binaryMode ..." bits. Thanks!

rationale.md
jkaster: Yes, we still generate the note into the method header to call out that the response will be binary

typescript/looker/rtl/nodeTransport.ts
```diff
@@ -172,23 +174,26 @@ export class NodeTransport implements ITransport {
...
+    }
+    if (contentType.match(/^application\/.*\bjson\b/g)) {
+      try {
+        result = result instanceof Object ? result : JSON.parse(result)
```
joeldodge79: I think I'm missing something - I thought `result` was always a `Buffer` now (except for maybe on L160 you allow overriding `encoding`... why?) If so, don't we want to `toString()` it and then `JSON.parse()` it before returning it? I'm coming from a strong python/requests bias as I'm working on this same code there right now so I might be totally off on this.

typescript/looker/rtl/constants.ts
```diff
@@ -26,3 +26,22 @@ export const lookerVersion = '6.21'
 export const apiVersion = '3.1'
 export const sdkVersion = `${apiVersion}.${lookerVersion}`
 export const environmentPrefix = 'LOOKERSDK'
+export const matchCharset = ';.*charset='
```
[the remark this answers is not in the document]
jkaster: For now I think that's ok. If you got ideas for making them more easily re-usable, share em! We *can* do the same trick with the versions stamp and munge the RegEx as needed from the values in `constants.ts` to be written to Python or other `constants.*`. 

Let's decide tomorrow if we want to keep updating this branch or make that a separate PR.

typescript/looker/rtl/constants.ts
```diff
@@ -26,3 +26,22 @@ export const lookerVersion = '6.21'
...
+ * @type {string} Regular expression for matching charset=utf-8 in Content-Type
+ */
+export const matchCharsetUtf8 = `${matchCharset}.*\\butf-8\\b`
+
+/**
+ * Matching rules for string/text types. String matches must be checked *before* binary matches
+ * @type {string} Regular expression for matching Content-Type headers
+ */
+export const matchModeString = `(^application\\/.*(\\bjson\\b|\\bxml\\b|\\bsql\\b|\\bgraphql\\b|\\bjavascript\\b|\\bx-www-form-urlencoded\\b)|^text\\/|${matchCharset})`
+
```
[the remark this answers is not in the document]
jkaster: Oh, that's nice.

typescript/looker/rtl/nodeTransport.ts
jkaster: You **may** be right, but I couldn't get it to behave the way I expected without this. I definitely want to improve this code if possible.

typescript/looker/rtl/nodeTransport.ts
jkaster: I think we covered in person, but basically we can't be **sure** the response isn't already converted to a deserialized object because of the request parameters the user could pass in.

PR #54, opened 2019-09-30
Python update readme and setup requirements

joeldodge79:
Found a bug for folks who only use pip: they were getting cattrs=0.9.0
instead of the cattrs=1.0.0rc0. Adding the minimum requirement in
setup.py fixes this.

Also tweaked the "sample project setup" documentation to be more
accurate for installing python 3.7 via pyenv as well as making sure the
pipenv environment is created using the correct version of python.

### DIFF
--- a/python/README.rst
+++ b/python/README.rst
@@ -16,14 +16,31 @@ and please indicate which language SDK you're using in the report.
 Sample project setup
 --------------------
 
-In these instructions we'll use `pyenv <https://github.com/pyenv/pyenv#installation>`_
-to install python3.7 and
-`pipenv <https://docs.pipenv.org/en/latest/#install-pipenv-today>`_
-to manage project dependencies. Here is how to install them on a mac
+Install python 3.7. We highly recommend using
+`pyenv <https://github.com/pyenv/pyenv#installation>`_ to install
+different versions of python. Mac users should use
+`homebrew <https://brew.sh/>`_ to install pyenv:
 
 .. code-block:: bash
 
-    brew update && brew install pyenv && brew install pipenv
+    brew install pyenv
+
+Follow the **remaining steps 3 - 5** of
+https://github.com/pyenv/pyenv#basic-github-checkout otherwise your python3.7
+installation may break.
+
+Now you're ready  install python 3.7:
+
+.. code-block:: bash
+
+    pyenv install 3.7.4
+
+Next we'll use `pipenv <https://docs.pipenv.org/en/latest/#install-pipenv-today>`_
+as an awesome enhancement to pip to manage project dependencies.
+
+.. code-block:: bash
+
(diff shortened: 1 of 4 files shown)


python/README.rst
```diff
@@ -36,14 +53,13 @@ Install python3.7 and use it for this directory
...
     cd looker-sdk-example/
-    pyenv install 3.7.4
     pyenv local 3.7.4
 
 Install looker_sdk using pipenv
 
 .. code-block:: bash
 
-    $ pipenv install --pre looker_sdk
+    pipenv install --python 3.7 --pre looker_sdk
```
josephaxisa: What do you think of splitting this into two commands? Clearer, unnecessary or more confusing? 
```
pipenv shell --python 3.7.4
pipenv install --pre looker_sdk
```

python/README.rst
```diff
@@ -16,14 +16,31 @@ and please indicate which language SDK you're using in the report.
...
 .. code-block:: bash
 
-    brew update && brew install pyenv && brew install pipenv
+    brew install pyenv
+
+Follow the **remaining steps 3 - 5** of
+https://github.com/pyenv/pyenv#basic-github-checkout otherwise your python3.7
+installation may break.
+
+Now you're ready  install python 3.7:
```
josephaxisa: add 'to' before 'install'

python/README.rst
```diff
@@ -16,14 +16,31 @@ and please indicate which language SDK you're using in the report.
...
+`pyenv <https://github.com/pyenv/pyenv#installation>`_ to install
+different versions of python. Mac users should use
+`homebrew <https://brew.sh/>`_ to install pyenv:
 
 .. code-block:: bash
 
-    brew update && brew install pyenv && brew install pipenv
+    brew install pyenv
+
+Follow the **remaining steps 3 - 5** of
```
josephaxisa: Might be worth numbering the steps if we're referring to them like this.

python/README.rst
```diff
@@ -36,14 +53,13 @@ Install python3.7 and use it for this directory
 .. code-block:: bash
```
josephaxisa: I think L34 should read something along the lines of, "Set the local Python version for our project to 3.7.4"

python/README.rst
joeldodge79: this is in reference to the link (https://github.com/pyenv/pyenv#basic-github-checkout) which does have the steps numbered :-)

python/README.rst
joeldodge79:
```suggestion
Now you're ready to install python 3.7:
```

python/README.rst
joeldodge79: I don't think there's a reason to launch them into `pipenv shell` I'd rather stick with a one-liner

PR #55, opened 2019-10-02
add a transport_options arg to generated methods

josephaxisa:
This will allow users to pass a `transport.TransportSettings` object to override defaults or provide new transport settings  for a given call.

For now I set the default connection timeout to an arbitrary 1800 seconds. Will change this to a more reasonable one later on. The read timeout is 180 seconds to mimic the TS implementation.

### DIFF
--- a/python/looker_sdk/rtl/transport.py
+++ b/python/looker_sdk/rtl/transport.py
@@ -60,7 +60,7 @@ class TransportSettings:
     base_url: str = ""
     api_version: str = "3.1"
     verify_ssl: bool = True
-    timeout: TConnectAndReadTimeout = (1800, 120)
+    timeout: TConnectAndReadTimeout = (120, 120)
     headers: Optional[MutableMapping[str, str]] = None
 
     @property
--- a/python/looker_sdk/rtl/transport.py
+++ b/python/looker_sdk/rtl/transport.py
@@ -24,7 +24,7 @@
 """
 import abc
 import enum
-from typing import Callable, Dict, MutableMapping, Optional, Union
+from typing import Callable, Dict, MutableMapping, Optional, Union, Tuple
 
 import attr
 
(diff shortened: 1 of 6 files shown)


python/looker_sdk/rtl/transport.py
```diff
@@ -44,14 +44,23 @@ class HttpMethod(enum.Enum):
...
 @attr.s(auto_attribs=True, kw_only=True)
 class TransportSettings:
     """Basic transport settings.
     """
 
-    base_url: str
+    base_url: str = ""
     api_version: str = "3.1"
     verify_ssl: bool = True
+    timeout: TConnectAndReadTimeout = (1800, 120)
```
joeldodge79: now that I see it, I feel like `1800` is way too long to wait to establish a connection. I say lets keep them the same number. Unless you feel strongly otherwise then we can leave it

```suggestion
    timeout: TConnectAndReadTimeout = (120, 120)
```

src/python.gen.ts
```diff
@@ -263,6 +265,7 @@ ${this.hooks.join('\n')}
   httpArgs(indent: string, method: IMethod) {
     let result = this.argFill('', this.argGroup(indent, method.cookieArgs))
     result = this.argFill(result, this.argGroup(indent, method.headerArgs))
+    result = this.argFill(result, `transport_options=transport_options`)
```
joeldodge79: Let's add this to `src/python.gen.spec.ts`

python/looker_sdk/rtl/transport.py
```diff
@@ -44,14 +44,23 @@ class HttpMethod(enum.Enum):
...
 @attr.s(auto_attribs=True, kw_only=True)
 class TransportSettings:
     """Basic transport settings.
     """
 
-    base_url: str
+    base_url: str = ""
     api_version: str = "3.1"
     verify_ssl: bool = True
+    timeout: TConnectAndReadTimeout = (120, 120)
```
joeldodge79:
```suggestion
    timeout: TConnectAndReadTimeoutSeconds = (120, 120)
```

joeldodge79: though we may want to rename `TConnectAndReadTimeoutSeconds` to something simpler like `TTimeoutSeconds`

Issue #56, opened 2019-10-03
Python SDK doesn't support binary downloads


joeldodge79:
for example `render_task_results` returns a `str` (and is annotated as such) when it should return a `bytes` (and be thusly annotated). The string it returns is `requests` best guess at trying to decode the raw bytes so the result is useless (e.g. "corrupted png")

PR #60, opened 2019-10-04
Python: binary reponse handling

joeldodge79:
Modified the transport interface so that it always returns bytes as well
as whether the "response mode" is binary or string and finally the text
encoding if any. The caller then decides how to de-serialize the
response.

fixes:#56

### DIFF
--- a/python/looker_sdk/rtl/api_methods.py
+++ b/python/looker_sdk/rtl/api_methods.py
@@ -44,7 +44,7 @@ TBody = Optional[
     ]
 ]
 TStructure = Optional[Union[Type[str], serialize.TStructure]]
-TReturn = Optional[Union[transport.TResponseValue, serialize.TDeserializeReturn]]
+TReturn = Optional[Union[str, bytes, serialize.TDeserializeReturn]]
 TQueryParams = MutableMapping[
     str, Union[None, bool, str, int, Sequence[int], Sequence[str], datetime.datetime]
 ]
@@ -73,15 +73,20 @@ class APIMethods:
         self.auth.logout()
 
     def _return(self, response: transport.Response, structure: TStructure) -> TReturn:
+        encoding = response.encoding
         if not response.ok:
-            raise error.SDKError(response.value)
+            raise error.SDKError(response.value.decode(encoding=encoding))
         ret: TReturn
         if structure is None:
             ret = None
-        elif structure is str:
+        elif response.response_mode == transport.ResponseMode.BINARY:
             ret = response.value
         else:
-            ret = self.deserialize(response.value, structure)
+            value = response.value.decode(encoding=encoding)
+            if structure is Union[str, bytes] or structure is str:  # type: ignore
+                ret = value
+            else:
+                ret = self.deserialize(value, structure)
         return ret
 
     def _convert_query_params(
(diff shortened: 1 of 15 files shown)


python/looker_sdk/rtl/transport.py
```diff
@@ -77,7 +86,27 @@ class Response:
     """
 
     ok: bool
-    value: TResponseValue
+    value: bytes
+    response_mode: ResponseMode
+    encoding: str = "utf-8"
```
joeldodge79: This is the change to the interface - the `Response` object now has `value: bytes` and the additional `response_mode` and `encoding` properties

jkaster: Thanks for calling this out.

src/python.gen.ts
```diff
@@ -195,16 +196,32 @@ ${this.hooks.join('\n')}
     return propDef
   }
 
+  private methodReturnType(method: IMethod) {
+    const type = this.typeMapMethods(method.type)
+    let returnType = type.name
+    if (method.responseIsBoth()) {
+      returnType = 'Union[str, bytes]'
```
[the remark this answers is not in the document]
joeldodge79: good catch - I was just narrowly assuming it's always a str because that's what I've seen in our API but of course it could be anything

python/looker_sdk/rtl/api_methods.py
```diff
@@ -73,15 +73,20 @@ def __exit__(self, *exc) -> None:
...
         ret: TReturn
         if structure is None:
             ret = None
-        elif structure is str:
+        elif response.response_mode == transport.ResponseMode.BINARY:
             ret = response.value
         else:
-            ret = self.deserialize(response.value, structure)
+            value = response.value.decode(encoding=encoding)
+            if structure is Union[str, bytes] or structure is str:  # type: ignore
```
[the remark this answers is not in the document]
joeldodge79: The generated methods pass in this `structure` argument (which is just the method return type) to tell us how to deserialize. In the case of `method.responseIsBoth()` methods which have `structure == Union[str, bytes]` (e.g. `run_url_encoded_query()`) or methods where the return type is just `str` (e.g `create_git_deploy_key()`),  we just return the string as is (since we already decoded the bytes on the line above)

All other `structure` types will get passed to `self.deserialize()`

There is a bug here: the `structure` argument to `self.deserialize(value, structure)` on L89 could theoretically be something like `Union[models.Dashboard, bytes]` (though it doesn't seem like we have anything like that in our Looker API). `self.deserializeI` won't take that type of structure argument so I need to add a check for this case and pull the non-bytes-type out of the `Union` and pass it instead

There might be another bug: I don't see it in our API but I suppose `int`, `float`, and other basic types could be method return types as well and I'm not handling those.

I might just file an issue for these last two potential bugs since this code does work for the Looker API right now.

python/looker_sdk/rtl/transport.py
```diff
@@ -74,17 +69,45 @@ def agent_tag(self) -> str:
...
+
+
+_STRING_MODE = re.compile(constants.RESPONSE_STRING_MODE)
+_BINARY_MODE = re.compile(constants.RESPONSE_BINARY_MODE)
```
jkaster:
```suggestion
_STRING_MODE = re.compile(constants.RESPONSE_STRING_MODE, re.IGNORECASE)
_BINARY_MODE = re.compile(constants.RESPONSE_BINARY_MODE, re.IGNORECASE)
```

PR #65, opened 2019-10-07
Typescript sdk streaming prototype

jkaster:
This is a _working_ streaming implementation for Typescript, but by no means the final implementation. Expect to see at least one more iteration

From the integration test that verifies streaming works:
```typescript
            const csvFile = './query.csv'
            const writer = fs.createWriteStream(csvFile)
            try {
              await sdk.stream.run_inline_query(async (readable: Readable) => {
                return new Promise<any>((resolve, reject) => {
                  readable.pipe(writer)
                    .on('error', reject)
                    .on('finish', resolve)
                })
              }, request)
              expect(isFileSync(csvFile)).toEqual(true)
              const contents = fs.readFileSync(csvFile, utf8)
              fs.unlinkSync(csvFile)
              expect(isFileSync(csvFile)).toEqual(false)
              expect(contents).toEqual(csv)
            } catch (e) {
              throw e
            }
```


src/sdkGenerator.spec.ts
```diff
@@ -1,25 +1,32 @@
...
+    """Create Look"""
+    response = self.post(
+                        f"/looks",
+            models.LookWithQuery,
+            query_params={"fields": fields},
+            body=body,
+            transport_options=transport_options
+    )
+    assert isinstance(response, models.LookWithQuery)
+    return response`)
```
joeldodge79: yikes, sorry we broke this

jkaster: No big deal. We still have a moving target!

PR #73, opened 2019-10-22
Python: s/looker_client/sdk/ in tests

joeldodge79:
and relax mypy Any checking in all tests so we don't
have such a trashpile of "  #type: ignore"


joeldodge79:
@josephaxisa I was going to add a test when I thought `create_user_attribute` was broken. I did some slight refactoring in prep for that change but it turns out we don't need the test because it's not broken :-) however, I still think this cleanup is worth merging in.

PR #74, opened 2019-10-29
typescript formatter: optional only if not required

joeldodge79:
explicitly setting a property to null (being "nullable") is different
from whether the presence of a property is required ("required").

### DIFF
--- a/src/typescript.gen.ts
+++ b/src/typescript.gen.ts
@@ -134,7 +134,7 @@ export interface IDictionary<T> {
   }
 
   declareProperty(indent: string, property: IProperty) {
-    const optional = (property.nullable || !property.required) ? '?' : ''
+    const optional = !property.required ? '?' : ''
     if (property.name === strBody) {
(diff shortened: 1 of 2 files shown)


src/typescript.gen.ts
```diff
@@ -134,7 +134,7 @@ export interface IDictionary<T> {
   }
 
   declareProperty(indent: string, property: IProperty) {
-    const optional = (property.nullable || !property.required) ? '?' : ''
+    const optional = !property.required ? '?' : ''
```
jkaster: Dang that was simple! Were there additional upstream changes?

joeldodge79: I'd already updated it in [this PR](https://github.com/looker-open-source/sdk-codegen/pull/35/files) and it looks like you incorporated that in [e5ebe98](https://github.com/looker-open-source/sdk-codegen/commit/e5ebe98afe38e9120e5d10022e13bd13fffa0261#diff-60c58a6f980980a1d10a67fe46d1169aR100) but you left in the `property.nullable` check

PR #80, opened 2019-11-01
python updates

joeldodge79:


src/python.gen.ts
```diff
@@ -464,7 +464,7 @@ ${this.hooks.join('\n')}
       if (pipEnvExists.includes('pipenv')) {
         // pipenv check completed without error
         run('pipenv', ['update'])
-        run('pipenv', ['run', 'black', `${this.codePath}/${this.packagePath}/sdk/`])
+        run('pipenv', ['run', 'black', `${this.codePath}/${this.packagePath}/sdk/*.py`])
```
jkaster: Is this called only once for all files? It appears to be called for every file.

python/looker_sdk/rtl/model.py
```diff
@@ -24,7 +24,7 @@
 """
 
 import collections
-from typing import Any, cast, Iterable, MutableSequence, Optional, TypeVar
+from typing import Any, cast, Iterable, Sequence, Optional, TypeVar
```
jkaster: Is this to make those collections read-only? Seems like a good approach.

joeldodge79: Unfortunately this is complicated... here's a verbose explanation (mostly for my own benefit):

I originally thought I needed `MutableSequence` because it inherits `Sequence` methods and adds `append`, `reverse`, `extend`, `pop`, `remove` and I'd wanted to be flexible to the end user as to how they constructed it. But most of the API models that have list-like properties are typed Sequence so you can't just do

```python
sdk.all_groups(ids=models.DelimSequence(user.group_ids))
```

because `user.group_ids` is of type `Sequence` which is a super type of `MutableSequence` so you get

```
Argument 1 to "DelimSequence" has incompatible type "Sequence[int]"; expected "MutableSequence[int]
``` 

But now I realize that `collections.UserList` gives me those mutation methods methods (unfortunately I have to override them to get typing enforcement) and it's ok to just use `Sequence[T]`

src/python.gen.ts
joeldodge79: I think `black` is only called once and the `*.py` in the path argument is needed to tell it what files to run on. I'd thought it only needed a directory but maybe that changed because it wasn't formatting files. Here's the output:

```
Locking [dev-packages] dependencies…
Locking [packages] dependencies…
⠇ Locking..✔ Success!
Updated Pipfile.lock (595abd)!
reformatted /Users/joeldodge/Documents/sdk_codegen/python/looker_sdk/sdk/methods.py
reformatted /Users/joeldodge/Documents/sdk_codegen/python/looker_sdk/sdk/models.py
All done! ✨ 🍰 ✨
2 files reformatted, 1 file left unchanged.
```

(the `1 file left unchanged` is `python/looker_sdk/sdk/__init__.py` which is empty)

jkaster: `reformatFile()` is called for each source file, currently. We could refactor for the Python scenario of a folder.

```typescript
// Reformat source files after generation
  reformat() {
    const result: string[] = []
    let files = ['sdk/methods', 'sdk/models']
    if (this.willItStream) files.push('sdk/streams')
    for (const name of files) {
      const sourceFile = this.fileName(name)
      const output = this.reformatFile(sourceFile)
      if (output) {
        result.push(output)
      }
    }
    return result
  }
```

joeldodge79: oh geez, I totally missed that. I'll update

PR #83, opened 2019-11-04
Kotlin Alpha SDK spike

jkaster:
Snuck in some Swift fixes here too, because they were similar issues.

Includes the Kotlin RTL, generator, tests, and readme


kotlin/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+When the SDK is installed and the server location and API credentials are configured in your `looker.ini` file, it's ready to be used.
+
+Verify authentication works and that API calls will succeed with code similar to the following:
+
+```kotlin
+val localIni = "/Users/Looker/sdk-codegen/looker.ini"
+val settings = ApiSettingsIniFile(localIni, "Looker")
+
+val client: HttpClient = HttpClient(Apache) {
```
brianmadden: This is only necessary when validating against a Looker install with a self-signed cert. Do our customer instances have non-self-signed certs? Regardless, I think it's probably worth noting here that the `TrustSelfSignedStrategy` and `NoopHostnameVerifier` are dangerous and that they likely _should not_ end up in anything that is production level code.

kotlin/src/main/com/looker/rtl/Transport.kt
```diff
@@ -0,0 +1,202 @@
...
+            }
+        } + addQueryParams(path, queryParams)
+    }
+
+    inline fun <reified T> request(
+            method: HttpMethod,
+            path: String,
+            queryParams: Values = mapOf(),
+            body: Any? = null,
+            noinline authenticator: Authenticator? = null): SDKResponse {
```
brianmadden: Why did we change the default here from `Authenticator = ::defaultAuthenticator` to make it nullable? Is this due to the needing an `Optional` class to wrap nullable fields?

kotlin/src/test/KotlinExample.kt
```diff
@@ -0,0 +1,109 @@
...
+import org.apache.http.ssl.SSLContextBuilder
+import java.io.File
+import java.lang.Thread.sleep
+
+class KotlinExample {
+    val localConfig = "/Users/Looker/Documents/sdk_codegen/looker.ini"
+    val settings = ApiSettingsIniFile(localConfig, "Looker")
+
+    /**
+     * Note: This is only to work with self-signed certs... this should probably be removed before this is posted anywhere
```
brianmadden: Probably worth changing this comment to specify that the pattern should not be used in production code instead. :D

kotlin/src/test/TestUserSession.kt
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.http.conn.ssl.TrustSelfSignedStrategy
+import org.apache.http.ssl.SSLContextBuilder
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+import org.junit.Test as test
+
+
+class TestUserSession {
+//    val localIni = "/Users/looker/Documents/sdk_codegen/looker.ini"
+    val localIni = "/Users/looker/sdk-codegen/looker.ini"
```
brianmadden: TODO: We should figure out a more elegant way to do this for testing.

jkaster: Yup. I've got an updated comment on that somewhere I'm surprised isn't showing up in this PR yet. Lots of cleanup remaining. May even get to it later today after this spike lands.

kotlin/src/test/KotlinExample.kt
jkaster: Changed

kotlin/src/main/com/looker/rtl/Transport.kt
jkaster: I've added a TODO

kotlin/README.md
jkaster: Thanks, replaced this code sample

PR #84, opened 2019-11-04
Added a python test for running looks in csv, json, and json_label format

jkaster:
Test showing three different render outputs for running a look

### DIFF
--- a/python/Pipfile
+++ b/python/Pipfile
@@ -20,7 +20,6 @@ pytest-pudb = "*"
 pytest-mock = "*"
 twine = "*"
 pillow = "*"
-v = {editable = true,version = "*"}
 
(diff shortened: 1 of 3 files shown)


python/Pipfile
```diff
@@ -20,6 +20,7 @@ pytest-pudb = "*"
 pytest-mock = "*"
 twine = "*"
 pillow = "*"
+v = {editable = true,version = "*"}
```
joeldodge79: was this intentional?

jkaster: Something stuck it in so I thought it should belong. I can remove it

joeldodge79: `pipenv uninstall v --dev`

Issue #71, opened 2019-10-17
Python: Modify setup.py to prevent installation on unsupported python versions 


josephaxisa:
Add `python_requires=">=3.7"` to `setup.py` to prevent installation on python versions older than 3.7.

joeldodge79:
actually, until cattrs supports 3.8 we need to pin this to 3.7 :-(

joeldodge79:
let's also add better keywords for SEO

PR #92, opened 2019-11-11
pin to python 3.7.x and add keywords

josephaxisa:


python/setup.py
```diff
@@ -17,10 +17,11 @@
     license="MIT",
     long_description=open("README.rst").read(),
     long_description_content_type="text/x-rst",
-    keywords=["looker_sdk", "Looker API 3.1"],
+    keywords=["Looker", "Looker API", "looker_sdk", "Looker API 3.1"],
```
jkaster:
```suggestion
    keywords=["Looker", "Looker API", "looker_sdk", "Looker SDK", "SDK", "Looker API 3.1"],
```

jkaster: Only accept the additional tags if you think they make sense

python/setup.py
```diff
@@ -17,10 +17,11 @@
...
     long_description=open("README.rst").read(),
     long_description_content_type="text/x-rst",
-    keywords=["looker_sdk", "Looker API 3.1"],
+    keywords=["Looker", "Looker API", "looker_sdk", "Looker SDK", "SDK", "Looker API 3.1"],
     name=NAME,
     package_data={"looker_sdk": ["py.typed", "looker_sdk/looker-sample.ini"]},
     packages=find_packages(),
+    python_requires=">=3.7.0, <3.8",
     url="https://pypi.python.org/pypi/looker_sdk",
     version=VERSION,
```
joeldodge79: let's bump the version and ship a new package

josephaxisa:
Closes #71

Issue #81, opened 2019-11-01
All: Request methods should support a fully specified url


jkaster:
Our `request` methods should work for both our API/SDK calls and calls to other HTTP resources. There are 3 modes:

- straight request path passthrough when `http://` or `https://` starts the request
- api path + relative path munging when the authenticator callback is passed in, meaning it's an API call
- base url + relative path when no authenticator is used

This will allow SDK users to *also* issue HTTP requests using our handy request methods for resources that are *not* in our SDK. Great additional functionality for a trivial level of effort on our part. 

- [x] Do for Typescript
- [ ] Do for Python - update: [more work required](https://github.com/looker-open-source/sdk-codegen/pull/97#issuecomment-555046500)
- [x] Do for Swift
- [x] Do for Kotlin

jkaster:
@joeldodge79 I believe you said our Python SDK already does this? If so, please feel free to check the box!

joeldodge79:
here's the typescript test for this: https://github.com/looker-open-source/sdk-codegen/blob/b6852f19325ce5705518b8337df7efa08696fb5d/typescript/looker/rtl/nodeTransport.spec.ts#L38-L48

jkaster:
Created #116 for the Python SDK. Closing this for the other SDKs. I'll avoid creating an "all sdk" issue in the future for better tracking and management, unless it applies specifically to codegen.

PR #97, opened 2019-11-12
Python SDK: support fully specified urls

josephaxisa:


python/looker_sdk/rtl/requests_transport.py
```diff
@@ -104,6 +103,28 @@ def request(
...
+    def make_path(
+        self,
+        path: str,
+        query_params: Optional[MutableMapping[str, str]],
+        authenticator: Optional[Callable[[], Dict[str, str]]] = None,
+        transport_options: Optional[transport.TransportSettings] = None,
+    ) -> str:
+        base = self.api_path if authenticator else transport_options.base_url
+        if not path.startswith(("https:", "http:")):
+            path = f"{base}{path}"
```
joeldodge79: I would signal from whether they provide an "absolute url". something like:

```python
parsed = urllib.parse.urlparse(path)
if parsed.scheme:
   ret = path
else:
   ret = f"{self.api_path}{path}"
```

python/looker_sdk/rtl/requests_transport.py
```diff
@@ -64,31 +65,29 @@ def request(
         headers: Optional[MutableMapping[str, str]] = None,
         transport_options: Optional[transport.TransportSettings] = None,
     ) -> transport.Response:
-
-        url = f"{self.api_path}{path}"
```
joeldodge79: I think maybe this whole change could boil down to the following (provided we're ok with continuing to let `requests` handle the `query_params` rather than doing it ourselves)

```python
url = path if urllib.parse.urlparse(path).scheme else f"{self.api_path}{path}"
```

josephaxisa: This is great, nice catch!

python/looker_sdk/rtl/requests_transport.py
```diff
@@ -104,6 +102,14 @@ def request(
 
         return ret
 
+    def make_path(
+        self, path: str, authenticator: Optional[Callable[[], Dict[str, str]]] = None
+    ) -> str:
+        base = self.api_path if authenticator else self.settings.base_url
```
joeldodge79: I'm pretty sure we don't want to use the presence/absence of `authenticator` to influence the url at all.

python/looker_sdk/rtl/requests_transport.py
```diff
@@ -64,20 +65,19 @@ def request(
         headers: Optional[MutableMapping[str, str]] = None,
         transport_options: Optional[transport.TransportSettings] = None,
     ) -> transport.Response:
-
-        url = f"{self.api_path}{path}"
+        request_path = self.make_path(path, authenticator)
```
joeldodge79: I don't think we need a separate function. https://refactoring.com/catalog/inlineFunction.html used to have more explanation but basically it's easier to just read

```suggestion
        url = path if urllib.parse.urlparse(path).scheme else f"{self.api_path}{path}"
```

inline then have to context switch to read the implementation of `make_path()`.

for sure we want to break up longer functions where it makes sense (and this function is definitely at my comfort edge of length) but because this logic only happens once and it's essentially one line it's better left inline.

python/tests/rtl/test_requests_transport.py
```diff
@@ -153,7 +162,38 @@ def test_request_error(settings):
...
+
+
+def test_full_path_without_auth_is_just_full_path(rtp):
+    actual = rtp.make_path("https://looker.sdk/login")
+    assert actual == "https://looker.sdk/login"
+
+
+def test_full_path_with_auth_is_just_full_path(rtp):
+    actual = rtp.make_path("https://looker.sdk/login", mock_auth)
+    assert actual == "https://looker.sdk/login"
```
joeldodge79: I think this is testing too deeply into the internals of the implementation (especially since I think `make_path()` should go away). I think the `test_retrieves_fully_qualified_url` is sufficient.

python/looker_sdk/rtl/requests_transport.py
jkaster: My intent was that the authenticator is used as a flag to determine whether api_path was used as the base for relative paths, or whether base_url was. The thinking was, if the call needed to be authenticated, it's an API call.

python/looker_sdk/rtl/requests_transport.py
jkaster: Just last night I was thinking that "make path" is the wrong function name and I intend to change it for the Typescript implementation. It's really "make url." The function takes the parameters, then provides the fully specified URL as output. That's the intent. It makes things like proxy sessions much easier to accomplish. The Typescript branch I've updated last night shows this.

python/looker_sdk/rtl/requests_transport.py
joeldodge79: I could imagine someone wanting to provide a random `authenticator` callback to some fully qualified path they've set.

On a related, higher level note: the python transport interface is pretty limited. The return is just `bytes` (following @dthorpe's suggestion about leaving the response alone and letting calling code do what it wants to it). The serialization layer is separate (and only called in `api_methods.py`). So I actually don't think our transport is very helpful. If I were an app developer, I'd just import `requests` directly and have a lot easier time of it thank using our transport :-)

python/looker_sdk/rtl/requests_transport.py
joeldodge79: @jkaster I have a feeling you and I should talk about what the desired behavior is :-) I don't fully understand the motivations to make the transport layer directly usable by sdk callers, at least how it's implemented in python currently

python/looker_sdk/rtl/requests_transport.py
jkaster: Fully qualified path + authenticator should still work. Good point about the python transport interface limitations. We'll need to discuss that disparity with the other SDKs, because the other SDKs do have the ability to return strongly typed responses back from a given request.

joeldodge79:
@josephaxisa - @jkaster, @dthorpe and I discussed this and, as I suspected, it the desired feature/behavior isn't so trivial to expose in the Python SDK. They'd like to expose our "transport + deserialization" layer to sdk users: a caller can provide a structure/type and a URL and get back the strongly typed object.

It may not require too much work to implement but would definitely take some thought and design (and I'd like to revisit our typing too). We all agreed it's a "nice to have" but not a priority right now. The other SDKs get it almost for free as a byproduct of the tighter coupling between transport and deserialization that the languages natively support. Python doesn't have this so we have to do some extra work :-)

Let's close this PR for now. I've updated the parent issue #81 to capture this.

PR #105, opened 2019-11-14
Python SDK: Add constructors for all input models

joeldodge79:
The logic for when to add a constructor to a model was flawed - "Write"
types are just copies of Models that have at least one read-only field.
Not every method gets a separate "Write" model.

This change also captures nested model types for input models to
generate constructors for those as well.

Another option would be to just generate a constructor for all models
but this way we at least still keep the code size down and once the IDE
support is there for attrs auto-generated constructors we can ditch them
all together.


python/tests/integration/test_methods.py
```diff
@@ -291,6 +291,7 @@ def test_search_looks_title_fields_filter(sdk: mtds.LookerSDK):
     assert "Order" in look.title
     assert look.description is None
 
+
 def test_search_look_and_run(sdk: mtds.LookerSDK):
```
jkaster: I don't see a test that exercises the new `__init__` generation support.

src/python.gen.ts
```diff
@@ -252,7 +265,7 @@ ${this.hooks.join('\n')}
    */
   construct(indent: string, type: IType) {
     // Skip read-only parameters
-    if (!(type instanceof WriteType)) return ''
+    if (!this.inputTypes.has(type)) return ''
     indent = this.bumper(indent)
     const bump = this.bumper(indent)
     let result = `\n\n${indent}def __init__(self, *, `
```
jkaster: Maybe we should have this line wrap and indent so the first "real" parameter is indented with the rest of the parameters?

src/python.gen.ts
```diff
@@ -376,7 +389,7 @@ ${this.hooks.join('\n')}
     }
 
     let attrsArgs = 'auto_attribs=True, kw_only=True'
-    if (type instanceof WriteType) {
+    if (this.inputTypes.has(type)) {
```
jkaster: I haven't grasped exactly what `inputTypes` does yet. Does the collection grow as as each method's parameters are added, and the constructor for a given method isn't required if the `inputType` has already been encountered?

jkaster: Oh .. is `inputTypes` so `__init__` is only generated for those models requiring it?

python/tests/integration/test_methods.py
joeldodge79: there's actually no runtime behavior difference. All the models get an `__init__` as `@attrs` will auto-generate one for you unless you tell it not to. so then it becomes a "meta" level test of what does the written code look like and I added that to the codegen side

src/python.gen.ts
joeldodge79: yeah, that would make the tests look prettier. I think I tried once and failed but I can try again. Once `black` reformats the file it all looks good so we ship beautiful code :-)

src/python.gen.ts
joeldodge79: yes, sorry if that wasn't clear - I can change the name to `methodInputModelTypes` or some such (and/or leave a code comment). I could have gone the route of manually generating constructors for all the models but I wanted to limit boilerplate/bloat as much as possible and at some point we can undo this when the code editors catch up with the `@attrs` auto-generated `__init__`s

python/tests/integration/test_methods.py
joeldodge79: haha, this change is because I had a local test failure due to missing data in my helltool. when I saved this file `black` reformatted it to add the blank line

src/python.gen.spec.ts
```diff
@@ -292,47 +292,94 @@ class ApiVersion(model.Model):
...
+    source_field_name: Optional[str] = None
+
+    def __init__(self, *,
+            field_name: Optional[str] = None,
+            source_field_name: Optional[str] = None):
+        self.field_name = field_name
+        self.source_field_name = source_field_name`)
     })
   })
 })
```
josephaxisa: Might need to add a new line at the end of the file.

src/python.gen.ts
jkaster: Thanks for confirming.

src/python.gen.spec.ts
joeldodge79: thanks! I started using intellij for typescript and it doesn't add a newline at end of file by default. I just foudn the setting (searching for "end of file" in settings)

PR #110, opened 2019-11-15
use x-looker-appid for user agent tag

### LINKED ISSUE #99: Python: use x-looker-appid for the user agent tag
Even though we may be able to set `User-Agent` for a given request library, we should set the Looker-specific header `x-looker-appid` for our SDK to make sure the SDK version is recognized by Looker because some request libraries (and browsers) will override the `User-Agent` header.

josephaxisa:
resolves #99


josephaxisa:
@joeldodge79 I believe https://github.com/looker-open-source/sdk-codegen/blob/e7332cc72f7718fd3cde69288f20049a74051be3/python/looker_sdk/rtl/requests_transport.py#L50 is redundant. Am I wrong?

Issue #104, opened 2019-11-14
Kotlin: verify api_version defaults to 3.1


jkaster:
If the api version isn't specified in the configuration, it should default to our active supported version, which is currently 3.1

jkaster:
Closed with #112

PR #112, opened 2019-11-16
Closing some quick Kotlin SDK issues

### LINKED ISSUE #100: Kotlin: set x-looker-appid for the agent tag
Even though we may be able to set `User-Agent` for a given request library, we should set the Looker-specific header `x-looker-appid` for our SDK to make sure the SDK version is recognized by Looker because some request libraries (and browsers) will override the `User-Agent` header.

jkaster:
- Additional API configuration settings supported
- Transport support for `verify_ssl` and `timeout` still pending
- Closes #100, #104

### DIFF
--- a/kotlin/src/main/com/looker/rtl/Constants.kt
+++ b/kotlin/src/main/com/looker/rtl/Constants.kt
@@ -6,6 +6,8 @@ import org.jetbrains.annotations.NotNull
 const val LOOKER_VERSION = "6.21"
 const val API_VERSION = "3.1"
 const val SDK_VERSION = "${API_VERSION}.${LOOKER_VERSION}"
+const val AGENT_TAG = "SDK-KT ${SDK_VERSION}"
+const val LOOKER_APPID = "x-looker-appid"
 const val ENVIRONMENT_PREFIX = "LOOKERSDK"
 
 const val MATCH_CHARSET = ";.*charset="
--- a/kotlin/src/main/com/looker/rtl/Constants.kt
+++ b/kotlin/src/main/com/looker/rtl/Constants.kt
@@ -30,12 +30,12 @@ class DelimArray<T> : Array<T>() {
 
 fun isTrue(value: String?) : Boolean {
     val low = value?.toLowerCase()
-    return low === "true" || low === "1" || low === "t" || low === "y" || low === "yes"
+    return low == "true" || low == "1" || low == "t" || low == "y" || low == "yes"
 }
 
 fun isFalse(value: String?) : Boolean {
     val low = value?.toLowerCase()
-    return low === "false" || low === "0" || low === "f" || low === "n" || low === "no"
+    return low == "false" || low == "0" || low == "f" || low == "n" || low == "no"
 }
 
 fun asBoolean(value: String?) : Boolean? {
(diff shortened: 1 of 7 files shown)


kotlin/src/main/com/looker/rtl/Constants.kt
```diff
@@ -6,13 +6,18 @@ import org.jetbrains.annotations.NotNull
 const val LOOKER_VERSION = "6.21"
 const val API_VERSION = "3.1"
 const val SDK_VERSION = "${API_VERSION}.${LOOKER_VERSION}"
+const val AGENT_TAG = "SDK-KT ${SDK_VERSION}"
```
josephaxisa: This is slightly different than what we did in the TS and Python implementations, where the agent tag was of the form `LANG-SDK {SDK_VERSION}`. 
```suggestion
const val AGENT_TAG = "KT-SDK ${SDK_VERSION}"
```

kotlin/src/main/com/looker/rtl/Transport.kt
```diff
@@ -164,18 +166,17 @@ class Transport(val options: TransportSettings) {
...
 
         val builder = HttpRequestBuilder()
         // Set the request method
         builder.method = method.value
 
         // Handle the headers
-        val agentTag = "LookerSDK Kotlin ${options.apiVersion}"
         val headers = options.headers.toMutableMap()
-        headers["User-Agent"] = agentTag
+        headers["User-Agent"] = AGENT_TAG
```
josephaxisa: I wasn't aware that we should leave the "User-Agent" tag in the headers now that we've added the "x-looker-appid" tag. I can change this in the python SDK if it's intended.

kotlin/src/main/com/looker/rtl/Constants.kt
jkaster: I had a conversation with @dthorpe this weekend about how to tag the SDKs. We're going to end up with something like:
`Looker SDK-XX 3.1.6.24` where `XX` is the platform specifier, `3.1` is the SDK version, and `6.24` is the Looker release version the SDK was generated from.
We may support additional attributes to identify the "app" making an API call, but that's our minimum.

kotlin/src/main/com/looker/rtl/Transport.kt
jkaster: It doesn't hurt to set `User-Agent` and it may end up preserved in some scenarios. `x-looker-appid` is something that should be preserved throughout the request lifecycle.

kotlin/src/main/com/looker/rtl/Constants.kt
```diff
@@ -23,6 +28,22 @@ class DelimArray<T> : Array<T>() {
 }
  */
 
+fun isTrue(value: String?) : Boolean {
+    val low = value?.toLowerCase()
+    return low === "true" || low === "1" || low === "t" || low === "y" || low === "yes"
```
brianmadden: I think the use of referential equality here may be a bit dangerous. If we're the ones instantiating the strings they should point to the same object, but if we're receiving this string from elsewhere, such as over the wire, they'll be different. This, for example, will return false:
```
"a" === buildString { "a" } 
```

Unless we absolutely want to ensure that these are the same object I'd relax this to `==`

kotlin/src/main/com/looker/rtl/Constants.kt
```diff
@@ -23,6 +28,22 @@ class DelimArray<T> : Array<T>() {
...
+}
+
+fun isFalse(value: String?) : Boolean {
+    val low = value?.toLowerCase()
+    return low === "false" || low === "0" || low === "f" || low === "n" || low === "no"
```
brianmadden: Same here, I'd relax these to `==`.

jkaster: Thanks. Fixed

kotlin/src/main/com/looker/rtl/Constants.kt
```diff
@@ -30,12 +30,12 @@ class DelimArray<T> : Array<T>() {
 
 fun isTrue(value: String?) : Boolean {
     val low = value?.toLowerCase()
-    return low === "true" || low === "1" || low === "t" || low === "y" || low === "yes"
+    return low == "true" || low == "1" || low == "t" || low == "y" || low == "yes"
```
dthorpe: Sorry, late to the party.  Why are we allowing/accepting loose boolean values?

jkaster: For implementation consistency with Python and Typescript's config processing

PR #124, opened 2019-11-27
Add transport and session implementations for the extension SDK

bryans99:
The SDK remains decoupled from both the extension and chatty packages. The extension SDK will provide an implementation of a new interface exported by this change. Note streams are not currently supported. Note the interface could be implemented by some other package so it begs the question as to whether extension is a good name. On the other hand, I can't see anything else wanting to use the new interface so I think the extension prefix is reasonable as a name.


typescript/looker/rtl/extensionSession.ts
```diff
@@ -0,0 +1,68 @@
...
+
+import { IApiSettings } from './apiSettings'
+import { ITransport, IRequestProps } from './transport'
+import { IAuthSession } from './authSession'
+
+export class ExtensionSession implements IAuthSession {
+  sudoId: string = ''
+  transport: ITransport
+
+  constructor(public settings: IApiSettings, transport: ITransport) {
```
jkaster: minor typescript syntax sugar you may already know about, you could do

```typescript
constructor(public settings: IApiSettings, public transport: ITransport)
```
instead of explicitly declaring `transport` above. 

That also reminds me I want to find out if the explicit setters for `this.settings` and `this.transport` are even required if the property scope is declared in the constructor signature.

typescript/looker/rtl/extensionTransport.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+import { ExtensionTransport } from './extensionTransport'
+import { IHostConnection } from './extensionSdk'
+
+describe('ExtensionTransport', () => {
+  const queryParams = {a:"b c", d: false, nil: null, skip: undefined}
+  const body = {song: 'how much is that doggy'}
+  let transport: ExtensionTransport
+  let connection: IHostConnection
+  let mockRequest: any
+  const streamCallback = (readable: any) => { return new Promise((resolve, reject) => reject())}
```
jkaster: This looks like a good approach because our streaming pattern isn't fully baked yet, anyway.

typescript/looker/rtl/extensionTransport.ts
```diff
@@ -0,0 +1,81 @@
...
+    authenticator?: any,
+    options?: Partial<ITransportSettings>,
+  ): Promise<SDKResponse<TSuccess, TError>> {
+    let params: any
+    if (queryParams) {
+        params = {}
+        Object.keys(queryParams).forEach(key => {
+            if (queryParams[key]) {
+                params[key] = queryParams[key]
+            }
```
jkaster: are your parameter passing constraints intentionally different than `encodeParams` in `transport.ts`? That routine is a bit more hardened, with encoding of query parameters, etc, and explicit checks only against `undefined` because this loop will omit parameter values you may not actually want to be omitted if they evaluate to "falsey."

typescript/looker/rtl/extensionTransport.ts
```diff
@@ -0,0 +1,81 @@
...
+        })
+    }
+
+    return this.hostConnection.request(
+        method,
+        path,
+        params,
+        body,
+        undefined,
```
jkaster: do you want to throw an error if the user *does* pass in an "authenticator" callback your transport deliberately ignores?

typescript/looker/rtl/extensionSession.ts
bryans99: I just did a straight copy from  the extension SDK and it had some "fixes" to get it to compile after the last core SDK change. Now that the class is in the coreSDK I can see it needs a little more adjustment. Will push that change soon.

typescript/looker/rtl/extensionTransport.ts
bryans99: I think thats a good idea

typescript/looker/rtl/extensionTransport.ts
bryans99: I've actually been thinking about this code and have come to the conclusion that it should probably be in helltool. This came about in the rush to get things ready for JOIN. Basically the host was failing if it received queryParams with undefined values. This was a quick and dirty to get Wil unblocked.

typescript/looker/rtl/extensionSession.ts
bryans99: Fixed as per your suggestion

typescript/looker/rtl/extensionTransport.ts
bryans99: Moved to helltool and removed from here

typescript/looker/rtl/extensionTransport.ts
bryans99: Decided that at some point authenticator may be needed so now its passed through. The extension SDK will ignore it for now

PR #130, opened 2019-12-10
Fix typescript test failures when tests run from root directory

bryans99:
Test failures were caused by tslint issues. When the tests were added, the tests were run from the typescript/looker directory and worked. This is because the tsconfig at root and in typescript/looker were different. This change makes an attempt at reconciling the differences. The format of the two tsconfig files are now similar so they can be compared side by side. Matching options are at the top, options with different values next, followed lastly by different options.

The options causing the test failures were:
    "noUnusedLocals": true,
    "noUnusedParameters": true,
These were not in the tsconfig in typescript/looker. Fixing not an option as this caused the typescript/looker build to fail. Also tried tslint disable but for some reason jest did not honour these. In the end removed those options in the root tsconfig to bring into line with typescript/looker tsconfig. yarn test from root no longer fail due to typescript issues.

Note that yarn lint fails from root (fine in typescript/looker). sdkConfig.ts has this line which fails with ERROR: 45:9  strict-type-predicates  Expression is always false.
 if (props.verify_ssl === undefined) props.verify_ssl  = true
I suspect this is because verify_ssl is a required property and as such should never be undefined. The code seems a little odd anyway but I would prefer to fix it in a separate PR.

### DIFF
--- a/tsconfig.common.json
+++ b/tsconfig.common.json
@@ -1,0 +1,18 @@
+{
+  "compilerOptions": {
+    "declaration": true,
+    "module": "commonjs",
+    "moduleResolution": "node",
+    "noImplicitReturns": true,
+    "noImplicitAny": true,
+    "noFallthroughCasesInSwitch": true,
+    "strict": true,
+    "esModuleInterop": true,
+    "plugins": [
+      {
+        "name": "tslint-language-service",
+        "alwaysShowRuleFailuresAsWarnings": true
+      }
+    ]
+  }
+}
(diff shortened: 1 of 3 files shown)


dthorpe:
Why do we have 2 tslint config files?  Do they need to be different?  If they can be consolidated, there's no chance of them getting out of sync again in the future.

jkaster:
One of the tslint config files is in the root of the codegen repo. The other is for the typescript packager in `typescript/looker`. I was just discussing with @lukelooker how we can share or inherit one from the other. I would rather go that route, or create a symlink, and also enforce the same linting rules for both paths of validating Typescript.

bryans99:
Just pushed another change that moves common configs to tsconfig.common.json.

It might be considered a start but I think there's more to do.

Issue #131, opened 2019-12-11
Python: Allow ini variables to be quoted


DucarrougeR:
testing the following code to test authentication in Python

```python
from looker_sdk import client, models, error
sdk = client.setup("looker.ini")
looker_api_user = sdk.me()
```

gave the following error:

```
File "C:\Python\venv\looker\lib\site-packages\looker_sdk\rtl\auth_session.py", line 82, in _get_admin_token
    self._login_admin()
  File "C:\Python\venv\looker\lib\site-packages\looker_sdk\rtl\auth_session.py", line 147, in _login_admin
    headers={"Content-Type": "application/x-www-form-urlencoded"},
  File "C:\Python\venv\looker\lib\site-packages\looker_sdk\rtl\auth_session.py", line 222, in _ok
    raise error.SDKError(response.value)
looker_sdk.error.SDKError: b'{"message":"Not found","documentation_url":"http://docs.looker.com/"}'
```

This was due to the values for the variables in the looker.ini file being quoted.

It would be great to either account for this option where values are quoted or to have a more descriptive message for these scenarios.

jkaster:
We should unquote any string configuration variables automatically before processing them, including environment variables

PR #136, opened 2019-12-13
Python SDK: Allow quoted string config vars

josephaxisa:
This addresses #131

### DIFF
--- a/python/looker_sdk/rtl/api_settings.py
+++ b/python/looker_sdk/rtl/api_settings.py
@@ -95,6 +95,8 @@ class ApiSettings(transport.TransportSettings):
         if env_verify_ssl:
             config_data["verify_ssl"] = env_verify_ssl
 
+        config_data = cls._clean_input(config_data)
+
         if not config_data.get("base_url"):
             raise error.SDKError(f"Required parameter base_url not found.")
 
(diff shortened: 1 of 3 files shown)


python/looker_sdk/rtl/api_settings.py
```diff
@@ -114,10 +116,19 @@ def read_ini(filename: str, section: Optional[str] = None) -> Dict[str, str]:
...
+    def _clean_input(cls, config_data: Dict[str, str]) -> Dict[str, str]:
+        for setting, value in list(config_data.items()):
+            # Remove empty setting values
+            if not isinstance(value, str):
+                continue
+            if value in ['""', "''", ""]:
+                config_data.pop(setting)
+            # Strip quotes from setting values
+            elif value.startswith(('"', "'")) or value.endswith(('"', "'")):
+                config_data[setting] = value.strip("\"'")
```
jkaster: You only want to strip beginning and ending quote values. Does `value.strip()` remove all occurrences?

python/tests/integration/test_methods.py
```diff
@@ -296,7 +296,7 @@ def test_search_look_and_run(sdk: mtds.LookerSDK):
     """run_look() should return CSV and JSON
     CSV will use column descriptions
     JSON will use column names
-    JSON_LABEL will use column descriptions
+    JSON_DETAIL will use column descriptions
```
jkaster: Did the behavior of JSON_LABEL change? Wasn't this test intended to verify the behavior of the new format, JSON_LABEL?

python/looker_sdk/rtl/api_settings.py
josephaxisa: It only removes leading/trailing occurrences: https://docs.python.org/3/library/stdtypes.html#str.strip

jkaster: Ah. Most languages I know call that `trim` of some sort. Thanks.

python/tests/integration/test_methods.py
josephaxisa: You're right. I just reverted this change.

PR #138, opened 2019-12-13
Python SDK: crud tests for looks

josephaxisa:
- Added a test that creates a test look using data.yml
- A fixture that deletes the test look before the crud tests run

### DIFF
--- a/python/tests/conftest.py
+++ b/python/tests/conftest.py
@@ -17,6 +17,11 @@ def dashboards(_test_data):
 
 
 @pytest.fixture(scope="session")
+def looks(_test_data):
+    return _test_data["looks"]
+
+
+@pytest.fixture(scope="session")
 def queries(_test_data) -> List[Dict[str, str]]:
     return _test_data["queries"]
 
@@ -51,6 +56,15 @@ def remove_test_dashboards(sdk: mtds.LookerSDK, dashboards):
         if len(search_results) > 0:
             for dashboard in search_results:
                 sdk.delete_dashboard(cast(str, dashboard.id))
+
+
+@pytest.fixture(scope="function")
+def remove_test_looks(sdk: mtds.LookerSDK, looks):
+    for l in looks:
+        search_results = sdk.search_looks(title=l["title"])
+        if len(search_results) > 0:
+            for look in search_results:
+                sdk.delete_look(cast(int, look.id))
 
 
 @pytest.fixture(name="test_users", scope="session")
--- a/python/tests/integration/test_methods.py
+++ b/python/tests/integration/test_methods.py
@@ -267,9 +267,9 @@ def test_crud_look(sdk: mtds.LookerSDK, looks):
         assert isinstance(look, ml.LookWithQuery)
         assert look.title == l.get("title")
         assert look.description == l.get("description")
-        assert look.deleted == l.get("deleted", False)
-        assert look.is_run_on_load == l.get("is_run_on_load", True)
-        assert look.public == l.get("public", False)
+        assert look.deleted == l.get("deleted")
+        assert look.is_run_on_load == l.get("is_run_on_load")
+        assert look.public == l.get("public")
         assert look.query_id == query.id
         assert look.space_id == l.get("space_id") or sdk.me().home_space_id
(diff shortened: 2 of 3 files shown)


python/tests/integration/test_methods.py
```diff
@@ -254,9 +245,47 @@ def test_it_runs_inline_query(sdk: mtds.LookerSDK, queries: TQueries):
...
+                deleted=l.get("deleted"),
+                is_run_on_load=l.get("is_run_on_load"),
+                public=l.get("public"),
+                query_id=query.id,
+                space_id=l.get("space_id") or str(sdk.me().personal_space_id),
+            )
+        )
+
+        assert isinstance(look, ml.LookWithQuery)
+        assert look.title == l.get("title")
```
jkaster: Is `obj.get("...")` the reference syntax for accessing properties of an object? That seems to be an anti-pattern

python/tests/conftest.py
```diff
@@ -53,6 +58,15 @@ def remove_test_dashboards(sdk: mtds.LookerSDK, dashboards):
                 sdk.delete_dashboard(cast(str, dashboard.id))
 
 
+@pytest.fixture(scope="function")
+def remove_test_looks(sdk: mtds.LookerSDK, looks):
+    for l in looks:
+        search_results = sdk.search_looks(title=l["title"])
+        if len(search_results) > 0:
+            for look in search_results:
+                sdk.delete_look(cast(int, look.id))
```
jkaster: If we add an additional look to data.yml, will this loop actually work to match all looks and delete them?

python/tests/integration/test_methods.py
josephaxisa: `obj.get("property")` returns `None` if no second argument is provided. If we did `obj["property"]` we'd get a `KeyError` when the property doesn't exist

josephaxisa: At the same time, some of these properties are required so perhaps they should be accessed using the latter

python/tests/conftest.py
josephaxisa: Yes. `looks` on line 62 is a list containing all looks defined in `data.yml`. As long as each look has a title, we'll be able to find it and delete it.

python/tests/integration/test_methods.py
jkaster: good note about the accessor error. I didn't consider that. I think for the required fields we should use `foo.bar` notation unless it becomes a hassle to check the property status.

python/tests/conftest.py
jkaster: sorry, I missed the iterator on line 63!

python/tests/integration/test_methods.py
josephaxisa: fwiw, we used the `.get` notation everywhere else throughout the tests.  In this case, `looks` is a list of dictionaries. The notation we can use is either `obj[key]` or `obj.get(key[, default])` (https://docs.python.org/3/library/stdtypes.html?highlight=dict%20get#dict.get). It wouldn't introduce any hassle to use `obj[key]` but I might need to go back and change all the other tests for consistency if we're going down this route. Based on the docs the only required parameters for `create_look` are `model` and `view` in `query`.

jkaster: I think `obj[key]` is slightly better than `obj.get(key...)` unless the semantics of failed gets changes

python/tests/integration/test_methods.py
```diff
@@ -254,9 +245,47 @@ def test_it_runs_inline_query(sdk: mtds.LookerSDK, queries: TQueries):
...
+        assert look.description == l.get("description")
+        assert look.deleted == l.get("deleted", False)
+        assert look.is_run_on_load == l.get("is_run_on_load", True)
+        assert look.public == l.get("public", False)
```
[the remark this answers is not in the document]
jkaster: I like making the yaml more explicit

python/tests/integration/test_methods.py
jkaster: Separate PR FTW

josephaxisa: Perfect, agreed. I'll leave these as is for now and create a separate PR first chance I get.

Issue #79, opened 2019-11-01
Swift: Create generic key/value collection encoder


jkaster:
Currently query param encoding is buried inside a routine that adds them to the path. It needs to be extracted and reusable.

This will be used by #76, and is related to #77 and #78.

jkaster:
Closed with #144

PR #144, opened 2019-12-27
Swift: Requests with bodies now process correctly

### LINKED ISSUE #76: Swift: Verify body arguments are encoded for endpoints appropriately
Write an integration test for creating a query to verify that body parameters are processed correctly. Then test the return results for `csv`, `json`, and `json_label` formats.

jkaster:
Closes #76, #79, #101 

This requires the updated API version that fixes body response encoding

A query unit test was added to show how the query methods can be used


typescript/looker/rtl/constants.ts
```diff
@@ -22,8 +22,8 @@
  * THE SOFTWARE.
  */
 
-export const lookerVersion = '6.24'
-export const apiVersion = '3.1'
+export const lookerVersion = '6.25'
+export const apiVersion = '4.0'
```
dthorpe: intentional?

jkaster: whoops, missed reverting that. Good catch.

PR #148, opened 2020-01-10
Adding typescript docker support

muckweaver:


muckweaver:
@rlgomes this look ok?

docker/typescript/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM base-sdk:latest
+MAINTAINER Rodney Gomes <rodney.gomes@looker.com>
```
rlgomes: that should be your name @muckweaver

muckweaver: I was hoping you'd take the blame on this.

PR #149, opened 2020-01-13
Python SDK: Better support for custom configuration readers

josephaxisa:
To do:
- [x] Fix a few remaining mypy issues
- [ ] Document how one would plugin their own configuration readers with this new implementation
- [x] Get rid of commented out tests once it is confirmed no coverage is lost

### DIFF
--- a/python/looker_sdk/rtl/api_settings.py
+++ b/python/looker_sdk/rtl/api_settings.py
@@ -25,7 +25,7 @@ with the settings as attributes
 """
 import configparser as cp
 import os
-from typing import cast, Dict, MutableMapping, Optional
+from typing import cast, Dict, Optional
 
 import attr
 import cattr
@@ -48,6 +48,9 @@ def _convert_bool(val: str, _: bool) -> bool:
 
 
 class PApiSettings(transport.PTransportSettings, Protocol):
+    filename: str
+    section: Optional[str]
+
     def get_client_id(self) -> Optional[str]:
         ...
 
@@ -56,13 +59,8 @@ class PApiSettings(transport.PTransportSettings, Protocol):
 
 
 @attr.s(auto_attribs=True, kw_only=True)
-class ApiSettings:
-    base_url: str = ""
-    api_version: str = "3.1"
-    verify_ssl: bool = True
-    timeout: int = 120
-    headers: Optional[MutableMapping[str, str]] = None
-    filename: str = ""
+class ApiSettings(transport.TransportSettings):
+    filename: str
     section: Optional[str] = None
 
     @classmethod
@@ -145,6 +143,3 @@ class ApiSettings:
         return os.getenv(
             f"{constants.environment_prefix}_CLIENT_SECRET"
         ) or self.read_config().get("client_secret")
-
-    def is_configured(self) -> bool:
-        return bool(self.base_url and self.api_version)
--- a/python/looker_sdk/rtl/api_settings.py
+++ b/python/looker_sdk/rtl/api_settings.py
@@ -25,10 +25,11 @@ with the settings as attributes
 """
 import configparser as cp
 import os
-from typing import cast, Dict, Optional
+from typing import cast, Dict, MutableMapping, Optional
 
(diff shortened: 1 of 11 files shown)


python/looker_sdk/rtl/api_settings.py
```diff
@@ -46,25 +47,28 @@ def _convert_bool(val: str, _: bool) -> bool:
     return converted
 
 
-@attr.s(auto_attribs=True, kw_only=True)
-class ApiSettings(transport.TransportSettings):
-    """API Configuration Settings.
+class PApiSettings(transport.PTransportSettings, Protocol):
+    def get_client_id(self) -> Optional[str]:
```
joeldodge79: I think we need `section: Optional[str]` here

python/looker_sdk/rtl/transport.py
```diff
@@ -45,8 +46,7 @@ class HttpMethod(enum.Enum):
     HEAD = 7
 
 
-@attr.s(auto_attribs=True, kw_only=True)
-class TransportSettings:
+class PTransportSettings(Protocol):
```
joeldodge79: oh goodness, so I started a deep dive into https://www.python.org/dev/peps/pep-0544/#protocol-members and the next section https://www.python.org/dev/peps/pep-0544/#explicitly-declaring-implementation ...

my basic understanding is that you can actually use protocol classes as baseclasses (**explicit** subtyping) and get the method implementations (e.g. `is_configured`) and default properties for free. However, because we're throwing `@attr.s` classes into the mix I feel like we should stick with **implicit** subtyping in which case we get rid of the member defaults and leave the `is_configured` method body as `...`

I think we can also keep a concrete settings class here too. so something like:

```python
class PTransportSettings(Protocol):
    base_url: str
    api_version: str
    verify_ssl: bool
    timeout: int
    headers: Optional[MutableMapping[str, str]]

    def is_configured(self) -> bool:
        ...


@attr.s(auto_attribs=True, kw_only=True)
class TransportSettings:
    base_url: str = ""
    api_version: str = "3.1"
    verify_ssl: bool = True
    timeout: int = 120
    headers: Optional[MutableMapping[str, str]] = None

    def is_configured(self) -> bool:
        return bool(self.base_url and self.api_version)
```

and then in api_settings.py leave `PApiSettings` exactly how you have it (add `section: Optional[str]` property) and change `ApiSettings` to

```python
@attr.s(auto_attribs=True, kw_only=True)
class ApiSettings(transport.TransportSettings:
    section: Optional[str] = None

    @classmethod
    def configure(...)
    .
    .
    .
     # keep all the remaining code but DITCH is_configured()
```

Let me know if that's not entirely clear. Basically we want to reflect the fact that the transport module can be standalone with it's own settings interface/type and the api_settings module has some extra stuff

python/looker_sdk/rtl/api_settings.py
```diff
@@ -46,25 +47,26 @@ def _convert_bool(val: str, _: bool) -> bool:
     return converted
 
 
-@attr.s(auto_attribs=True, kw_only=True)
-class ApiSettings(transport.TransportSettings):
-    """API Configuration Settings.
+class PApiSettings(transport.PTransportSettings, Protocol):
+    filename: str
```
joeldodge79: I feel like `filename` is an implementation detail `ApiSettings`. It used to be `_filename`, was there a reason to promote it?

and now that I think about it, same thing for `section` (sorry for the back adn forth on that). `filename` and `section` are particular to reading a config file and don't belong in the more general `PApiSettings` type that other (db? redis? dynamodb?) config settings might implement.

josephaxisa: `_filename` was promoted because we are assigning it outside of the class in `configure`:

```python
@classmethod
def configure(
    cls, filename: str = "looker.ini", section: Optional[str] = None
) -> PApiSettings:
    api_settings = cls(filename=filename, section=section)
    .
    .
    .
    settings = converter.structure(config_data, ApiSettings)
```

PR #158, opened 2020-01-31
Fixed tests to not use thelook

muckweaver:
To run:
docker-compose build base; docker-compose build python
docker-compose run python

### DIFF
--- a/python/tests/integration/test_methods.py
+++ b/python/tests/integration/test_methods.py
@@ -269,7 +269,7 @@ def test_crud_look(sdk: mtds.LookerSDK, looks):
         assert look.description == l.get("description")
         assert look.deleted == l.get("deleted")
         assert look.is_run_on_load == l.get("is_run_on_load")
-        #assert look.public == l.get("public")
+        assert look.public == l.get("public")
         assert look.query_id == query.id
         assert look.space_id == l.get("space_id") or sdk.me().home_space_id
         assert look.user_id == l.get("user_id") or sdk.me().id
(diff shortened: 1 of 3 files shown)


python/tests/integration/test_methods.py
```diff
@@ -269,7 +269,7 @@ def test_crud_look(sdk: mtds.LookerSDK, looks):
         assert look.description == l.get("description")
         assert look.deleted == l.get("deleted")
         assert look.is_run_on_load == l.get("is_run_on_load")
-        assert look.public == l.get("public")
+        #assert look.public == l.get("public")
```
muckweaver: @joeldodge79 This is the line I had to comment out. If it's not, the test will fail and I couldn't figure out why.

joeldodge79: hm, on first glance I can't see why it fails. gonna try to reproduce locally and also build with docker

test/data.yml
```diff
@@ -11,32 +11,17 @@ users:
     last_name: Aardvark
   - first_name: Amanda
     last_name: Aardvark
-queries:
-  - id: "#user_query"
-    model: thelook
-    view: users
+queries_system_activity:
```
joeldodge79: @jkaster These data.yml file changes are fine with me, ok for you?

jkaster: for sure, if `yarn test` works for the Typescript tests from the root of the codegen repo

joeldodge79: @muckweaver on second thought, can we revert this top level yaml key back to `queries`? leave the body the same but update the pytest fixture back to `queries` etc. that will make it so we don't have to update all the other SDK tests

muckweaver:
@joeldodge79 I added back i the 'queries' section of data.yml

python/tests/integration/test_methods.py
joeldodge79: this assert works for me in "local" mode. I can't seem to get `docker-compose run python` to execute successfully:

muckweaver:
@joeldodge79 I changed the data.yml to have the public value be true and un-commented out the test for it in test_methods.py

PR #159, opened 2020-02-07
Added support for multiple API versions in the same SDK package

### LINKED ISSUE #72: Python: don't use global catttr converter
currently the python SDK uses the global cattr convert:

```python
import cattr
.
.
.
cattr.register_structure_hook
```

This can interfere with and be polluted by other code that uses the global cattr converter when {un}structure hooks are registered. Instead we should do something like 

```python
import cattr
.
.
.
# use sdk_converter object everywhere now instead of cattr
sdk_converter = cattr.Converter()
sdk_converter.register_structure_hook
```

jkaster:
See the updated main readme for an overview of what changed. There's a lot

fixes #72


python/looker_sdk/client.py
```diff
@@ -23,25 +23,102 @@
...
-    return methods.LookerSDK(
+    return methods31.Looker31SDK(
+        auth_session.AuthSession(settings, transport, serialize.deserialize31),
+        serialize.deserialize31,
+        serialize.serialize,
+        transport,
+    )
+
+
+def init31(
```
joeldodge79: we actually need to delete this entire file and re-add the `init{31,40}` methods into `python/looker_sdk/__init__.py` I can jump on that now

jkaster: Thanks. I realize what you asked me about yesterday now!

PR #160, opened 2020-02-10
Python SDK: Add assertions for body type parameters

josephaxisa:
This came about as a result of this [issue](https://github.com/looker-open-source/sdk-examples/issues/35).

Todo:
- [x] Check in generated methods/models


src/python.gen.spec.ts
```diff
@@ -248,6 +248,25 @@ def run_url_encoded_query(
     })
   })
 
+  describe('body params type assertions' , () => {
```
joeldodge79: I'd imagine you need to update the [assert response is model add_group_group](https://github.com/looker-open-source/sdk-codegen/blob/7925bdb363fd060dfa8b1e05cf6d92007a149c25/src/python.gen.spec.ts#L271) test too?

src/python.gen.ts
```diff
@@ -381,6 +393,7 @@ ${this.hooks.join('\n')}
 
     return this.methodSignature(indent, method)
       + this.summary(bump, method.summary)
+      + (method.bodyParams.length > 0 ? this.bodyParamsTypeAssertions(bump, method.bodyParams) : '')
```
jkaster: Seems this decision could be made in `bodyParamsTypeAssertions` instead so this becomes just a plain:
```typescript
+ this.bodyParamsTypeAssertions(bump, method.bodyParams)
```

src/python.gen.spec.ts
josephaxisa: That test focusses on the http call and hence does not include the statement where we assert the type for body parameters. I am going to move my tests under the `method body` describe block though.

src/python.gen.ts
```diff
@@ -367,6 +366,30 @@ ${this.hooks.join('\n')}
...
+        if (param.location === strBody) {
+          let conditionStr = param.required ? '' : `${indent}if ${param.name}:\n${bump}`
+          let type = this.writeableType(param.type) || param.type
+          let bodyType = this.typeMapMethods(type).name
+          if (bodyType.startsWith('Sequence')) {
+            bodyType = "Sequence"
+          } else if (bodyType.startsWith('MutableMapping')) {
+            bodyType = "MutableMapping"
+          } else if (bodyType.startsWith("models.DelimSequence")) {
+            bodyType = "models.DelimSequence"
```
jkaster: minor nit: inconsistent quoting

src/python.gen.ts
```diff
@@ -138,7 +138,6 @@ import datetime
 from typing import MutableMapping, Optional, Sequence
 
 import attr
-import cattr
```
joeldodge79: oops, thx!

src/python.gen.ts
```diff
@@ -367,6 +366,30 @@ ${this.hooks.join('\n')}
       + `${returnStmt}`
   }
 
+  bodyParamsTypeAssertions(indent: string, bodyParams: IParameter[]): string {
+    const bump = indent + this.indentStr
+    let assertions: string = ''
+    if (bodyParams.length > 0) {
```
joeldodge79: This is a question for @jkaster (not germane to this particular PR) -

it seems odd that we've implemented the body param as an array (`method.bodyParams`). In the old swagger spec a method just has a parameters list and presumably only one has the `"in": "body"` property. But maybe that could be multiple? However, the openAPI spec has a separate `requestBody` top level key for a method and does not have an `"in": "body"` type `"parameters"` instance:

swagger
```json
    "/content_metadata_access/{content_metadata_access_id}": {
      "put": {
        "tags": [
          "Content"
        ],
        "operationId": "update_content_metadata_access",
        "summary": "Update Content Metadata Access",
        "description": "### Update type of access for content metadata.\n",
        "parameters": [
          {
            "name": "content_metadata_access_id",
            "in": "path",
            "description": "Id of content metadata access",
            "required": true,
            "type": "string"
          },
          {
            "name": "body",
            "in": "body",
            "description": "Content Metadata Access",
            "required": true,
            "schema": {
              "$ref": "#/definitions/ContentMetaGroupUser"
            }
          }
        ],
```

openAPI:
```json
    "/content_metadata_access/{content_metadata_access_id}": {
      "put": {
        "tags": [
          "Content"
        ],
        "operationId": "update_content_metadata_access",
        "summary": "Update Content Metadata Access",
        "description": "### Update type of access for content metadata.\n",
        "parameters": [
          {
            "name": "content_metadata_access_id",
            "in": "path",
            "description": "Id of content metadata access",
            "required": true,
            "schema": {
              "type": "string"
            }
          }
        ],
        "requestBody": {
          "$ref": "#/components/requestBodies/ContentMetaGroupUser"
        },
```

src/python.gen.ts
```diff
@@ -367,6 +366,30 @@ ${this.hooks.join('\n')}
...
+          let type = this.writeableType(param.type) || param.type
+          let bodyType = this.typeMapMethods(type).name
+          if (bodyType.startsWith('Sequence')) {
+            bodyType = 'Sequence'
+          } else if (bodyType.startsWith('MutableMapping')) {
+            bodyType = 'MutableMapping'
+          } else if (bodyType.startsWith('models.DelimSequence')) {
+            bodyType = 'models.DelimSequence'
+          }
+          assertions += (`${conditionStr}${indent}assert isinstance(${param.name}, ${bodyType})`)
```
joeldodge79: Given my comment above I don't think there's ever more than 1 element in `bodyParams`, however, if there were I think you want to add a `\n` here and remove L388

src/python.gen.ts
jkaster: "How many body parameters?" is a really good question that mirrors the investigation I had while starting this project. From what I've seen for **our** API we'll only ever have one body parameter, and our generator basically caters to that. It still seems possible we could encounter a spec that has multiple body parameters.

PR #165, opened 2020-02-11
Typescript tests running in the container

muckweaver:
Set your looker.ini also, set the .env or have looker.ini variables set as env vars.
In a shell:

docker-compose build base; docker-compose build typescript
docker-compose run typescript


muckweaver:
Test Suites: 7 failed, 12 passed, 19 total
Tests:       12 failed, 182 passed, 194 total
Snapshots:   0 total
Time:        138.283s
Ran all test suites.

Current issues: 

1. Can't generate sdk if api_version is missing
2. Tests want verify_ssl to be set to 'true' which you can't do if you're running local
3. Tests look for [looker31] section
4. Tests don't see base_url value - may be related to [looker31] section missing

docker-compose.yml
```diff
@@ -19,3 +19,17 @@ services:
       tty: true
       volumes:
         - '.:/sdk_codegen'
+
+    typescript:
+      environment:
+        - LOOKERSDK_BASE_URL=${LOOKERSDK_BASE_URL}
```
jkaster: You can add `LOOKERSDK_INI=/path/to/looker.ini` here if you want the test to using a different INI file

docker-compose.yml
```diff
@@ -19,3 +19,17 @@ services:
...
+        - LOOKERSDK_API_VERSION=${LOOKERSDK_API_VERSION}
```
jkaster: This shouldn't be required any more. We should resolve that

typescript/looker/test/methods.spec.ts
```diff
@@ -50,7 +50,7 @@ const root = fs.existsSync(dataFile) ? '' : '../../'
 const testData = yaml.safeLoad(fs.readFileSync(`${root}${dataFile}`, utf8))
 const localIni = `${root}looker.ini`
 const users: Partial<IUser>[] = testData['users']
-const queries: Partial<IQuery>[] = testData['queries']
+const queries: Partial<IQuery>[] = testData['queries_system_activity']
```
jkaster: Do all the test pass using this testData section instead of `queries`?

docker-compose.yml
muckweaver: Oh this is like this because I'm not sure if there is a way to set env variables for all containers.

docker-compose.yml
muckweaver: I tried to remove API_VERSION and more tests failed and complained it was missing. Not warnings.

typescript/looker/test/methods.spec.ts
muckweaver: If you use 'queries' it will look for "thelook" which not everyone has. If you use 'queries_system_activity' it will use internal looker data. Are there other places in the tests where testData is set?

jkaster: It's the methods tests in the various SDKs that use the data.yml or data.yml.json file. Seems like we should remove our reliance on `thelook` being present and use the system activity model instead, and replace the data in `queries` with `queries_system_activity` model and remove `queries_system_activity` from data.yml.

PR #166, opened 2020-02-12
Python SDK: dynamic imports for generated models

josephaxisa:


python/looker_sdk/rtl/auth_session.py
```diff
@@ -28,9 +28,14 @@
...
 from looker_sdk.rtl import auth_token
+from looker_sdk.rtl import constants
 from looker_sdk.rtl import serialize
 from looker_sdk.rtl import transport
-from looker_sdk.sdk.api31 import models
+
+if constants.api_version == "3.1":
+    from looker_sdk.sdk.api31 import models
+elif constants.api_version == "4.0":
+    from looker_sdk.sdk.api40 import models  # type: ignore
```
joeldodge79: best practice (that I haven't always followed) is to put a comment above this line explaining why we need the `# type: ignore` (usually including the specific mypy error)

joeldodge79: actually, I think you can get rid of the ignore by adding this line above the conditional import:
```python
models: types.ModuleType
```

joeldodge79: ug, sorry. I dug into this more and, rather than trying to provide comments I'm going to just push another commit up to this branch.

Issue #161, opened 2020-02-10
Python: run all integration tests against 3.1 and 4.0 versions


joeldodge79:
parametrize `sdk()` fixture to produce 3.1 and 4.0 clients and fix all tests to run accordingly (right now we hard-code importing api31 models in tests)

PR #168, opened 2020-02-12
Python SDK: Run all tests against both 3.1 and 4.0

josephaxisa:
This addresses #161


josephaxisa:
@joeldodge79 @jkaster This is still a WIP but I would love to hear your thoughts.

python/tests/integration/test_methods.py
```diff
@@ -7,18 +7,20 @@
...
 import pytest  # type: ignore
 
-from looker_sdk.sdk.api31 import methods as mtds
-from looker_sdk import models as ml
+from looker_sdk.sdk.api31 import models as models31
+from looker_sdk.sdk.api40 import models as models40
 
 
-def test_crud_user(sdk: mtds.LookerSDK):
+def test_crud_user(sdk, models):
```
jkaster: What happens when the models vary here? Is there a union type you can declare?

python/tests/integration/test_netrc.py
```diff
@@ -41,10 +39,10 @@ def create_netrc_file(sdk: mtds.LookerSDK):
     reason="netrc file cannot be created because it already exists or $HOME is undefined",  # noqa: B950
 )
 @pytest.mark.usefixtures("create_netrc_file")
-def test_netrc_does_not_override_ini_creds(sdk: mtds.LookerSDK):
+def test_netrc_does_not_override_ini_creds(sdk, models):
```
jkaster: Same question about typed parameters here

python/tests/conftest.py
```diff
@@ -97,9 +92,25 @@ def create_test_users(
...
+
 @pytest.fixture(scope="session")
-def sdk():
-    # TODO multiplex this fixture for both init31 and init40
-    sdk = looker_sdk.init31("../looker.ini")
+def sdk(api_version: str):
+    if api_version == "3.1":
+        sdk = looker_sdk.init31("../looker.ini")
+    elif api_version == "4.0":
+        sdk = looker_sdk.init40("../looker.ini")  # type: ignore
```
joeldodge79: declare `sdk: Union[methods31:Looker31SDK, methods40.Looker40SDK]` above and you should be able to remove the `# type: ignore`

python/tests/integration/test_methods.py
joeldodge79: we can declare the type for the `sdk` arg (`sdk: Union[methods31.Looker31SDK, methods40.Looker40SDK]`) but mypy doesn't support using modules as types. I tried `models: Union[methods31, methods40]` and get `Module "looker_sdk.sdk.api31.methods" is not valid as a type` and `Module "looker_sdk.sdk.api40.methods" is not valid as a type`

so I think we'll just have to leave type annotations off the `models` argument here for now. An alternative would be to ditch the `models` fixture and import names directly from either models31/models40 but that would be a lot of extra code and against our practice of importing/using modules rather than names directly from modules. Maybe mypy will implement this at some point?

python/tests/conftest.py
```diff
@@ -52,55 +54,70 @@ def _get_test_data() -> TTestData:
...
+def remove_test_dashboards(sdk40: methods40.Looker40SDK, dashboards):
     # Clean up any test dashboards that may exist. We do this at the beginning
     # instead of the end of tests in case we want to view the dashboards after the test
     for d in dashboards:
-        search_results = sdk.search_dashboards(title=d["title"])
+        search_results = sdk40.search_dashboards(title=d["title"])
         if len(search_results) > 0:
             for dashboard in search_results:
-                sdk.delete_dashboard(cast(str, dashboard.id))
+                sdk40.delete_dashboard(cast(str, dashboard.id))
```
jkaster: if dashboard.id can be a LookML dashboard ID ref, I don't know why we still have integer dashboard.id for this structure. We should fix this for 4.0 or someone won't be able to do anything with LookML dashboards?

python/tests/conftest.py
```diff
@@ -52,55 +54,70 @@ def _get_test_data() -> TTestData:
...
+            models40.WriteUser(first_name=u["first_name"], last_name=u["last_name"])
         )
 
         if user.id:
             user_ids.append(user.id)
             email = f"{u['first_name']}.{u['last_name']}{email_domain}"
-            sdk.create_user_credentials_email(
-                user.id, looker_sdk.models.WriteCredentialsEmail(email=email)
+            sdk40.create_user_credentials_email(
+                user_id=user.id, body=models40.WriteCredentialsEmail(email=email)
```
jkaster: is the use of named parameters here caused by a change in the generator?

python/tests/conftest.py
josephaxisa: This is just a typing cast, so it returns the `dashboard.id` unchanged but signals to `mypy` that the return value is a `str`. The reason it's needed is because `search_dashboards` returns a `models.Dashboard` where `id` is of type `Optional[str]`. Without the cast `mypy` complains that we might be passing a `None` type when a `str` is expected to `delete_dashboard`.

python/tests/conftest.py
josephaxisa: This is my bad. I didn't mean to check in this change. I'll revert so it conforms with the rest of the file.

PR #170, opened 2020-02-13
Added LOOKERSDK_INI environment variable for python tests

jkaster:


python/tests/conftest.py
```diff
@@ -100,6 +100,7 @@ def create_test_users(
 @pytest.fixture(scope="session")
 def sdk():
     # TODO multiplex this fixture for both init31 and init40
-    sdk = looker_sdk.init31("../looker.ini")
+    filename = os.getenv("LOOKERSDK_INI", "../looker.ini")
```
josephaxisa: LGTM, I just have one minor comment. I noticed you changed `nodeSession.ts` to also use the `LOOKERSDK_INI` env variable and to retain consistency between SDKs we should also do this in `api_settings.py`. Once that's done this line won't be needed. However, this is outside of the scope of this PR.

jkaster: The `LOOKERSDK_INI` check is in `nodeSettings.spec.ts`, which is a test file. It shouldn't be in the run-time SDK code.

josephaxisa: Ugh, you're right. The change in `nodeSession.ts` is unrelated. Not sure what I was thinking when I wrote the above.

PR #171, opened 2020-02-14
Switching to AnyCodable as a solution to Swift's severe lack of support for deserializing JSON values to Any

jkaster:
- Modified the Swift generator to use `AnyCodable` instead of `Variant?`
- Commented out all references to `Variant` in the Swift SDK for probable eventual removal
- Added a `serialize` function to produce JSON from an SDK object (the opposite of `deserialize`)
- Added a test called `testDictFromJson` that shows how `vis_config` values can be examined. There are probably better code patterns for this.

### DIFF
--- a/swift/looker/Package.swift
+++ b/swift/looker/Package.swift
@@ -16,17 +16,21 @@ let package = Package(
     dependencies: [
         // Dependencies declare other packages that this package depends on.
         // .package(url: /* package url */, from: "1.0.0"),
+      .package(
+           url: "https://github.com/Flight-School/AnyCodable",
+           from: "0.2.3"
+       ),
     ],
     targets: [
         // Targets are the basic building blocks of a package. A target can define a module or a test suite.
         // Targets can depend on other targets in this package, and on products in packages which this package depends on.
         .target(
             name: "looker",
-            dependencies: [],
+            dependencies: ["AnyCodable"],
             path: ".",
             sources: ["rtl", "sdk"]),
         .testTarget(
             name: "lookerTests",
-            dependencies: ["looker"]),
+            dependencies: ["looker", "AnyCodable"]),
     ]
 )
--- a/swift/looker/Tests/lookerTests/transportTests.swift
+++ b/swift/looker/Tests/lookerTests/transportTests.swift
@@ -72,7 +72,7 @@ class transportTests: XCTestCase {
             print(error)
         }
     }
-    
+
     func testRegexExtension() {
         checkRegex(Constants.matchModeString, "string match")
         checkRegex(Constants.matchCharset, "charset match")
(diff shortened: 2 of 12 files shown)


swift/looker/Package.swift
```diff
@@ -16,17 +16,21 @@ let package = Package(
     dependencies: [
         // Dependencies declare other packages that this package depends on.
         // .package(url: /* package url */, from: "1.0.0"),
+      .package(
+           url: "https://github.com/Flight-School/AnyCodable",
```
rlm1023: I don't know what the good ones are, but I assume this package is under a license we can use?

swift/looker/Tests/lookerTests/transportTests.swift
```diff
@@ -133,12 +133,86 @@ class transportTests: XCTestCase {
...
+        XCTAssertNotNil(vis_config)
+        XCTAssertNotNil(data)
+        XCTAssertNotNil(json)
+        let bool = vis_config["bool"] as! Bool
+        XCTAssertTrue(bool)
+        let int = vis_config["int"] as! Int64
+        XCTAssertEqual(int, 1)
+        let nada = vis_config["nada"]
+
+        if "\(nada!)" == "nil" {
```
rlm1023: That is an interesting if statement.

swift/looker/Tests/lookerTests/transportTests.swift
```diff
@@ -133,12 +133,86 @@ class transportTests: XCTestCase {
...
+            // nada is nil as expected
+        } else {
+            XCTAssertTrue(false, "nada should be nil, not '\(nada!)'")
+        }
+        let dub = vis_config["dub"] as! Double
+        XCTAssertEqual(dub, 2.3)
```
rlm1023: Floating point comparison :clench:

Maybe use this?
https://developer.apple.com/documentation/xctest/xctassertequalwithaccuracy

swift/looker/Tests/lookerTests/transportTests.swift
```diff
@@ -133,12 +133,86 @@ class transportTests: XCTestCase {
...
+    let visJson = """
+{
+"bool":true,
+"int":1,
+"dub":2.3,
+"str":"Simple string",
+"date":"2018-03-15T13:16:34.692-07:00",
+"nada":null,
+"dict": {"A":4, "B": 2, "C": true},
+"ratnest": [ { "one": 1, "two": "two" }, "three", {"four":4} ]
```
rlm1023: 🤣

swift/looker/Package.swift
jkaster: Yes, it's MIT. I'll get the dependency notification going today.

swift/looker/Tests/lookerTests/transportTests.swift
jkaster: I tried about 5 different methods of testing `Any?` for `nil` and none of the documented ways like `let foo = ...` were working for me. If you find a cleaner **reliable** way to do it, please let me know.

swift/looker/Tests/lookerTests/transportTests.swift
jkaster: Oh yeah, thanks for the reminder. I meant to do that. Will update. The version you link here is obsolete but there is an `accuracy` tag for `XCTAssertEqual`

PR #176, opened 2020-02-25
Implemented error handling by default for Swift

jkaster:
- `sdk.ok()` now throws if an error is encountered
- created a `listGetter` test harness like the one for Kotlin
- updated `methodsTests.swift` to use `listGetter`

### DIFF
--- a/swift/looker/Tests/lookerTests/baseTransportTests.swift
+++ b/swift/looker/Tests/lookerTests/baseTransportTests.swift
@@ -15,6 +15,7 @@ fileprivate let repoPath : String = testRootPath + "/../../"
 fileprivate let localIni : String = ProcessInfo.processInfo.environment["LOOKERSDK_INI"] ?? (repoPath + "looker.ini")
 
 
+@available(OSX 10.15, *)
 class TestConfig {
     
     var rootPath = repoPath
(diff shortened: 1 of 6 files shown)


swift/looker/Tests/lookerTests/methodsTests.swift
```diff
@@ -85,78 +85,94 @@ class methodsTests: XCTestCase {
         } else {
             XCTAssertTrue(false, "Couldn't cast data from jsonData")
         }
+                
+        } catch {
+            print(error)
```
rlm1023: Might even throw it to fail the test? Not sure what would happen here if it failed.

swift/looker/Tests/lookerTests/baseTransportTests.swift
```diff
@@ -15,6 +15,7 @@ fileprivate let repoPath : String = testRootPath + "/../../"
 fileprivate let localIni : String = ProcessInfo.processInfo.environment["LOOKERSDK_INI"] ?? (repoPath + "looker.ini")
 
 
+@available(OSX 10.15, *)
```
rlm1023: Interesting that these appeared/changed?

swift/looker/Tests/lookerTests/methodsTests.swift
jkaster: I think I'll leave it with the current pattern for now. Still learning Swift practices.

swift/looker/Tests/lookerTests/baseTransportTests.swift
jkaster: It's because I pushed more initialization into `TestConfig()` that requires 10.15

PR #179, opened 2020-02-27
Kotlin models are now serializable

jkaster:
Added `: Serializable` to Kotlin SDK models

### DIFF
--- a/kotlin/src/test/TestAsync.kt
+++ b/kotlin/src/test/TestAsync.kt
@@ -8,64 +8,65 @@ import kotlin.test.assertNotNull
 import org.junit.Test as test
 
 class TestAsync {
-    private val sdk = TestConfig().sdk
-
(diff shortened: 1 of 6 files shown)


kotlin/src/test/TestAsync.kt
```diff
@@ -8,64 +8,65 @@ import kotlin.test.assertNotNull
...
-            assertNotNull(dashboards)
-            assertNotNull(looks)
-            assertEquals(limits.first!!.toInt(), dashboards.count(), "${limits.first} viewed Dashboards")
-            assertEquals(limits.second!!.toInt(), looks.count(), "${limits.second} viewed Looks")
-        }
-    }
+    // for >7.2
+//    private val sdk = TestConfig().sdk
+//
+//    // see https://kotlinlang.org/docs/reference/coroutines/composing-suspending-functions.html#structured-concurrency-with-async
```
rlm1023: Why'd all this get commented out?

jkaster: because it's for properties that are not in 7.2

PR #181, opened 2020-02-28
This fix failure of SDK in extensions as path is prefixed by API version

bryans99:
…ion.

Prefix now does not happen if base_url is not set (as is the case for extension transport). Node and browser transports must set the base url.


bryans99:
Can you ping me when this is published. I need to update the extension packages and republish :)

jkaster:
Will do. I've got my encoding PR up for review right now. Should be today.

PR #182, opened 2020-02-28
Parameter encoding for the SDKs

jkaster:
Input parameters to API calls should now all be correctly encoded

There may still be a little bit left to do for Python


src/sdkModels.ts
```diff
@@ -209,6 +211,10 @@ export class Parameter implements IParameter {
   asHashString() {
     return `${this.name}:${this.type.name}${this.required ? '' : '?'}${this.location}`
   }
+
+  doEncode() {
+    return this.type.name === 'string' || this.type.name === 'datetime' || this.type.name === 'date'
+  }
```
joeldodge79: 👍

src/typescript.gen.spec.ts
```diff
@@ -202,17 +202,17 @@ async all_datagroups(
...
-result_format = this.encodePathParam(result_format)
+`model_name = encodeParam(model_name)
+view_name = encodeParam(view_name)
+result_format = encodeParam(result_format)
 `
       const actual = gen.encodePathParams('', method)
       expect(actual).toEqual(expected)
     })
-    it('encodes only string path params', () => {
+    it('encodes only string or date path params', () => {
```
joeldodge79: hm, the name of the test is misleading. `run_look` has `look_id` (int) and `result_format` (string) so maybe

```suggestion
    it('encodes does not encode int path params', () => {
```

or at least a comment to the effect of
```suggestion
    // TODO eventually add method that has a date type path param
    it('encodes only string or date path params', () => {
```

typescript/looker/rtl/transport.ts
```diff
@@ -336,7 +364,7 @@ export function encodeParams(values?: Values) {
   const keys = Object.keys(values)
   const params = keys
     .filter(k => values[k] !== undefined) // `null` and `false` will both be passe
-    .map(k => k + '=' + encodeURIComponent(values[k]))
+    .map(k => k + '=' + encodeParam(values[k].toString()))
```
joeldodge79: ah, I didn't realize you completely control the query string formation. in python we rely on the requests library to do the query string encoding for us. I just read the [source code](https://github.com/psf/requests/blob/master/requests/models.py#L83) which in turn uses [std lib urllib.parse.urlencode](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode) which will handle the `/` for us so I think we're good there. The main difference is that the underlying library is assuming it's not already encoded and not checking it. We may want to check it ourselves (and decode if it's encoded) before handing it off. Even though that's a little hacky I'm still happy to leave the encoding logic in python-requests for now. If we ever need to implement a different python transport then we'll have to do it like you've done here.

jkaster: Does python encode date values in the format our API expects? IOW, UTC.

src/sdkModels.ts
jkaster: Note that, because Swift and Kotlin method parameters are immutable, we can't easily take advantage of `param.doEncode()` until we clean up the code generator design a bit more.

typescript/looker/rtl/transport.ts
joeldodge79: hm, maybe I'm seeing this wrong but now it looks to me like you're calling `.toString()` before it gets inside `encodeParam()` which will skip the special `if (value instanceof Date)` check, no?

jkaster: It's possible we could make all SDKs use built-in param encoding, but the pattern varies enough and the problem is clearly defined enough that I opted for this pattern since it was pretty easy to implement for each runtime.

joeldodge79: so I do think you have a bug here. On the python side I'm sure datetime is broken so we probably need to bite the bullet and conform

jkaster: yeah, that `toString()` should be removed. Fixing.

joeldodge79: actually, we do handle [converting query params](https://github.com/looker-open-source/sdk-codegen/blob/master/python/looker_sdk/rtl/api_methods.py#L101) before passing them down to python-requests (sending values that look like `'2020-02-28T14:45Z'`) so it's not currently broken in python

PR #184, opened 2020-03-06
Enable caller to select which LookerSDK is used by the ExtensionSDK

dthorpe:
examples: 
ExtensionSDK.createClient(host)  -> constructs LookerAPI31, DefaultSettings

ExtensionSDK.createClient(host, LookerAPI31) -> constructs LookerAPI31, DefaultSettings

ExtensionSDK.createClient(host, LookerAPI40) -> constructs LookerAPI40, DefaultSettings

ExtensionSDK.createClient(host, LookerAPI40, mySpecialSettings) -> constructs LookerAPI40

### DIFF
--- a/typescript/looker/rtl/apiMethods.ts
+++ b/typescript/looker/rtl/apiMethods.ts
@@ -41,6 +41,18 @@ export class APIMethods {
     this.apiPath = authSession.settings.base_url === '' ? '' : `${authSession.settings.base_url}/api/${this.apiVersion}`
   }
 
+  /** A helper method for constructing with a type as a param
+   *
+   * Example:
+   *   APIMethods.create(Looker40SDK, authSession)
+   *
+   * @param type
+   * @param authSession
+   */
+  static create<T extends APIMethods>(type: (new (authSession: IAuthSession) => T), authSession: IAuthSession): T {
+    return new type(authSession)
+  }
+
   /** A helper method for simplifying error handling of SDK responses.
    *
    * Pass in a promise returned by any SDK method, and it will return a promise
--- a/typescript/looker/rtl/extensionSdk.ts
+++ b/typescript/looker/rtl/extensionSdk.ts
@@ -22,7 +22,7 @@
  * THE SOFTWARE.
  */
 
-import { DefaultSettings } from './apiSettings'
+import { DefaultSettings, IApiSettings } from './apiSettings'
 import { Looker31SDK } from '../sdk/3.1/methods'
 import { ExtensionSession } from './extensionSession'
 import { ExtensionTransport } from './extensionTransport'
@@ -30,6 +30,8 @@ import {
   Authenticator,
   ITransportSettings,
 } from './transport'
+import { APIMethods } from "./apiMethods";
+import { IAuthSession } from "./authSession";
 
 export interface IHostConnection {
   request(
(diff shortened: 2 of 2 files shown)


typescript/looker/rtl/apiMethods.ts
```diff
@@ -41,6 +41,18 @@ export class APIMethods {
...
+  /** A helper method for constructing with a type as a param
+   *
+   * Example:
+   *   APIMethods.create(Looker40SDK, authSession)
+   *
+   * @param type
+   * @param authSession
+   */
+  static create<T extends APIMethods>(type: (new (authSession: IAuthSession) => T), authSession: IAuthSession): T {
+    return new type(authSession)
```
dthorpe: This fn isn't used by this PR, and the example rings a little hollow, since the real use case is when the caller doesn't have a concrete sdk type, just passing through a type param.  Can be deleted if desired.

typescript/looker/rtl/extensionSdk.ts
```diff
@@ -22,14 +22,16 @@
...
+import { DefaultSettings, IApiSettings } from './apiSettings'
 import { Looker31SDK } from '../sdk/3.1/methods'
 import { ExtensionSession } from './extensionSession'
 import { ExtensionTransport } from './extensionTransport'
 import {
   Authenticator,
   ITransportSettings,
 } from './transport'
+import { APIMethods } from "./apiMethods";
+import { IAuthSession } from "./authSession";
```
jkaster: Surprised prettier didn't transform those to single quote, no semis

typescript/looker/rtl/extensionSdk.ts
```diff
@@ -42,15 +44,28 @@ export interface IHostConnection {
...
     const transport = new ExtensionTransport(settings, hostConnection)
     const session = new ExtensionSession(settings, transport)
-    return new Looker31SDK(session)
+    if (type) {
+      return new type(session)
+    }
+    else
+    {
+      // work around TS2322
+      return ((new Looker31SDK(session)) as any) as T
```
jkaster: Does the consumer of this constructor provide type checking for API 4.0 methods vs. API 3.1 methods with this construction, or does this function think it returns `Looker31SDK | Looker40SDK`?

IOW, does

```typescript
let sdk = LookerExtensionSDK.createClient(host, Looker40SDK)
```

provide edit-time type checking for API 4.0?

If so, this looks good to me. If not, that's why we're using `.init31()` and `.init40()` methods for the other SDKs.

typescript/looker/rtl/extensionSdk.ts
joeldodge79: you probably have your IDE configured to run it automatically.

dthorpe: Those were auto-added by Intellij

typescript/looker/rtl/extensionSdk.ts
dthorpe: The return type is `T extends APIMethods`, and the type passed as a parameter leads to type inferencing on the return type.  There is no union type.  In the code snippet you have above, `sdk` will be exactly `Looker40SDK` type.

This is also future-proof; it'll work for any future LookerSDK types, as long as they extend APIMethods.

dthorpe:
@bryans99 extension-sdk's engagement with LookerSDK is solely via the static `createClient` method, right?  Since the new params are optional, what corresponding changes are needed in extension-sdk?

typescript/looker/rtl/extensionSdk.ts
jkaster: Excellent. Need to adopt that for other SDKs that can support it. Definitely for the other TS factories at least.

PR #187, opened 2020-03-13
Improved the legacy code generator

jkaster:
- The API 4.0 specification is now used by default for `yarn legacy`
- fixed namespace parameter specifier for the OpenAPI generator
- `x-looker-nullable` is replaced with `nullable` in the converted OpenAPI spec files
- legacy now only generates C# by default because the other active languages are supported by our code generator
- updated readme with more instructions for the legacy generator

### DIFF
--- a/src/convert.spec.ts
+++ b/src/convert.spec.ts
@@ -1,7 +1,7 @@
 /*
  * The MIT License (MIT)
  *
- * Copyright (c) 2019 Looker Data Sciences, Inc.
+ * Copyright (c) 2020 Looker Data Sciences, Inc.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
--- a/src/convert.spec.ts
+++ b/src/convert.spec.ts
@@ -1,0 +1,75 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2019 Looker Data Sciences, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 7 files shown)


src/convert.spec.ts
```diff
@@ -0,0 +1,75 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2019 Looker Data Sciences, Inc.
```
dthorpe: Does Intellij have a new file template feature somewhere?  It's getting tedious to copy this header around in new files.  Plus, update the year for new files

src/convert.ts
```diff
@@ -41,24 +42,56 @@ const lintCheck = async (fileName: string) => {
     const linter = new Spectral()
     if (!linter) return fail('Lint', 'no response')
     const spec = parseWithPointers(readFileSync(fileName))
-    linter.run({
+    linter
+      .run({
         parsed: spec,
-        getLocationForJsonPath,
+        getLocationForJsonPath
```
dthorpe: missing trailing comma?

src/convert.ts
```diff
@@ -41,24 +42,56 @@ const lintCheck = async (fileName: string) => {
...
+ * @param {string} openApiFile name of the Open API file to process
+ * @returns {Promise<string>} name of the file written
+ */
+export const swapNullableInFile = async (openApiFile: string) => {
+  if (!isFileSync(openApiFile)) {
+    return quit(`${openApiFile} was not found`)
+  }
+  log(`replacing "x-looker-nullable" with "nullable" in ${openApiFile} ...`)
+  const contents = swapXLookerNullable(readFileSync(openApiFile))
+  await fs.writeFileSync(openApiFile, contents)
```
dthorpe: await on sync function?

src/legacy.ts
```diff
@@ -24,41 +24,61 @@
...
   return results
 }
-
-try {
-  const config = SDKConfig()
-  Object.entries(config).forEach(async ([name, props]) => runConfig(name, props))
-} catch (e) {
-  quit(e)
-}
+;(async () => {
```
dthorpe: why the wrapper?

dthorpe: Oh, to make it async.

src/convert.spec.ts
jkaster: Oh, good point on year. Nope, I'm just copy/pastaing .. and have a wishlist item to inject the license in all source files where it's missing w/ a command-line script. We need something that will work for all source file types, of course.

src/convert.ts
jkaster: yeah, just noticed that

PR #194, opened 2020-03-24
Fix the sha256 test.

rlm1023:


rlm1023:
Yeah, there's probably a more correct solution than "line the cards up just right", but in the interest of getting it working...

PR #197, opened 2020-03-31
Add setToken to kotlin's AuthToken class to handle updating tokens.

rlm1023:


kotlin/src/main/com/looker/rtl/OAuthSession.kt
```diff
@@ -127,7 +127,7 @@ class OAuthSession(override val apiSettings: ApiSettings, override val transport
                 val config = this.apiSettings.readConfig()
                 // fetch the token
                 this.requestToken(mapOf(
-                        "grant_type" to "request_token",
+                        "grant_type" to "refresh_token",
```
rlm1023: Separate bug fix.

kotlin/src/main/com/looker/rtl/AuthToken.kt
```diff
@@ -32,9 +32,9 @@ data class AuthToken(
         @JsonProperty("access_token")
         var accessToken: String = "",  // TODO: Consider making this/these vals and using new objects instead of mutating
         @JsonProperty("token_type")
-        val tokenType: String = "",
+        var tokenType: String = "",
         @JsonProperty("expires_in")
-        var expiresIn: Long = 0,
+        var expiresIn: Long = 0L,
         @JsonProperty("refresh_token")
         var refreshToken: String? = "") {
```
dthorpe: Is this default value = empty string going to interfere with the token.refresh_token != nil check below? or other places that need to know if the refresh_token is nil?

rlm1023: I too think this default value is a bit strange. But the check below is for what the AccessToken has on it, which gets optionally set _into_ this field.

This default value could cause strange behavior in OAuthSession where this value is read, but I decided against making larger changes where it seems to be working.

PR #199, opened 2020-04-03
Python sdk can use python3.8

joeldodge79:
Updated local development environment to be strict about using 3.8.2 for
now (so we all have the same experience). Don't forget to upgrade pyenv
and install (can be done from any directory):

```
brew upgrade pyenv
pyenv install 3.8.2
```

then, from within your `sdk-codegen/python/` directory do the following:
```
rm -rf `pipenv --venv`
pipenv install --dev
```


joeldodge79:
I cheated a little bit: had to test live on pypi.org (test.pypi.org just doesn't cut it because `cattrs` etc isn't there). had a small snag but now `0.1.3b10` is live (based on this PR) and working...

josephaxisa:
> I cheated a little bit: had to test live on pypi.org (test.pypi.org just doesn't cut it because `cattrs` etc isn't there). had a small snag but now `0.1.3b10` is live (based on this PR) and working...

You might be able use this next time: https://packaging.python.org/guides/using-testpypi/#using-testpypi-with-pip

joeldodge79:
ah, I should have used `--extra-index-url` so that `cattrs` et all installed. good to know, thanks!

PR #202, opened 2020-04-09
Python SDK: settings refactor

joeldodge79:
Make the python ApiSettings conform to the other sdks: instance properties/members will be the properties declared on `PTransportSettings` and `readConfig` will produce a dictionary with keys corresponding to further settings that `AuthSession` needs

I also refactored AuthSession to remove references to "admin" and
substituted "user" with "sudo" and DRY'ed out some logout code there.

### DIFF
--- a/python/looker_sdk/__init__.py
+++ b/python/looker_sdk/__init__.py
@@ -21,7 +21,6 @@
 # THE SOFTWARE.
 
 from typing import Optional
-import warnings
 
 from looker_sdk.rtl import api_settings
 from looker_sdk.rtl import requests_transport
@@ -47,21 +46,16 @@ def init31(
 ) -> methods31.Looker31SDK:
     """Default dependency configuration
     """
-    settings = api_settings.ApiSettings.configure(config_file, section)
-    if settings.api_version is not None:
-        warnings.warn(message=DeprecationWarning(API_SETTINGS_API_VERSION_DEPRECATED))
-    settings.api_version = "3.1"
-    settings.headers = {"Content-Type": "application/json"}
+    settings = api_settings.ApiSettings(config_file, section)
     if not settings.is_configured():
-        raise InitError(
-            f"Missing required configuration values like base_url and api_version."
-        )
+        raise InitError(f"Missing required configuration values.")
     transport = requests_transport.RequestsTransport.configure(settings)
     return methods31.Looker31SDK(
-        auth_session.AuthSession(settings, transport, serialize.deserialize31),
+        auth_session.AuthSession(settings, transport, serialize.deserialize31, "3.1"),
         serialize.deserialize31,
         serialize.serialize,
         transport,
+        "3.1",
     )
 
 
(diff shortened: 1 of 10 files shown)


python/looker_sdk/rtl/api_settings.py
```diff
@@ -40,105 +38,125 @@
...
+    looker_url: Optional[str]
+    redirect_uri: Optional[str]
+
+    @property
+    def client_id(self) -> Optional[str]:
         ...
 
-    def get_client_secret(self) -> Optional[str]:
+    @property
+    def client_secret(self) -> Optional[str]:
```
joeldodge79, line 50: @jkaster I wanted to have a conversation with you about what our settings interface should look like. I see that in typescript you have a `[key: string]: any` but I'm wondering if we should take the route of being more explicit in the interface enumerating all the possible settings? That way someone can implement they're own settings provider and not have to guess at what's in `[key: string]: any` (or read the code) in order to make their settings work for our different RTL components.

Let's chat about it

python/looker_sdk/rtl/api_settings.py
```diff
@@ -40,105 +38,125 @@
...
         ...
 
 
-@attr.s(auto_attribs=True, kw_only=True)
```
joeldodge79, line 62: attrs classes are supposed to be pretty simple data containers. this settings implementation is a bit more than just that so I think it's easier to reason about the code when explicitly typed out (rather than relying on attrs/cattrs)

python/looker_sdk/rtl/api_settings.py
jkaster: Yeah, let's catch up on this. The reason the Typescript definition is so loose is because the configuration values someone may retrieve via (typically) the `readConfig()` overload could be values we know nothing about. Their authentication process could be completely different than any we've provided samples for, without us knowing what values they'd need. For example, maybe they're retrieving credentials from a key store on GCP, Azure, or AWS.

joeldodge79: recap on this conversation:
1/ instance properties/members will be the properties on `PTransportSettings`
2/ `readConfig` will produce a dictionary with keys corresponding to further settings that `AuthSession` needs
3/ nice to have: `AuthSession` should validate the necessary keys are present in `readConfig` (e.g. for api3 auth verify `client_id` and `client_secret` are returned, for oauth verify `client_id`, `looker_url`, and `redirect_uri` are returned)

joeldodge79:
@jkaster and @josephaxisa this is ready for review

python/looker_sdk/__init__.py
```diff
@@ -47,21 +46,16 @@ def init31(
...
-    if settings.api_version is not None:
-        warnings.warn(message=DeprecationWarning(API_SETTINGS_API_VERSION_DEPRECATED))
-    settings.api_version = "3.1"
-    settings.headers = {"Content-Type": "application/json"}
+    settings = api_settings.ApiSettings(config_file, section)
     if not settings.is_configured():
-        raise InitError(
-            f"Missing required configuration values like base_url and api_version."
-        )
+        raise InitError(f"Missing required configuration values.")
```
josephaxisa, line 51: This can be a normal string
```suggestion
        raise InitError(Missing required configuration values.")
```

python/looker_sdk/__init__.py
```diff
@@ -70,19 +64,14 @@ def init40(
...
-    if settings.api_version is not None:
-        warnings.warn(message=DeprecationWarning(API_SETTINGS_API_VERSION_DEPRECATED))
-    settings.api_version = "4.0"
-    settings.headers = {"Content-Type": "application/json"}
+    settings = api_settings.ApiSettings(config_file, section)
     if not settings.is_configured():
-        raise InitError(
-            f"Missing required configuration values like base_url and api_version."
-        )
+        raise InitError(f"Missing required configuration values.")
```
josephaxisa, line 69: Same here
```suggestion
        raise InitError("Missing required configuration values.")
```

python/looker_sdk/rtl/api_methods.py
```diff
@@ -61,12 +61,21 @@ def __init__(
...
+            auth.settings.base_url, f"/api/{api_version}/"
+        )
         self.deserialize = deserialize
         self.serialize = serialize
         self.transport = transport
 
+    def _path(self, path: str) -> str:
+        if path[0] == "/":
+            path = path[1:]
+        return urllib.parse.urljoin(self.api_path, path)
```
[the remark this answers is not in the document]
joeldodge79: oh, interesting. once again I did this a little different: I simplified transport to blindly accept and attempt whatever path is given w/out modification. I made both ApiMethods and AuthSession craft the fully formed `f"https://{base_url}/api/{version}/{path}"` to pass into transport.

I need to spend some more time groking what you have in typescript for `ApiMethods.makeUrl` vs `BaseTransport.makeUrl`

PR #205, opened 2020-04-14
Improved error handling for Swift HTTP requests

jkaster:
This should handle more failure conditions mo bettah

### DIFF
--- a/swift/looker/rtl/baseTransport.swift
+++ b/swift/looker/rtl/baseTransport.swift
@@ -99,14 +99,27 @@ class BaseTransport : ITransport  {
 
         var result: RequestResponse? = nil
         let task = self.session.dataTask(with: req!) { data, response, error in
-            let res = response as? HTTPURLResponse
             var sdkError: SDKError?
-            if let err = error, let r = res {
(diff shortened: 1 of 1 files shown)


swift/looker/rtl/baseTransport.swift
```diff
@@ -99,14 +99,27 @@ class BaseTransport : ITransport  {
...
+                } else {
+                    if let d = data {
+                        sdkError = SDKError(msg, reason: String(data: d, encoding: .utf8))
+                    } else {
+                        sdkError = SDKError(msg, reason: "Unknown failure")
+                    }
+                }
+            } else {
+                if let r = response as? HTTPURLResponse {
+                    if !BaseTransport.ok(r) {
```
rlm1023, line 116: You can combine these two to reduce nesting. But the execution result is slightly different...
```
if let r = response as? HTTPURLResponse, !BaseTransport.ok(r) {
```

jkaster: I'm just trying to eliminate all potential paths when compounded errors might result in an untrapped error condition. Without being able to reliably test that line I'm more comfortable leaving it broken out.

PR #207, opened 2020-04-20
Python OAuth session support

joeldodge79:
Added OAuth session whose `_login` method calls `/api/token` with a
`refresh_token` grant_type. This assumes that you've set `self.token` with
an access token that has a `refresh_token`. We've provided a convenience
function to do this: `redeem_auth_code` which takes a temporary code
you've been given (see below) and calls `/api/token` with an
`authorization_code` grant_type.

In order to get the temporary code you must ask the authorization server
for it which will usually require a login, a redirect to authorize the
access, and then a final redirect to the location you supplied in the
initial request with a `code` query string parameter. For Looker
authorization we've provided another convenience function:
`create_auth_code_request_url` which will construct a URL with the
appropriate query string parameters to the Looker authorization server.

PS tests/integration/api40/test_methods.py tests are broken because the
spec changed. The failing tests are not related to this change. I'll
open a separate PR to fix them

### DIFF
--- a/python/looker_sdk/rtl/auth_session.py
+++ b/python/looker_sdk/rtl/auth_session.py
@@ -22,13 +22,17 @@
 
 """AuthSession to provide automatic authentication
 """
+import hashlib
+import secrets
 from typing import cast, Dict, Optional, Type, Union
 import urllib.parse
 
+import attr
 
 from looker_sdk import error
 from looker_sdk.rtl import api_settings
 from looker_sdk.rtl import auth_token
+from looker_sdk.rtl import model
 from looker_sdk.rtl import serialize
 from looker_sdk.rtl import transport
 from looker_sdk.sdk.api31 import models as models31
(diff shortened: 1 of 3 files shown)


python/looker_sdk/rtl/auth_session.py
```diff
@@ -215,3 +219,110 @@ def _ok(self, response: transport.Response) -> str:
...
+    @attr.s(auto_attribs=True, kw_only=True)
+    class AuthCodeGrantTypeParams(GrantTypeParams):
+        code: str
+        code_verifier: str
+        grant_type: str = "authorization_code"
+
+    @attr.s(auto_attribs=True, kw_only=True)
+    class RefreshTokenGrantTypeParams(GrantTypeParams):
+        refresh_token: str
+        grant_type: str = "refresh_token"
```
joeldodge79, line 290: These types are undocumented in the API so there are no generated models for them. I've inlined them here because I only understand they're existence  in the python SDK as an implementation detail of this `OAuthSession` class. Alternatively they could go in `looker_sdk.rtl.model` as more "public" place but I assume they're undocumented in the API for a reason?

python/looker_sdk/rtl/auth_session.py
```diff
@@ -215,3 +219,110 @@ def _ok(self, response: transport.Response) -> str:
...
+        self.token = auth_token.AuthToken(access_token)
+
+    def _login(self) -> None:
+        params = self.RefreshTokenGrantTypeParams(
+            client_id=self.client_id,
+            redirect_uri=self.redirect_uri,
+            refresh_token=self.token.refresh_token,
+        )
+        access_token = self._request_token(params)
+        self.token = auth_token.AuthToken(access_token)
```
joeldodge79, line 328: this is another area where the python RTL differs a bit from the other RTLs: instead of `getToken()` we basically have `self.token` and `_login()` sets it. We can discuss making this interface more inline with the other RTLs but maybe in a separate PR?

jkaster: Separate PR makes sense to me.

python/looker_sdk/rtl/auth_session.py
jkaster: We could possibly formalize "authorization credential sets" but I think it's still too early in the auth change cycle for that. For now I think checking the "is configured" function to find out if all required properties are provided on SDK initialization and providing details in the error message if not could be the way to do this. @dthorpe may have other ideas.

PR #221, opened 2020-05-18
SDK generation for Looker 7.8

jkaster:
Updated SDK generated files to Looker 7.8 APIs


python/looker_sdk/sdk/api31/methods.py
```diff
@@ -6057,7 +6080,7 @@ def search_folders(
     #
     # Boolean search params accept only "true" and "false" as values.
     #
-    # GET /groups/search -> Sequence[models.Group]
+    # GET /groups/search -> Sequence[models.GroupSearch]
```
joeldodge79, line 6083: this feels like not a great change for a "stable" API :-(

jkaster: Agreed, but codegen is just a consumer. Should take up the issue with the endpoint owners.

PR #224, opened 2020-05-26
Improved error handling for Kotlin transport

jkaster:
SDK errors now report the operation that failed and uniformly throw a `java.lang.Error` exception.

- test added
- linty freshness

### DIFF
--- a/kotlin/src/main/com/looker/rtl/Transport.kt
+++ b/kotlin/src/main/com/looker/rtl/Transport.kt
@@ -28,6 +28,7 @@ import io.ktor.client.HttpClient
 import io.ktor.client.call.call
 import io.ktor.client.call.receive
 import io.ktor.client.engine.okhttp.OkHttp
+import io.ktor.client.features.ClientRequestException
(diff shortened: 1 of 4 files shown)


kotlin/src/main/com/looker/rtl/Transport.kt
```diff
@@ -282,10 +283,16 @@ class Transport(val options: TransportOptions) {
 //            }
 //        }
 
-        val result = runBlocking {
-            SDKResponse.SDKSuccessResponse(client.call(builder).response.receive<T>())
+        var result: SDKResponse
```
brianmadden, line 286: You can still use a `val` here I believe. You should be able to do:
```
val result = try {
            runBlocking {
                SDKResponse.SDKSuccessResponse(client.call(builder).response.receive<T>())
            }
        } catch (e: Exception) {
            SDKResponse.SDKErrorResponse("$method $path $e")
        } finally {
            client.close()
        }
```

https://kotlinlang.org/docs/reference/exceptions.html#try-is-an-expression

PR #225, opened 2020-05-28
Python3.6 sdk support

joeldodge79:
This change updates the generated models files as well as some RTL updates to support python3.6

* the python3.6 typing module uses _ForwardRef instead of ForwardRef

* datetime parsing: `%z` only works in >3.7, fall back to dateutil.parseiso otherwise

* add tox config:
  To test against all supported python versions you run: `pipenv run tox`
  (first you must make sure that you've pyenv installed the versions
  listed in .python-version)


python/tests/integration/api40/test_methods.py
```diff
@@ -267,7 +267,7 @@ def test_crud_look(sdk: mtds.Looker40SDK, looks):
                 is_run_on_load=l.get("is_run_on_load"),
                 public=l.get("public"),
                 query_id=query.id,
-                space_id=l.get("space_id") or str(sdk.me().personal_space_id),
+                folder_id=l.get("folder_id") or str(sdk.me().personal_folder_id),
```
joeldodge79, line 270: decided to finally fix the 4.0 tests

joeldodge79:
I'm not sure where to document this (contributing readme on main project?) But in order to `pipenv run tox` you need to do a one time `pyenv install 3.6.10 && pyenv install 3.7.6 && pyenv install 3.8.2`

josephaxisa:
> I'm not sure where to document this (contributing readme on main project?) But in order to `pipenv run tox` you need to do a one time `pyenv install 3.6.10 && pyenv install 3.7.6 && pyenv install 3.8.2`

I think having a contributing section [here](https://github.com/looker-open-source/sdk-codegen/blob/master/python/README.rst) makes sense.

PR #226, opened 2020-05-28
Added Method.rateLimited as a boolean property

jkaster:
- `Method.rateLimited` is `true` if any of the 7 rate limited attributes are set for the operation (schema) of a method

- `ApiModel.methods`, `ApiModel.refs`, `ApiModel.tags` (and methods contained therein), and `ApiModel.types` all now have their keys in sorted order when the model load completes

- removed `ApiModel.sortedMethods()` and `ApiModel.sortedTypes()` since they are no longer necessary for code generation of methods and models in locale-sorted order


packages/sdk-codegen/src/sdkModels.ts
```diff
@@ -1707,18 +1740,30 @@ export class ApiModel implements ISymbolTable, IApiModel {
...
+   * @param {IKeyedCollection<T>} list to sort
+   * @returns {IKeyedCollection<T>} newly sorted list
+   */
+  sortList<T>(list: IKeyedCollection<T>): IKeyedCollection<T> {
+    const result = {}
+    const sortedKeys = Object.keys(list).sort((a, b) => a.localeCompare(b))
+    for (const key of sortedKeys) {
+      result[key] = list[key]
+    }
+    return result
```
josephaxisa, line 1754: nice!

packages/scripts/src/sdkGenerator.ts
```diff
@@ -97,7 +97,7 @@ export class MethodGenerator extends Generator<Models.IApiModel> {
     const items: string[] = []
     // reset refcounts for ALL types so dynamic import statement will work
     Object.entries(this.model.types).forEach(([_, type]) => (type.refCount = 0))
```
josephaxisa, line 99: This caught my eye. What do you think of replacing `Object.entries` with `Object.values` here to avoid the slightly less easy on the eyes destructuring syntax? 
```suggestion
    Object.values(this.model.types).forEach(type => type.refCount = 0)
```

packages/sdk-codegen/src/sdkModels.ts
jkaster: I'd prefer a sorted dictionary type/interface but this will have to do for now

packages/scripts/src/sdkGenerator.ts
jkaster: This loop is going to need to change to iterate sorted collections in the right order

python/looker_sdk/sdk/api31/models.py
```diff
@@ -5323,6 +5323,181 @@ def __init__(self, *, inherits: Optional[bool] = None):
         self.inherits = inherits
 
 
+@attr.s(auto_attribs=True, kw_only=True)
```
joeldodge79, line 5326: is this against 7.8?

something is goofed up for the python generated code: Any `Write` model should have an `__init__` method. Also, it seems suspicious to me that there are so many new write classes in `3.1` (across all sdks). Looks like "id" got made read-only for `ContentValidation` things - we should probably investigate and follow up with the owner...

joeldodge79: another suspicious thing: with all these new (improperly generated) Write classes, nothing changed in `methods.py`...

jkaster: Thanks, Joel. I'll look into this some and let you know. It's probably something to do with the dynamic type generation after the spec is loaded. I think I made `loadDynamicTypes()` a bit too dumb/eager.

packages/sdk-codegen/src/sdkModels.spec.ts
```diff
@@ -170,7 +171,179 @@ describe('sdkModels', () => {
...
+    it('X-Rate-Limit-Reset', () => {
+      const yes = ({
+        'X-Rate-Limit-Reset': 0,
+      } as unknown) as OAS.OperationObject
+      const no = ({
+        'x-rate-limit-reset': 0,
+      } as unknown) as OAS.OperationObject
+      const nada = {} as OAS.OperationObject
+
+      expect(Method.isRateLimited(yes)).toBe(true)
```
dthorpe, line 328: What will applications use this `Method.isRateLimited()` for?

packages/sdk-codegen/src/sdkModels.spec.ts
```diff
@@ -170,7 +171,179 @@ describe('sdkModels', () => {
...
+        'x-looker-rate-limited': false,
+      } as unknown) as OAS.OperationObject
+      const nada = {} as OAS.OperationObject
+
+      expect(Method.isRateLimited(yes)).toBe(true)
+      expect(Method.isRateLimited(no)).toBe(false)
+      expect(Method.isRateLimited(nada)).toBe(false)
+    })
+
+    it('X-RateLimit-Limit', () => {
```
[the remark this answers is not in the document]
jkaster: *we* may not in Looker. the spec does, and codegen could be used for more than Looker's API spec. Also true for API Explorer, which consumes this package.

packages/sdk-codegen/src/sdkModels.spec.ts
jkaster: API Explorer will use it to indicate an endpoint is rate limited.

package.json
```diff
@@ -186,7 +186,6 @@
   },
   "keywords": [],
   "resolutions": {
-    "set-value": "^2.0.1",
-    "typescript": "3.8.2"
```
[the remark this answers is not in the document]
jkaster: Thanks for catching that. That should **definitely** still be there.

packages/scripts/package.json
```diff
@@ -47,8 +47,5 @@
     "prettier": "^2.0.5",
     "swagger2openapi": "^6.0.3"
   },
-  "resolutions": {
-    "typescript": "3.8.2"
```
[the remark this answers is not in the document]
jkaster: I had 3.8.2 pinned for all `package.json` files. Some `yarn` process must have removed them. Wiping my `node_modules` and restarting the yarn process from scratch to see if I need to put them back. Thanks for catching this.

packages/sdk-codegen/src/sdkModels.ts
```diff
@@ -82,29 +90,29 @@ export interface ISymbol {
   search(rx: RegExp, criteria: SearchCriteria): boolean
 }
 
-export type IKeyedCollection<T> = Record<string, T>
-export type IMethodList = IKeyedCollection<IMethod>
-export type ITypeList = IKeyedCollection<IType>
-export type ITagList = IKeyedCollection<IMethodList>
-export type IPropertyList = IKeyedCollection<IProperty>
-export type IKeyList = Set<string>
+export type KeyedCollection<T> = Record<string, T>
```
joeldodge79, line 93: is this just a style/convention change or is something more significant happening here?

packages/sdk-codegen/src/sdkModels.ts
```diff
@@ -886,10 +920,65 @@ export class Method extends SchemadSymbol implements IMethod {
...
+    /**
+     * if the body parameter is specified and is optional, at least 2 optional parameters are required
+     */
+    const offset = body && `required` in body && !body.required ? 1 : 0
+    return this.optionalParams.length - offset > 1
+  }
+
+  makeTypes(api: IApiModel): KeyList {
+    if (this.mayUseRequestType()) {
+      api.getRequestType(this)
```
joeldodge79, line 965: this reads a little funny to me. I've always felt that `getRequestType` **making** a request type if one doesn't exist is a little beyond the scope of the name. Reading this line the "command query separation" violation stands out to me even more. If breaking apart the getting and making is too cumbersomeI maybe we can settle for a kludgey name like `getOrCreateRequestType` ?

packages/sdk-codegen/src/sdkModels.ts
jkaster: Naming convention. Those originally were interfaces, then I changed them to types.

packages/sdk-codegen/src/sdkModels.ts
jkaster: Let's create a separate PR for this rename. The cross-package dependency issues with the rename are major until I get yarn workspace links working correctly

PR #231, opened 2020-06-06
Recognizing `any` in a specification format

jkaster:
This is a step closer to correct collection type generation.

All current SDK languages fortunately have a type that can be deserialized as an "any" type either natively or with an extension (e.g. Swift's `AnyCodable`), so until we clean up the Looker API specs to correctly indicate a hash or array's value type in all circumstances, these collections will be `any` instead.

We have some types, like `MergeQuery.vis_config` that incorrectly claim to be `string, string` when payloads are `string, any`. There are other offenders as well.

Until all those are cleaned up, this should help resolve any deserialization issues from innacurate payloads. I've tested it for Typescript, Kotlin, and Swift.

@joeldodge79 is planning on looking at this on Monday, so once he verifies it works for him, we can proceed. There's probably an import statement for python's Any that needs to be added to the code generator, and I'm available to help with that (due to codegen package dependency update difficulties right now) if that's necessary.


packages/sdk-codegen/src/python.gen.ts
```diff
@@ -105,6 +105,7 @@ export class PythonGen extends CodeGen {
     string: { default: this.nullStr, name: 'str' },
     uri: { default: this.nullStr, name: 'str' },
     void: { default: this.nullStr, name: 'None' },
+    any: { default: this.nullStr, name: 'Any' },
```
josephaxisa, line 108: The import statement in `methodsPrologue` and `modelsPrologue` need to be modified to include `Any` from the `typing` module. Right now it's not recognized causing `mypy` to complain.

Once that's done, `mypy` will still throw some 805 errors because of any generated model with at least one attribute annotated with `Any`. We might need to remove https://github.com/looker-open-source/sdk-codegen/blob/875d5cd9a0213abb5b257328b88602c43c618c36/python/mypy.ini#L4 as I don't see a way to dial this down a notch from [here](https://mypy.readthedocs.io/en/stable/command_line.html).

jkaster: Thanks for confirming an import statement is required. I hadn't yet researched which import was appropriate because I was trying to wrap up for the night.

This will require another round of change/publish/upgrade/test but I'll wait until Monday for it.

packages/sdk-codegen/src/python.gen.spec.ts
```diff
@@ -436,7 +436,7 @@ class Workspace(model.Model):
         id: The unique id of this user workspace. Predefined workspace ids include "production" and "dev"
         projects: The local state of each project in the workspace
     """
-    can: Optional[MutableMapping[str, bool]] = None
+    can: Optional[MutableMapping[str, Any]] = None
```
joeldodge79, line 439: I feel like this should remain `bool`

packages/sdk-codegen/src/python.gen.ts
```diff
@@ -499,7 +500,8 @@ ${this.hooks.join('\n')}
         case 'HashType':
           return {
             default: this.nullStr,
-            name: `MutableMapping[str, ${map.name}]`,
+            // TODO fix the invalid Looker API spec formats like MergeQuery's vis_config
+            name: `MutableMapping[str, Any]`, // `MutableMapping[str, ${map.name}]`,
```
joeldodge79, line 504: Is this over-reaching? shouldn't it remain `${map.name}`?

jkaster: Quite possibly. I can go less extreme and only remap string to Any

packages/sdk-codegen/src/python.gen.spec.ts
jkaster: fair enough. I'll take another pass

python/looker_sdk/sdk/api40/methods.py
```diff
@@ -4675,17 +4675,17 @@ def query_task(
     # Query Tasks whose results have expired will have a status of 'expired'.
     # If the user making the API request does not have sufficient privileges to view a Query Task result, the result will have a status of 'missing'
     #
-    # GET /query_tasks/multi_results -> MutableMapping[str, str]
+    # GET /query_tasks/multi_results -> MutableMapping[str, Any]
```
joeldodge79, line 4678: it still feels like we're over-reaching - in this case I think the results are actually meant to be `str, str`?

jkaster: The problem right now is the spec indicates `str, str` for hash types incorrectly for more than `vis_config`. Creating the "fix this" list is as much work as actually fixing it, so we lose *some* strong typing by downgrading to `Any` but most importantly things can be deserialized and also serialized back to what our endpoints require.

joeldodge79: sounds right.

python/looker_sdk/sdk/api31/methods.py
```diff
@@ -3682,7 +3682,7 @@ def fetch_integration_form(
     # POST /data_actions/form -> models.DataActionForm
     def fetch_remote_data_action_form(
         self,
-        body: MutableMapping[str, str],
+        body: MutableMapping[str, Any],
```
joeldodge79, line 3685: why don't the other languages have these `methods` updates to `fetch_remote_data_action_form` and `query_task_multi_results` ? I see that TS still uses `IDictionary<string>` in these cases.

joeldodge79: if this is correct then we need [this PR](https://github.com/looker-open-source/sdk-codegen/pull/233)

jkaster: As I mentioned above, TS doesn't need this because the JS run-time tolerates the ambiguity when parsing our JSON payloads (much like Ruby and untyped Python do).

Kotlin and Swift *should* have these updates as well. Maybe I'm misunderstanding?

```kotlin
  fun fetch_remote_data_action_form(
    /**
     * @param {Map<String,Any>} body
     */
    body: Map<String,Any>) : SDKResponse {
      return this.post<DataActionForm>("/data_actions/form", mapOf(), body)
  }
```

```swift
    func fetch_remote_data_action_form(
        /**
         * @param {StringDictionary<AnyCodable>} body
         */
        _ body: StringDictionary<AnyCodable>,
        options: ITransportSettings? = nil
    ) -> SDKResponse<DataActionForm, SDKError> {
        let result: SDKResponse<DataActionForm, SDKError> = self.post("/data_actions/form", nil, try! self.encode(body), options)
        return result
    }
```

joeldodge79: ah, I didn't see the changes in this PR for kotlin/swift so I assumed they were different but now I see they were changed in an earlier PR.

That makes sense for TS except that I'd suspect that the typing could potentially be wrong and perhaps require casting (`as`)?

jkaster: All Typescript type checking is gone at run-time, so if JS doesn't complain about the JSON payload format, it's not going to complain.

Because this is a hack workaround anyway (we need to fix our spec then back out this demotion of `string` to `any` for the generation of the SDKs that care about the payload type at run-time) I don't think doing the same hack for Typescript is needed.

PR #235, opened 2020-06-18
Look# experimental proof of concept

jkaster:
Initial SDK tests are working. Lots left to clean up, but since the run time and code generator produce something usable, we're done with this spike.


csharp/rtl/sdkrtl.csproj
```diff
@@ -0,0 +1,8 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
dthorpe, line 1: should that be Looker...?

jkaster: Good question! I'm not sure. Something to check on. Thanks.

PR #241, opened 2020-06-21
Enum support for OpenAPI and Looker spec enum types

jkaster:
## Counting on enums

This PR generates enums for all 5 language SDKs. If a parameter, property, or type has an `x-looker-values` or `enum` value list declaration, a corresponding `enum` type is created. Previously, parameters or properties that were usually typed as `string` are now an enum type.

### Examples
**C#**
```csharp
public enum PermissionType
{
  [EnumMember(Value = "view")]
  view,
  [EnumMember(Value = "edit")]
  edit
}
```

**Kotlin**
```kotlin
enum class PermissionType : Serializable {
  view,
  edit
}
```

**Python**
```python
class PermissionType(enum.Enum):
    view = "view"
    edit = "edit"
```

**Swift**
```swift
enum PermissionType: String, Codable {
    case view = "view"
    case edit = "edit"
}
```
**Typescript**

```ts
export enum PermissionType {
  view = 'view',
  edit = 'edit',
}
```

### TODO - Python unit test
The **Python SDK** still needs an enum property serialization/deserialization test. All other SDKs have at least one working test.

## BONUS: C# generator optimization
This PR also optimizes C# method generation so generic type specifiers are **only** required for responses that take more than one type.

```sh
yarn sdk c#
```
or
```sh
yarn sdk csharp
```
will generate the C# SDK code. C# is also generated with all other languages if a language specifier isn't included.


packages/sdk-codegen/src/sdkModels.spec.ts
```diff
@@ -132,9 +246,9 @@ describe('sdkModels', () => {
         apiTestModel.types.CreateQueryTask
       )
       expect(type).toBeDefined()
-      expect(type!.properties.query_id.required).toEqual(true)
-      expect(type!.properties.result_format.required).toEqual(true)
-      expect(type!.properties.source.required).toEqual(false)
+      expect(type?.properties.query_id.required).toEqual(true)
+      expect(type?.properties.result_format.required).toEqual(true)
+      expect(type?.properties.source.required).toEqual(false)
```
joeldodge79, line 251: does TS complain about `type` possibly being `undefined` inspite of the `toBeDefined` assertion above?

packages/sdk-codegen/src/sdkModels.spec.ts
```diff
@@ -172,36 +286,144 @@ describe('sdkModels', () => {
...
+        ],
+        description: 'RF5',
+        nullable: true,
+      }
+
+      it('enum types are renamed and not overwritten', () => {
+        const api = new ApiModel({} as OAS.OpenAPIObject)
+        const actual1 = api.resolveType(rf1)
+        const name1 = titleCase(rf1.name)
+        expect(actual1.name).toEqual(name1)
```
joeldodge79, line 373: I'm more a fan of just using the expected value (e.g. `"Rf1"` in this case ... I think): 
* I find it easier to read the test
* it's an additional check point in case you update `titleCase` impl and its tests but forget some edge case covered here perhaps

packages/sdk-codegen/src/sdkModels.spec.ts
```diff
@@ -172,36 +286,144 @@ describe('sdkModels', () => {
...
+
+        // Returns first enum for same values
+        const actual5 = api.resolveType(rf5, undefined, 'Foo')
```
joeldodge79, line 376: this is where I love languages that support named argument... what does `undefined` and `'Foo'` do here again? (I could go look but maybe a comment about why you're calling `api.resolveType` this way would be easier for a test reader to follow the logic

python/looker_sdk/sdk/api31/models.py
```diff
@@ -196,7 +242,7 @@ class ContentMetaGroupUser(model.Model):
     can: Optional[MutableMapping[str, bool]] = None
     id: Optional[str] = None
     content_metadata_id: Optional[str] = None
-    permission_type: Optional[str] = None
+    permission_type: Optional[models.PermissionType] = None
```
joeldodge79, line 245: need an additional tweak to python formatter: these type annotations need to be quoted ([lazy type evaluation](https://www.python.org/dev/peps/pep-0563/) is only 3.7+) and they're defined inside this same module so no need for the `models.` access:

```suggestion
    permission_type: Optional["PermissionType"] = None
```

in short, should be same as `ApiVersion` for example

joeldodge79: another option would be to leave out these python generated changes and merge it in. Then I can do the fixup to get the annotations correct as well as verify that the [de]serialization is working

packages/sdk-codegen/src/codeGen.ts
```diff
@@ -181,7 +191,11 @@ export abstract class CodeGen implements ICodeGen {
     return commentBlock(description, indent, this.commentStr)
   }
 
-  commentHeader(indent: string, text: string | undefined) {
+  commentHeader(
+    indent: string,
+    text: string | undefined,
+    _commentStr?: string
```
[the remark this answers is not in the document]
jkaster: It's for inserting the MIT license header into the generated source files. The Typescript `header/header` lint rule expects it to look like [license-header.js](config/license-header.js) so I modified the comment function to accept the type of internal comment string to use.

packages/sdk-codegen/src/sdkModels.spec.ts
jkaster: yeah, sadly the jest `toBeDefined()` test is not a signal to the Typescript linter.

packages/sdk-codegen/src/sdkModels.spec.ts
jkaster: Fair enough. This was a TDD-ish test I was using to check the renaming rules more than the casing rules. I can hard code the expectation. Pins the logic better.

packages/sdk-codegen/src/sdkModels.spec.ts
jkaster: Comment added

python/looker_sdk/sdk/api31/models.py
jkaster: Let's hold off on merging this PR until Python is right, too.

jkaster: Got the quote in there, but forwardrefs for enum types may still need to be generated?

packages/sdk-codegen/src/python.gen.ts
```diff
@@ -132,6 +135,7 @@ class ${this.packageName}(api_methods.APIMethods):
 # ${this.warnEditing()}
 import datetime
 from typing import Any, MutableMapping, Optional, Sequence
+import enum
```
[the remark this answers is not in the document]
jkaster: I fixed this and pushed the change, but for some reason Github is refusing to update the status of this comment. 😖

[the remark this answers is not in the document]
jkaster: I also put

```python
import attr
```

first, but GitHub is absolutely refusing to show the code updates

PR #242, opened 2020-06-22
Python SDK: Enum Support

josephaxisa:
- Made model attributes of type enum string forward references with a registered structure hook
- Added a unit test for a model containing an enum type attribute
- Broke up structure_hook into reserved_kw_structure_hook and forward_ref_structure_hook


python/.python-version
```diff
@@ -1,3 +1,3 @@
-3.6.10
-3.7.6
 3.8.2
+3.7.6
+3.6.10
```
jkaster, line 3: Does this change default python version selection?

jkaster:
> It might be easiest to go over this with you guys in a meeting for better understanding.

That would be fine with me!

packages/sdk-codegen/src/codeGen.ts
```diff
@@ -235,9 +239,9 @@ export abstract class CodeGen implements ICodeGen {
         )
         propertyValues = props.join(this.enumDelimiter)
       } else {
-        Object.values(type.properties).forEach((prop) =>
-          props.push(this.declareProperty(bump, prop))
-        )
+        Object.values(type.properties)
+          .sort(this.sortProperties)
```
jkaster, line 243: what's this sort call for?

packages/sdk-codegen/src/codeGen.ts
```diff
@@ -223,6 +223,10 @@ export abstract class CodeGen implements ICodeGen {
     return this.itself ? `${this.itself}.${value}` : value
   }
 
+  sortProperties(_a: IProperty, _b: IProperty) {
+    return 0
+  }
```
jkaster, line 228: Did you intend to leave this in? I realize it's not sorting anything right now, but I don't think we should be sorting properties for type declarations. This destroys any intentional grouping/order the developer may have created.

packages/sdk-codegen/src/python.gen.ts
```diff
@@ -441,11 +481,12 @@ ${this.hooks.join('\n')}
...
     const isEnum = type instanceof EnumType
     const baseClass = isEnum ? 'enum.Enum' : 'model.Model'
-    let attrsArgs = 'auto_attribs=True, kw_only=True'
     let usesReservedPythonKeyword = false
 
     if (!isEnum) {
-      for (const prop of Object.values(type.properties)) {
+      for (const prop of Object.values(type.properties).sort(
+        this.sortProperties
+      )) {
```
jkaster, line 489: We should not be sorting properties of a type.

python/tests/rtl/test_serialize.py
```diff
@@ -40,64 +42,183 @@
...
-    finally_: Optional[Sequence["ChildModel"]] = None
+    finally_: Optional[Sequence[int]] = None
+
+    # Because this model has "bare" forward ref annotated properties
+    # (enum1, enum2, model_no_refs1, and model_no_refs2) we need to tell
+    # the attr library that they're actually ForwardRef objects so that
+    # cattr will match our forward_ref_structure_hook structure hook
+    #
+    #
+    # Note: just doing a the following:
```
jkaster, line 108: nit
```suggestion
    # Note: just doing the following:
```

python/tests/rtl/test_serialize.py
```diff
@@ -40,64 +42,183 @@
...
+        id: Optional[int] = None,
         name: Optional[str] = None,
         class_: Optional[str] = None,
-        finally_: Optional[Sequence["ChildModel"]] = None
+        finally_: Optional[Sequence[int]] = None,
     ):
+        """Keep mypy and IDE suggestions happy.
+
+        We cannot use the built in __init__ generation attrs offers
+        because mypy complains about unkown keyword argument, even
```
jkaster, line 154:
```suggestion
        because mypy complains about unknown keyword argument, even
```

packages/sdk-codegen/src/codeGen.ts
joeldodge79: I should have just closed this PR till it was ready to review. I was going to go through and annotate the PR with why I made certain changes, including this one:

In python I reverted to having all classes define their own `__init__` for consistency and readability. That exposed a bug:

when relying on `kw_only=True` argument to the `attr.s()` class decorator, the order of required vs optional property declarations didn't matter. But now that we're not using that argument, and instead passing `init=False` - `attr` blows up saying that "optional declarations cannot come before required ones". So, in order for the python sdk to "compile" we need to make sure that all required properties are bubbled up to the top.

packages/sdk-codegen/src/codeGen.ts
joeldodge79: yes, that was intended.

packages/sdk-codegen/src/python.gen.ts
joeldodge79: let's talk about it - I can try to find another solution for python but that might be a bit of work

packages/sdk-codegen-scripts/src/convert.ts
```diff
@@ -213,7 +213,7 @@ export const convertSpec = (
     // patch to fix up small errors in source definition (not required, just to ensure smooth process)
     // indent no spaces
     // output to openApiFilename
-    run('swagger2openapi', [
+    run('yarn swagger2openapi', [
```
joeldodge79, line 216: part of versionstamp bugfix

packages/sdk-codegen-scripts/src/fetchSpec.spec.ts
```diff
@@ -60,6 +60,11 @@ describe('fetch operations', () => {
     expect(actual).not.toEqual('')
   })
 
+  it('gets lookerVersion with supplied versions', async () => {
+    const actual = await fetchLookerVersion(props, {looker_release_version: '7.10.0'})
+    expect(actual).toEqual('7.10')
+  })
+
```
joeldodge79, line 67: part of versionstamp bugfix

packages/sdk-codegen-scripts/src/fetchSpec.ts
```diff
@@ -289,17 +289,16 @@ export const fetchLookerVersion = async (
...
-      const matches = versions.looker_release_version.match(/^\d+\.\d+/i)
-      lookerVersion = matches[0]
     } catch (e) {
       warn(`Could not retrieve looker release version: ${e.message}`)
+      return ''
     }
   }
-  return lookerVersion
+  const matches = versions.looker_release_version.match(/^\d+\.\d+/i)
+  return matches[0]
```
joeldodge79, line 301: part of versionstamp bugfix

packages/sdk-codegen/src/codeGen.ts
```diff
@@ -235,7 +239,7 @@ export abstract class CodeGen implements ICodeGen {
         )
         propertyValues = props.join(this.enumDelimiter)
       } else {
-        Object.values(type.properties).forEach((prop) =>
+        this.typeProperties(type).forEach((prop) =>
```
joeldodge79, line 242: this was my solution to allow custom behavior for language generators and keep the previous behavior the default. open to other ways if this doesn't seem right.

packages/sdk-codegen/src/python.gen.spec.ts
```diff
@@ -450,14 +465,13 @@ return response`
...
 
-  // TODO reinstate these tests after package dependencies are updated correctly
   describe('type creation', () => {
     it('with arrays and hashes', () => {
       const type = apiTestModel.types.Workspace
       const actual = gen.declareType(indent, type)
       expect(type.properties.id.type.name).toEqual('string')
       expect(actual).toEqual(`
-@attr.s(auto_attribs=True, kw_only=True)
+@attr.s(auto_attribs=True, init=False)
```
joeldodge79, line 474: all models now have these arguments

packages/sdk-codegen/src/python.gen.spec.ts
```diff
@@ -467,14 +481,159 @@ class Workspace(model.Model):
...
-    projects: Optional[Sequence["Project"]] = None`)
+    projects: Optional[Sequence["Project"]] = None
+
+    def __init__(self, *,
+            can: Optional[MutableMapping[str, bool]] = None,
+            id: Optional[str] = None,
+            projects: Optional[Sequence["Project"]] = None):
+        self.can = can
+        self.id = id
+        self.projects = projects`)
```
joeldodge79, line 492: and all models now have a constructor

packages/sdk-codegen/src/python.gen.spec.ts
```diff
@@ -467,14 +481,159 @@ class Workspace(model.Model):
...
+        listens_to_filters: Array of listeners for faceted filters
+        allow_multiple_values: Whether the filter allows multiple filter values
+        required: Whether the filter requires a value to run the dashboard
+        ui_config: The visual configuration for this filter. Used to set up how the UI for this filter should appear.
+    """
+    dashboard_id: str
+    name: str
+    title: str
+    type: str
+    id: Optional[str] = None
```
joeldodge79, line 522: it was both `mypy` as well as the `attr` runtime that complained about actual class level "optional before required" ordering.

packages/sdk-codegen/src/python.gen.spec.ts
```diff
@@ -467,14 +481,159 @@ class Workspace(model.Model):
...
+        sql_on: SQL ON expression describing the join condition
+        sql_table_name: SQL table name to join
+        type: The join type: left_outer, full_outer, inner, or cross
+        view_label: Label to display in UI selectors
+    """
+    name: Optional[str] = None
+    dependent_fields: Optional[Sequence[str]] = None
+    fields: Optional[Sequence[str]] = None
+    foreign_key: Optional[str] = None
+    from_: Optional[str] = None
```
joeldodge79, line 593: Let's make sure our keyword translation generation is happening

packages/sdk-codegen/src/python.gen.spec.ts
```diff
@@ -513,6 +680,43 @@ class PermissionType(enum.Enum):
...
+        roles: Roles assigned to group
+    """
+    query_id: int
+    result_format: "ResultFormat"
+    user: "UserPublic"
+    roles: Optional[Sequence["Role"]] = None
+    __annotations__ = {
+        "query_id": int,
+        "result_format": ForwardRef("ResultFormat"),
+        "user": ForwardRef("UserPublic"),
```
joeldodge79, line 704: only classes that have these "bare forward ref" types need the `__annotations__` correction

packages/sdk-codegen/src/python.gen.spec.ts
```diff
@@ -536,7 +740,7 @@ class PermissionType(enum.Enum):
       })
       const actual = gen.declareType(indent, inputType)
       expect(actual).toEqual(`
-@attr.s(auto_attribs=True, kw_only=True, init=False)
```
joeldodge79, line 539: cruft: `kw_only` is ignored when `init=False` is supplied

packages/sdk-codegen/src/python.gen.spec.ts
```diff
@@ -76,21 +82,30 @@ class LookerSDK(api_methods.APIMethods):
...
+forward_ref_structure_hook = functools.partial(sr.forward_ref_structure_hook, globals(), sr.converter)
+translate_keys_structure_hook = functools.partial(sr.translate_keys_structure_hook, sr.converter)
+sr.converter.register_structure_hook(
+    ForwardRef("LookmlModelExploreJoins"),  # type: ignore
+    forward_ref_structure_hook  # type:ignore
+)
+sr.converter.register_structure_hook(
+    LookmlModelExploreJoins,  # type: ignore
+    translate_keys_structure_hook  # type:ignore
+)
```
joeldodge79, line 108: running `gen.declareType` above so that we can make sure we're registering the `translate_keys_structure_hook` on models. We weren't before and I think there was a silent bug with `LookmlModelExploreJoins` where `from_`/`"from"` was getting lost in translation...

the unittests passed because we explicitly register this structure hook on "Model"

packages/sdk-codegen/src/python.gen.spec.ts
```diff
@@ -50,11 +50,17 @@ describe('python generator', () => {
...
 import enum
 from typing import Any, MutableMapping, Optional, Sequence
 
+try:
+    from typing import ForwardRef  # type: ignore
+except ImportError:
+    from typing import _ForwardRef as ForwardRef  # type: ignore
+
+import attr
+
```
joeldodge79, line 63: need this in the prologue now that we use `ForwardRef` in some `__annotations__` in model declarations

packages/sdk-codegen/src/python.gen.ts
```diff
@@ -183,17 +195,30 @@ ${this.hooks.join('\n')}
...
+      let annotation = propType
+      if (this.isBareForwardRef(property)) {
+        annotation = `ForwardRef(${propType})`
+      }
+      propDef = `${this.bumper(indent)}"${propName}": ${annotation}`
+    } else {
+      if (!property.required) {
+        propType += ` = ${this.nullStr}`
+      }
+      propDef = `${indent}${propName}: ${propType}`
```
joeldodge79, line 216: dual purposing `declareProperty` to also declare an entry in the `__annotations__` dict. If that feels too icky I could probably factor out the shared bit (getting the type and translating when needed) and have `declareProperty` and `declareAnnotation`(?) use it. I'll admit I was rushing through this part and reading it now it's kind of hard to tease apart what's what/when/where

packages/sdk-codegen/src/python.gen.ts
```diff
@@ -236,20 +261,10 @@ ${this.hooks.join('\n')}
...
 
-  private addMethodInputModelType(type: IType) {
-    this.methodInputModelTypes.add(type)
-    for (const prop of Object.values(type.properties)) {
-      if (prop.type.elementType) {
-        this.addMethodInputModelType(prop.type.elementType)
-      }
-    }
-  }
-
```
joeldodge79, line 247: again, all models now get a constructor, not just the "input" types

packages/sdk-codegen/src/python.gen.ts
```diff
@@ -263,9 +278,28 @@ ${this.hooks.join('\n')}
   }
 
   initArg(indent: string, property: IProperty) {
-    return `${indent}self.${property.name} = ${property.name}`
+    let propName = property.name
+    if (this.pythonKeywords.has(propName)) {
+      propName = propName + '_'
+    }
+    return `${indent}self.${propName} = ${propName}`
```
joeldodge79, line 281: discovered this bug when `LookmlModelExploreJoins` got a constructor (it was getting `self.from = from` which is not valid python syntax)

packages/sdk-codegen/src/python.gen.ts
```diff
@@ -263,9 +278,28 @@ ${this.hooks.join('\n')}
...
+  typeProperties(type: IType) {
+    return Object.values(type.requiredProperties).concat(Object.values(type.optionalProperties))
   }
 
+  private isBareForwardRef = (prop: IProperty) => {
+    // contains a "bare" forward reference e.g. `result_format: "ResultFormat"`
+    // so we need to emit the `__annotations__` property
+    return (
+      prop.required &&
+      (prop.type.customType || prop.type.className == 'EnumType') &&
```
joeldodge79, line 293: I believe `(prop.type.customType || prop.type.className == 'EnumType')` captures all the generated models/enums - i.e. none of the IntrinsicType

packages/sdk-codegen/src/python.gen.ts
```diff
@@ -263,9 +278,28 @@ ${this.hooks.join('\n')}
...
+      !['ArrayType', 'HashType', 'DelimArrayType'].includes(
+        prop.type.className
+      )
```
joeldodge79, line 296: and that this further filters the result to a "singular"/"bare" model reference. e.g.

`role: "Role"`

Not `role: Sequence["Role"]` or `role: MutableMapping[str, "Role]`

packages/sdk-codegen/src/python.gen.ts
```diff
@@ -263,9 +278,28 @@ ${this.hooks.join('\n')}
...
+
+  typeProperties(type: IType) {
+    return Object.values(type.requiredProperties).concat(Object.values(type.optionalProperties))
   }
 
+  private isBareForwardRef = (prop: IProperty) => {
+    // contains a "bare" forward reference e.g. `result_format: "ResultFormat"`
+    // so we need to emit the `__annotations__` property
+    return (
+      prop.required &&
```
joeldodge79, line 292: i.e. not `role: Optional["Role"]`

packages/sdk-codegen/src/python.gen.ts
```diff
@@ -276,17 +310,25 @@ ${this.hooks.join('\n')}
...
+        annotations.push(this.declareProperty(indent, prop, true))
+        args.push(this.declareConstructorArg('', prop))
+        inits.push(this.initArg(bump, prop))
+      })
+    let result = ''
+    if (Object.values(type.properties).some(this.isBareForwardRef)) {
+      result = `\n${indent}__annotations__ = {\n`
+      result += `${annotations.join(',\n')}`
+      result += `\n${indent}}`
+    }
```
joeldodge79, line 326: prepending the `__annotations__` declaration to the constructor - feels a little out of place as it's more in the `declareType` section but I didn't want to override/copy that method in here to modify it.

packages/sdk-codegen/src/python.gen.ts
```diff
@@ -299,7 +341,11 @@ ${this.hooks.join('\n')}
     } else {
       propType = `Optional[${mappedType.name}] = ${this.nullStr}`
     }
-    return `${indent}${property.name}: ${propType}`
+    let propName = property.name
+    if (this.pythonKeywords.has(propName)) {
+      propName = propName + '_'
+    }
+    return `${indent}${propName}: ${propType}`
```
joeldodge79, line 344: again, `LookmlModelExploreJoins` getting a constructor exposed a bug here in the constructor signature

packages/sdk-codegen/src/python.gen.ts
```diff
@@ -457,25 +502,21 @@ ${this.hooks.join('\n')}
...
-
-      const forwardRef = `ForwardRef("${type.name}")`
+    const forwardRef = `ForwardRef("${type.name}")`
+    this.hooks.push(
+      `sr.converter${this.apiRef}.register_structure_hook(\n${bump}${forwardRef},  # type: ignore\n${bump}${this.structureHookFR}  # type:ignore\n)`
+    )
+    if (usesReservedPythonKeyword) {
       this.hooks.push(
-        `sr.converter${this.apiRef}.register_structure_hook(\n${bump}${forwardRef},  # type: ignore\n${bump}${this.structureHook}  # type:ignore\n)`
+        `sr.converter${this.apiRef}.register_structure_hook(\n${bump}${type.name},  # type: ignore\n${bump}${this.structureHookTK}  # type:ignore\n)`
```
joeldodge79, line 509: looks like we actually were registering the "translate_keys_structure_hook" functionality before, so `LookmlModelExploreJoins` probably was working, but now we're testing it :-)

packages/sdk-codegen/src/python.gen.ts
```diff
@@ -441,11 +487,10 @@ ${this.hooks.join('\n')}
     const attrs: string[] = []
     const isEnum = type instanceof EnumType
     const baseClass = isEnum ? 'enum.Enum' : 'model.Model'
-    let attrsArgs = 'auto_attribs=True, kw_only=True'
```
joeldodge79, line 444: now it's always `auto_attribs=True, init=False` so no need to have a separate variable

packages/sdk-codegen/src/sdkModels.ts
```diff
@@ -1392,6 +1402,26 @@ export class Type implements IType {
...
+  }
+
+  get requiredProperties() {
+    return this.filterRequiredProps(true)
+  }
+
+  get optionalProperties() {
+    return this.filterRequiredProps(false)
+  }
+
```
joeldodge79, line 1424: is this what you had in mind @jkaster ?

python/looker_sdk/rtl/serialize.py
```diff
@@ -96,21 +91,41 @@ def serialize(api_model: TModelOrSequence) -> bytes:
...
-    for reserved in keyword.kwlist:
-        if reserved in data:
-            data[f"{reserved}_"] = data.pop(reserved)
-    if isinstance(type_, ForwardRef):
-        type_ = eval(type_.__forward_arg__, context, locals())
-    instance = converter.structure_attrs_fromdict(data, type_)
+    data = _tr_data_keys(data)
+    actual_type = eval(forward_ref.__forward_arg__, context, locals())
+    if issubclass(actual_type, enum.Enum):
+        instance = converter.structure(data, actual_type)
```
joeldodge79, line 123: handle `enum: "Enum"` type declarations in deserialization (well, with the help of the `__annotations__` hack converting that to `enum: ForwardRef("Enum")`)

yarn.lock
```diff
@@ -14144,7 +14144,7 @@ typedarray@^0.0.6:
   resolved "https://registry.yarnpkg.com/typedarray/-/typedarray-0.0.6.tgz#867ac74e3864187b1d3d47d996a78ec5c8830777"
   integrity sha1-hnrHTjhkGHsdPUfZlqeOxciDB3c=
 
-typescript@3.8.2, typescript@^3.8.2, typescript@^3.8.3:
+typescript@3.8.2, typescript@^3.6.0, typescript@^3.8.2, typescript@^3.8.3:
```
joeldodge79, line 14147: I have no idea why I have this change, should I discard it? maybe I have some old version of typescript on my cloudtop?

packages/sdk-codegen/src/codeGen.ts
jkaster: I think this looks reasonable

packages/sdk-codegen/src/python.gen.ts
jkaster: tests are good!

packages/sdk-codegen/src/python.gen.spec.ts
jkaster: Good to know

packages/sdk-codegen/src/python.gen.spec.ts
jkaster: I really appreciate all the comments. Very helpful!

packages/sdk-codegen/src/python.gen.ts
jkaster: Did you need to change the signature of all the other `ICodeGen` implementations for this? Having it as a default `false` parameter seems ok to me. Make me think about "overloading" of annotations with the attributes for JSON property renames, but that's really an implementation detail someone could look at in this function.

packages/sdk-codegen/src/python.gen.ts
jkaster: Fwiw, I created a `reserve(name)` function in some of the other generators to abstract the "fix this reserved word" problem

packages/sdk-codegen/src/python.gen.ts
jkaster: I used 
```ts
prop.type instanceof EnumType 
```
as the check for the other enum generators. FWIW, I do preserve the base type of the `EnumType` as the element type so we know whether the enum is a string or an int. The Looker spec **only** has string enums, though.

packages/sdk-codegen/src/python.gen.ts
jkaster: fwiw this could be
```ts
      result = `\n${indent}__annotations__ = {\n` +
        `${annotations.join(',\n')}` +
        `\n${indent}}`
```
but you probably know that

packages/sdk-codegen/src/sdkModels.ts
jkaster: yup!

python/looker_sdk/rtl/serialize.py
jkaster: Sorry Python doesn't have a better built-in serialization solution!

yarn.lock
jkaster: No, don't discard it. Let `yarn.lock` do what `yarn.lock` will

josephaxisa:
I cannot start a review since I opened this PR, but this LGTM. Nice work Joel!

packages/sdk-codegen/src/python.gen.ts
joeldodge79: no, I was surprised that TS let me change this signature w/out complaining that the class no longer conformed to `ICodeGen` - I'm not deeply familiar with how that is supposed to work. Reading https://www.typescriptlang.org/docs/handbook/interfaces.html#implementing-an-interface didn't help... just a bit more confused

packages/sdk-codegen/src/python.gen.ts
joeldodge79: yeah, I was gonna dig more into how we're doing this in python when I look into implementing the kebab stuff. I'd like to try to make it conform more to what you have for the other language generators

packages/sdk-codegen/src/python.gen.ts
joeldodge79: so should I do this?

```suggestion
      (prop.type.customType || prop.type instanceof EnumType) &&
```

packages/sdk-codegen/src/python.gen.ts
joeldodge79: oh right, I find that my brain get's a little overloaded when dealing with intricacies of one language and then switching to another - and I forget simpler/stylistic stuff like this :-)

python/looker_sdk/rtl/serialize.py
joeldodge79: oh, me too. Python was my gold standard scripting language for a while but the Typing implementation and this lack of easy/builtin class-instance/json serialization has disillusioned me. I'd say Typescript is quickly gaining ground for me except the huge barrier between "I wrote this typescript code" and "I want to run this typescript script" (even as a node script) is still daunting to me.

packages/sdk-codegen/src/python.gen.ts
jkaster: That would work. I don't think you even need the customType test because it won't be an EnumType unless it is an enum

joeldodge79: The goal is to catch model types like "User" as well as enum types like "ResultFormat" but not int/str/bool etc types. Can I do `!(prop.type instanceof IntrinsicType)` ?

python/looker_sdk/rtl/serialize.py
jkaster: Yeah, the scripting in this repo is a bit awkward now, but if you look at the existing node scripts in the root package.json you can see the pattern that should always work.

packages/sdk-codegen/src/python.gen.ts
jkaster: if you're trying to ignore just intrinsic types, yes. You can use `instanceof` or `!type.intrinsic`

packages/sdk-codegen/src/python.gen.ts
jkaster: LOL. Let's worry about it if it breaks :)

packages/sdk-codegen/src/python.gen.ts
```diff
@@ -263,7 +274,27 @@ ${this.hooks.join('\n')}
...
+      Object.values(type.optionalProperties)
+    )
+  }
+
+  private isBareForwardRef = (prop: IProperty) => {
+    // contains a "bare" forward reference e.g. `result_format: "ResultFormat"`
+    // so we need to emit the `__annotations__` property
+    return (
+      prop.required &&
+      (prop.type.customType || prop.type instanceof EnumType) &&
```
jkaster, line 295: Oh, EnumType should definitely have a truthy `customType`. If it's not, we should get around to fixing that.

packages/sdk-codegen/src/sdkModels.ts
```diff
@@ -304,6 +304,16 @@ export interface IType extends ISymbol {
    */
   properties: PropertyList
 
+  /**
+   * key/value collection of required properties for this type
+   */
+  requiredProperties: PropertyList
```
jkaster, line 310: Do you need this as a list, or if it's just for natural order iteration, it could be `IProperty[]`

packages/sdk-codegen/src/python.gen.ts
joeldodge79: I think it doesn't which is why I had to add the EnumType check in here.

python/.python-version
joeldodge79: sorry, missed this question: yes, the first one in the list is the default version when you cd into the directory. There are more than one version listed so that `tox` can find them and use them

packages/sdk-codegen/src/sdkModels.ts
joeldodge79: I wanted it to be the same type as `properties` - `ICodeGen.typeProperties` returns a `IProperty[]` because that's all `declareType` cares about is natural order iteration

PR #243, opened 2020-06-23
Separate package for TryIt 

josephaxisa:


package.json
```diff
@@ -86,7 +86,7 @@
     "eslint-plugin-mdx": "^1.6.8",
     "eslint-plugin-prettier": "^3.1.3",
     "eslint-plugin-sort-keys-fix": "^1.1.1",
-    "jest": "^25.1.0",
+    "jest": "^26.0.1",
```
jkaster, line 89: I wonder if this works with newer Typescript versions?

packages/api-explorer/src/scenes/MethodScene/utils.ts
```diff
@@ -32,7 +32,7 @@ import {
   IType,
   IMethod,
 } from '@looker/sdk-codegen'
-import { TryItInput } from '../../components'
+import { TryItInput } from '../../../../try-it'
```
jkaster, line 35: I think this is the wrong import now? should be `@looker/try-it`?

PR #246, opened 2020-06-25
Special variable name support for JSON serialization/deserialization

jkaster:
Support for "special" symbol names 

- symbols could have spaces or dashes in the name
- `ISymbol.jsonName` is empty unless the name needs to be made safe. If that's the case, `ISymbol.name` will have the "safe" version of the name and `ISymbol.jsonName` will have the original version for decorating JSON serialization attributes
- `ISymbol.hasSpecialNeeds` is `true` if it's not a "normal" symbol name
- `sdkModels.ts` has a `safeName` function now
- There's no conversion for Unicode characters (because we don't need it yet) but the only place this handling would be required should be in `safeName`
- Typescript, Swift, Kotlin, and C# all support special name serialization and deserialization now
- There's a `HyphenType` entry in the test `openApiRef.json` that can be used to test the generators


packages/sdk-codegen/src/sdkModels.spec.ts
```diff
@@ -71,38 +73,62 @@ describe('sdkModels', () => {
...
+      expect(camelCase(actual)).toEqual(expected)
     })
-    it('FOOBAR is FOOBAR', () => {
-      expect(camelCase('FOOBAR')).toEqual('FOOBAR')
+  })
+
+  describe('safeName', () => {
+    it.each<[string, string]>([
+      ['', ''],
+      ['foo-bar', 'foo_bar'],
```
joeldodge79, line 98: do we only need to worry about kebab and space names?

swift/looker/Tests/lookerTests/transportTests.swift
```diff
@@ -16,15 +16,15 @@ struct SimpleUser : SDKModel {
 
 @available(OSX 10.15, *)
 class transportTests: XCTestCase {
-
+
```
joeldodge79, line 19: trailing whitespace?

kotlin/src/main/com/looker/sdk/4.0/models.kt
```diff
@@ -41,16 +42,17 @@ data class AccessToken (
   /**
    * Access Token used for API calls (read-only)
    */
+  @param:JsonProperty("access_token") @get:JsonProperty("access_token")
```
joeldodge79, line 45: why did this show up? seems like `name == safeName(name)`

packages/sdk-codegen/src/typescript.gen.spec.ts
```diff
@@ -351,16 +384,75 @@ body: ICreateDashboardRenderTask`)
...
+        const expected = `/**
+ * ### Get information about the current user; i.e. the user account currently calling the API.
+ *
+ * GET /user -> IUser
+ */
+async me(
+  /**
+   * @param {string} hi-test Requested fields.
+   */
+  'hi-test'?: string,
```
joeldodge79, line 417: interesting. how do you access this parameter inside the function?

swift/looker/Tests/lookerTests/transportTests.swift
jkaster: XCode editor FTW

kotlin/src/main/com/looker/sdk/4.0/models.kt
jkaster: whoops. that was some test code to verify imports and behavior. It will disappear when I regenerate. Good catch!

packages/sdk-codegen/src/typescript.gen.spec.ts
jkaster:
```typescript
method['hi-test']
```

packages/sdk-codegen/src/sdkModels.spec.ts
jkaster: To answer your top-level question: I'm going to look into the possibility of aliasing the json keys for the Typescript information. However, I think that's going to require completely changing the way we do the types.

Regarding conversion to `safeName()` .. as I mentioned in `TODO` for that function, there's no unicode recognition or processing right now. Our near-term need is to support hyphens. I just thought we should support spaces as well because someone might try to use that.

I think any special symbol names in an API spec is a bad idea when so many programming languages will have issues with them, so the current implementation is a bit of a forcing function that will cause the code generator to break when symbol names start to get too "out there." This will prompt a conversation with the implementer of the questionable endpoint implementation.

packages/sdk-codegen/src/typescript.gen.spec.ts
joeldodge79: go figure

jkaster:
> Also, how do you feel about python handling this translation purely in the RTL with no special generated annotations?

I would be very interested in learning how you'd do that

PR #252, opened 2020-07-01
BrowserSession now uses CORS and OAuth

jkaster:
- made `yarn register` script for `api-explorer` to automatically register itself as a CORS app. See `register.ts` for details
- Working through requirements, but need to get `refresh_token` assigned. I think I need an automatic login implementation


jkaster:
The next step is getting the `OAuthScene` to respond correctly to the Looker OAuth login redirect and login the sdk user for TryIt calls.

packages/api-explorer/src/scenes/OAuthScene/OAuthScene.tsx
```diff
@@ -0,0 +1,29 @@
...
+        console.error(err)
+      }
+    }
+
+    login().then(() => console.log('OAuth login completed'))
+  })
+
+  // TODO display OAuth login completed with a click to redirect, and redirect automatically after 3 seconds or so?
+  // TODO Or just show "OAuth login completed" as a disappearing banner on the redirected page?
+  return <Redirect to="/" />
```
josephaxisa, line 28: We might want to redirect to `/:specKey` here otherwise the redirect will result in another redirect because of:
https://github.com/looker-open-source/sdk-codegen/blob/35e28c22e09471ca2a0efd24a4e14b4e2222282f/packages/api-explorer/src/routes/AppRouter.tsx#L54-L55
unless react-router does some optimization under the hood.

I know this is still a WIP but you'll probably also need a local state here to cause a rerender upon logging in. Perhaps something along the lines of:

```Typescript
export const OAuthScene: React.FC<OAuthSceneProps> = ({ sdk, specKey }) => {
   const [isAuthenticated, setIsAuthenticated] = useState(false)
   const [loading, setLoading] = useState(true)
    
   ...
 
   return (
   <>
     {loading && <Text>Establishing OAuth session...</Text>
     {!loading && !isAuthenticated && 
       <Text>
         Failed to establish OAuth session. Ensure your app is registered.
       </Text>
     // whichever scenario we end up going with from the two described in your TODO:
     {isAuthenticated && <Redirect to=`/${specKey} />} 
   </>
   )
}
```

jkaster: Thanks for the tips and sample code!

packages/api-explorer/src/scenes/OAuthScene/OAuthScene.tsx
```diff
@@ -72,9 +77,7 @@ export const OAuthScene: React.FC<OAuthSceneProps> = ({ sdk }) => {
           </Text>
         </>
       )}
-      {!loading && sdk && sdk.authSession.isAuthenticated() && (
-        <Redirect to={newUrl} />
-      )}
+      {!loading && auth && auth.isAuthenticated() && history.push(newUrl)}
```
josephaxisa, line 80: I think `history.push()` is working but it's not matching any of our defined routes. 

I also think that this line can be removed and replaced with `history.push(newUrl)` on line 66. That way, we reserve the value of the return statement for JSX content while the component logic resides in the useEffect hook.

PR #258, opened 2020-07-13
Kotlin gradle default action only builds jar

jkaster:
- Use `gradle test` to run the Kotlin SDK tests
- Updated Gradle to 6.5.1


jkaster:
Agreed on the number of comments, but I've got another ticket I'll address in the next sprint that will resolve that WIP.

PR #261, opened 2020-07-14
production build configs for api-explorer and extension-api-explorer packages

josephaxisa:
- Added webpack production build configs for the `api-explorer` and `extension-api-explorer` packages
- Pointed webpack to the root babel config and removed  the one in`api-explorer`
- Added production build script for both packages (`yarn package`)
- Extended and reused the root jest config in both packages. 
- Addressed issues relating to eslint config rule overrides and consequently fixed the pre-commit hook
- Also added a no-var-requires override for JS files since this is often needed in config files. JS files in this monorepo should either be config files or a result of transpiled Typescript, with the latter existing in directories ignored by the linter.


josephaxisa:
Perhaps there should be a base webpack config in the `api-explorer` and extend that in `extension-api-explorer` as the two are identical at the moment. Thoughts?

whscullin:
Yeah, that will probably make maintenance going forward easier.

packages/api-explorer/webpack.dev.config.js
```diff
@@ -0,0 +1,44 @@
...
+  ...base,
+  mode: 'development',
+  devServer: {
+    contentBase: path.join(__dirname, 'public'),
+    historyApiFallback: true,
+    publicPath: '/dist/',
+    headers: {
+      'Access-Control-Allow-Origin': '*',
+      'Access-Control-Allow-Methods': '*',
+      'Access-Control-Allow-Headers': '*',
```
jkaster, line 40: I wonder if we can make these more restrictive. @dthorpe @bryans99 thoughts?

bryans99:
Built the extension-api-explorer by running yarn package in the extension-api-explorer package. Output was put into the api-explorer dist directory. 
Grabbed that out put and added it to my marketplace install and did an install. Unfortunately resulted in the following
image

packages/api-explorer/package.json
```diff
@@ -108,8 +70,7 @@
     "react-markdown": "^4.3.1",
     "react-router": "^5.1.2",
     "react-router-dom": "^5.1.2",
-    "styled-components": "^4.4.1",
-    "ts-jest": "^26.0.0"
+    "styled-components": "^4.4.1"
```
whscullin, line 73: As long as you're cleaning these up you should remove the duplicate types here, and probably move ace into dependendencies.

PR #262, opened 2020-07-14
Make "rate_limited" a search term

jkaster:
- group `rate_limited` into `activityType` search criterion
- use `\t` instead of `\n` as the search term delimiter so whitespace wildcards are simpler to express in Javascript regex


jkaster:
Merged master so there's really only 1 commit for this PR although the first two commits seem to be part of this PR. I probably should have rebased instead?

josephaxisa:
> Merged master so there's really only 1 commit for this PR although the first two commits seem to be part of this PR. I probably should have rebased instead?

Yep! FWIW, `git rebase --onto master HEAD~1` would have done the trick. It's ok though, I recognize the two other commits and the diff is only for the one commit anyway.

PR #270, opened 2020-07-22
Lerna ES build watcher

josephaxisa:
Added a "watch" script that runs Babel in watch mode for all packages for a more pleasant development experience. 

Unfortunately I couldn't find a good way to also run `tsc -b tsconfig.build.json` in parallel so type errors presented during development might be false positives. On the other hand this is not very important since types don't matter at runtime, the pre-commit hook will flag and prevent them from being checked in and so does `yarn build`.

Also, naming is hard and I was torn between `watch` and `build:watch` for this script. Let me know if you have a preference.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,8 @@
     "test:ts": "yarn jest --reporters=default --reporters=jest-junit",
     "test:iphone": "xcodebuild test -project swift/looker/looker.xcodeproj -scheme looker-Package -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 11,OS=13.4.1' | xcpretty --test --color",
     "test:jest": "DOT_ENV_FILE=.env.test jest",
-    "bootstrap": "lerna clean -y && lerna bootstrap"
+    "bootstrap": "lerna clean -y && lerna bootstrap",
+    "watch": "lerna run --parallel watch"
   },
   "devDependencies": {
     "@babel/cli": "^7.8.4",
--- a/packages/api-explorer/package.json
+++ b/packages/api-explorer/package.json
@@ -17,7 +17,7 @@
   },
   "scripts": {
     "test": "jest",
-    "bundle": "tsc && webpack --config webpack.prod.config.js",
+    "build": "tsc && webpack --config webpack.prod.config.js",
     "develop": "webpack-dev-server --https --disable-host-check --config webpack.dev.config.js",
     "docs": "typedoc --mode file --out docs",
     "register": "ts-node -O '{ \"module\": \"commonjs\", \"target\": \"es2019\" }' scripts/register.ts",
--- a/packages/api-explorer/package.json
+++ b/packages/api-explorer/package.json
@@ -18,6 +18,7 @@
   "scripts": {
     "test": "jest",
     "develop": "webpack-dev-server --https --disable-host-check --config webpack.dev.config.js",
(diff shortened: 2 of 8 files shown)


packages/api-explorer/package.json
```diff
@@ -17,10 +17,11 @@
   },
   "scripts": {
     "test": "jest",
+    "bundle": "tsc && webpack --config webpack.prod.config.js",
```
lukelooker, line 20: I'd probably go with `build` here :)

```suggestion
    "build": "tsc && webpack --config webpack.prod.config.js",
```

package.json
```diff
@@ -44,7 +44,8 @@
     "test:ts": "yarn jest --reporters=default --reporters=jest-junit",
     "test:iphone": "xcodebuild test -project swift/looker/looker.xcodeproj -scheme looker-Package -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 11,OS=13.4.1' | xcpretty --test --color",
     "test:jest": "DOT_ENV_FILE=.env.test jest",
-    "bootstrap": "lerna clean -y && lerna bootstrap"
+    "bootstrap": "lerna clean -y && lerna bootstrap",
+    "watch": "lerna run --parallel watch"
```
lukelooker, line 48: You might be able to remove the watch scripts from all of the sub-projects by doing something like:

```
    "watch": "yarn lerna exec --scope '@looker/*' --stream 'BABEL_ENV=build babel src --root-mode upward --out-dir lib --source-maps --extensions .ts,.tsx --no-comments' --watch",
```

package.json
```diff
@@ -44,7 +44,8 @@
     "test:ts": "yarn jest --reporters=default --reporters=jest-junit",
     "test:iphone": "xcodebuild test -project swift/looker/looker.xcodeproj -scheme looker-Package -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 11,OS=13.4.1' | xcpretty --test --color",
     "test:jest": "DOT_ENV_FILE=.env.test jest",
-    "bootstrap": "lerna clean -y && lerna bootstrap"
+    "bootstrap": "lerna clean -y && lerna bootstrap",
```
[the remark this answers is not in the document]
jkaster: Bootstrap event has some deprecation warnings now, IIRC

package.json
jkaster: That would be handy

josephaxisa: Unfortunately that throws saying that `watch` is an unknown lerna argument

josephaxisa: and if `--watch` is moved to be part of the babel cmd then it is left hanging on the second package build:

```
$ yarn lerna exec --scope '@looker/*' --stream 'BABEL_ENV=build babel src --root-mode upward --out-dir lib --source-maps --extensions .ts,.tsx --no-comments --watch'
lerna notice cli v3.22.1
lerna info versioning independent
lerna notice filter including "@looker/*"
lerna info filter [ '@looker/*' ]
lerna info Executing command in 7 packages: "BABEL_ENV=build babel src --root-mode upward --out-dir lib --source-maps --extensions .ts,.tsx --no-comments --watch"
@looker/sdk-codegen-utils: Successfully compiled 2 files with Babel (545ms).
@looker/sdk: Successfully compiled 34 files with Babel (3907ms).
```

PR #276, opened 2020-07-25
yarn refresh in packages/api-explorer updates to the latest API specs

jkaster:
- run `yarn refresh` in packages/api-explorer
- Cleaned up some codegen script files
- Put APIX script utils in a separate file
- updated all 3 API specs for APIX bundling

If desired, we could pretty-print the JSON output when we write it to the spec folder but typically we wouldn't be looking at the fetched/converted file directly so I didn't bother, particularly when our IDEs can prettify if necessary

### DIFF
--- a/packages/api-explorer/scripts/utils.ts
+++ b/packages/api-explorer/scripts/utils.ts
@@ -90,7 +90,6 @@ export const registerOAuthApp = async (
     app = await sdk.ok(sdk.update_oauth_client_app(guid, appInfo))
     console.log(`Updated ${guid} settings`)
     console.debug({ app })
-    return result
   } catch (e) {
     try {
       const app = await sdk.ok(sdk.register_oauth_client_app(guid, appInfo))
--- a/packages/api-explorer/scripts/utils.ts
+++ b/packages/api-explorer/scripts/utils.ts
@@ -1,0 +1,128 @@
+/*
+
+ MIT License
+
+ Copyright (c) 2020 Looker Data Sciences, Inc.
+
+ Permission is hereby granted, free of charge, to any person obtaining a copy
+ of this software and associated documentation files (the "Software"), to deal
+ in the Software without restriction, including without limitation the rights
+ to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ copies of the Software, and to permit persons to whom the Software is
+ furnished to do so, subject to the following conditions:
+
+ The above copyright notice and this permission notice shall be included in all
+ copies or substantial portions of the Software.
+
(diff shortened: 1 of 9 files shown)


packages/sdk-codegen-scripts/src/legacyGenerator.ts
```diff
@@ -0,0 +1,100 @@
...
+    language,
+    '-o',
+    apiPath,
+    '--enable-post-process-file',
+    options,
+  ])
+}
+
+/**
+ * generate all languages for the specified configuration
```
josephaxisa, line 79:
```suggestion
 * Generate all languages for the specified configuration
```

josephaxisa, line 79:
```suggestion
 * Generate all languages for the specified configuration
```

packages/sdk-codegen-scripts/src/legacyGenerator.ts
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Generate API bindings with the OpenAPI legacy code generator
+ * DEPRECATED: This is using the legacy code generator perform the generation for specific API version, configuration,
+ * and language
+ *
+ * @param {string} fileName specification file name
+ * @param {IGeneratorSpec} spec for language generator options
+ * @param {ISDKConfigProps} props SDK configuration properties
+ * @returns {Promise<Buffer | string>}
```
josephaxisa, line 53: I think IntelliJ is generating old style doc comments for you

packages/api-explorer/scripts/utils.ts
```diff
@@ -0,0 +1,128 @@
...
+  try {
+    console.log(
+      `Checking if "${guid}" is registered as an OAuth application ...`
+    )
+    let app = await sdk.ok(sdk.oauth_client_app(guid))
+    console.log(`${guid} is already registered as ${app.display_name}`)
+    app = await sdk.ok(sdk.update_oauth_client_app(guid, appInfo))
+    console.log(`Updated ${guid} settings`)
+    console.debug({ app })
+    return result
```
josephaxisa, line 93: This return might not be needed due to L103.
```suggestion
```

packages/sdk-codegen-scripts/src/legacyGenerator.ts
jkaster: That's just code I moved from legacy.ts to legacyGenerator.ts. Comment cleanup is a batch task I'm hoping to automate, possibly with a lint fix.

PR #282, opened 2020-07-30
Python SDK safe enum deserialization

joeldodge79:
Prior to this change python SDK would raise when the API returned an
invalid enum value. Given an enum type of `MyAPIEnum` and an invalid
value of `""` this error would be raised

`ValueError: '' is not a valid MyAPIEnum`

Now it will deserialize to `MyAPIEnum.invalid_api_enum_value`

Side note: I really wanted to but was unable to dynamically modify the
`MyAPIEnum.invalid_api_enum_value` enum member value to be the actual
invalid value returned by the API. So
`MyAPIEnum.invalid_api_enum_value.value` will always be
`"invalid_api_enum_value"`

### DIFF
--- a/package.json
+++ b/package.json
@@ -134,7 +134,9 @@
       "@typescript-eslint/no-explicit-any": "off",
       "@typescript-eslint/camelcase": "off",
       "@typescript-eslint/interface-name-prefix": "off",
-      "@typescript-eslint/no-unused-vars": "warn",
+      "@typescript-eslint/no-unused-vars": [
+        "warn", {"args": "all", "argsIgnorePattern": "^_"}
+      ],
       "sort-keys-fix/sort-keys-fix": "off",
       "no-useless-constructor": "off",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -134,7 +134,9 @@
       "@typescript-eslint/no-explicit-any": "off",
       "@typescript-eslint/camelcase": "off",
       "@typescript-eslint/interface-name-prefix": "off",
-      "@typescript-eslint/no-unused-vars": "warn",
+      "@typescript-eslint/no-unused-vars": [
+        "warn", {"args": "all", "argsIgnorePattern": "^_"}
+      ],
```
joeldodge79, line 139: I was seeing `Warning: eslint: '_indent' is defined but never used.` for things like `streamsPrologue(_indent: string)`.

jkaster: yeah we need to tweak eslint settings

josephaxisa: We needed this in APIX too. Thanks for fixing.

PR #284, opened 2020-07-31
Make Swift SDK public or open as needed

jkaster:
Adds `open` or `public` where it should be needed for the Swift SDK to be packageable


bin/smoke
```diff
@@ -40,7 +40,7 @@ if [[ $tests == *kotlin* ]]
 then
   cd kotlin || exit
   echo "Smoking Kotlin SDK ..."
-  gradle test
```
rlm1023, line 43: Random non-Swift file changes above here... needed?

jkaster: Yeah, just a minor smoke test fix for a smoke test requiring less setup for Kotlin

PR #286, opened 2020-08-07
Making server url configuration always go to local storage

jkaster:
- removed the radio group toggle for choosing between session and local storage
- config is always saved to local storage


jkaster:
I'll leave form sizing to the styling work.

bryans99:
Confused as to why request for review after merge into master?

jkaster:
I just commented, but it shouldn't have resubmitted a review. Sorry for the false alarm.

PR #287, opened 2020-08-11
Cleaned up the performance tracker, added post-processing metrics

jkaster:
Perf tracker still needs design work, but it's functional

- added post-request processing metric to the performance tracker
- upgraded yarn packages including Looker components
- fixed all broken tests cause by Looker component upgrade

image

I'm adding all of you just so you're aware of what it has. Don't need reviews from _all_ of you. ;)

### DIFF
--- a/packages/api-explorer/specs/Looker.3.1.oas.json
+++ b/packages/api-explorer/specs/Looker.3.1.oas.json
@@ -2,7 +2,7 @@
   "openapi": "3.0.0",
   "info": {
     "version": "3.1.0",
-    "x-looker-release-version": "7.13.0",
+    "x-looker-release-version": "7.15.0",
     "title": "Looker API 3.1 Reference",
     "description": "### Authorization\n\nThe Looker API uses Looker **API3** credentials for authorization and access control. Looker admins can\ncreate API3 credentials on Looker's **Admin/Users** page. Pass API3 credentials to the **/login** endpoint to\nobtain a temporary access_token. Include that access_token in the Authorization header of Looker API requests.\nFor details, see [Looker API Authorization](https://looker.com/docs/r/api/authorization)\n\n### Client SDKs\n\nThe Looker API is a RESTful system that should be usable by any programming language capable of making\nHTTPS requests. Client SDKs for a variety of programming languages can be generated from the Looker API's Swagger\nJSON metadata to streamline use of the Looker API in your applications. A client SDK for Ruby is available\nas an example. For more information, see [Looker API Client SDKs](https://looker.com/docs/r/api/client_sdks)\n\n### Try It Out!\n\nThe 'api-docs' page served by the Looker instance includes 'Try It Out!' buttons for each API method. After logging\nin with API3 credentials, you can use the \"Try It Out!\" buttons to call the API directly from the documentation\npage to interactively explore API features and responses.\n\nNote! With great power comes great responsibility: The \"Try It Out!\" button makes API calls to your live Looker\ninstance. Be especially careful with destructive API operations such as `delete_user` or similar.\nThere is no \"undo\" for API operations.\n\n### Versioning\n\nFuture releases of Looker will expand this API release-by-release to securely expose more and more of the core\npower of Looker to API client applications. API endpoints marked as \"beta\" may receive breaking changes without\nwarning (but we will try to avoid doing that). Stable (non-beta) API endpoints should not receive breaking\nchanges in future releases.\nFor more information, see [Looker API Versioning](https://looker.com/docs/r/api/versioning)\n\n### In This Release\n\nThe following are a few examples of noteworthy items that have changed between API 3.0 and API 3.1.\nFor more comprehensive coverage of API changes, please see the release notes for your Looker release.\n\n### Examples of new things added in API 3.1 (compared to API 3.0):\n\n* [Dashboard construction](#!/3.1/Dashboard/) APIs\n* [Themes](#!/3.1/Theme/) and [custom color collections](#!/3.1/ColorCollection) APIs\n* Create and run [SQL Runner](#!/3.1/Query/run_sql_query) queries\n* Create and run [merged results](#!/3.1/Query/create_merge_query) queries\n* Create and modify [dashboard filters](#!/3.1/Dashboard/create_dashboard_filter)\n* Create and modify [password requirements](#!/3.1/Auth/password_config)\n\n### Deprecated in API 3.0\n\nThe following functions and properties have been deprecated in API 3.0.  They continue to exist and work in API 3.0\nfor the next several Looker releases but they have not been carried forward to API 3.1:\n\n* Dashboard Prefetch functions\n* User access_filter functions\n* User API 1.0 credentials functions\n* Space.is_root and Space.is_user_root properties. Use Space.is_shared_root and Space.is_users_root instead.\n\n### Semantic changes in API 3.1:\n\n* [all_looks()](#!/3.1/Look/all_looks) no longer includes soft-deleted looks, matching [all_dashboards()](#!/3.1/Dashboard/all_dashboards) behavior.\nYou can find soft-deleted looks using [search_looks()](#!/3.1/Look/search_looks) with the `deleted` param set to True.\n* [all_spaces()](#!/3.1/Space/all_spaces) no longer includes duplicate items\n* [search_users()](#!/3.1/User/search_users) no longer accepts Y,y,1,0,N,n for Boolean params, only \"true\" and \"false\".\n* For greater client and network compatibility, [render_task_results](#!/3.1/RenderTask/render_task_results) now returns\nHTTP status **202 Accepted** instead of HTTP status **102 Processing**\n* [all_running_queries()](#!/3.1/Query/all_running_queries) and [kill_query](#!/3.1/Query/kill_query) functions have moved into the [Query](#!/3.1/Query/) function group.\n\n\nIf you have application code which relies on the old behavior of the APIs above, you may\ncontinue using the API 3.0 functions in this Looker release. We strongly suggest you\nupdate your code to use API 3.1 analogs as soon as possible.\n\n",
     "contact": {
@@ -10,10 +10,7 @@
       "email": "support@looker.com",
       "url": "https://looker.com"
     },
-    "license": {
-      "name": "EULA",
-      "url": "https://self-signed.looker.com:9999/eula"
-    }
+    "license": { "name": "EULA", "url": "http://localhost:9999/eula" }
   },
   "tags": [
     { "name": "ApiAuth", "description": "API Authentication" },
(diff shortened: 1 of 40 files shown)


packages/api-explorer/src/components/DocResponses/DocResponses.spec.tsx
```diff
@@ -42,12 +42,12 @@ describe('DocResponses', () => {
     const responseTree = buildResponseTree(responses)
     const expectedRespStatuses = Object.keys(responseTree)
     expect(
-      screen.getAllByRole('button', {
+      screen.getAllByRole('tab', {
```
jkaster, line 45: this is a good component attribute change

packages/run-it/src/components/PerfTracker/perfTableUtils.tsx
```diff
@@ -41,63 +41,70 @@ export const tableColumns: ActionListColumns = [
     primaryKey: true,
     title: 'URL',
     type: 'string',
-    widthPercent: 18,
+    widthPercent: 15,
```
jkaster, line 44: one more column. gotta use the shrinkolator until we decide which columns to prioritize in the grid, or adjust the display with of the content pane.

packages/run-it/src/components/PerfTracker/perfUtils.ts
```diff
@@ -39,6 +39,9 @@ export interface IResourceLoadTimes {
   fetchUntilResponseEnd: number
   requestUntilResponseEnd: number
   startUntilResponseEnd: number
+  processStart: number
+  processEnd: number
+  processDuration: number
```
jkaster, line 44: Post-processing values for extracting text, binary, parsing JSON, etc

packages/sdk/src/rtl/browserTransport.ts
```diff
@@ -46,22 +46,114 @@ import { BaseTransport } from './baseTransport'
...
+    removePerf(end, 'mark')
+    mark(name, endMark)
+    let start = `${name}-${startMark}`
+    if (performance.getEntriesByName(start, 'mark').length === 0) {
+      start = `${url}-${startMark}`
+    }
+    removePerf(name, 'measure')
+    performance.measure(`${name}`, start, end)
+  }
+}
```
jkaster, line 127: This code gets mildly ambiguous because sometimes the timing resource for the _start_ is already in existence so the name is more precise, and sometimes it's not. Also, depending on whether the end-user uses `rawRequest()` by itself or followed by `request()` we may need to ignore the previous _end_ mark.

bryans99:
Comments on usability
1. Wasn't obvious what each of the lines signify. Perhaps a tooltip with explanatory blurb
2. Not obvious that processDuration is a line (only found it by accident)
3. showAll toggle doesn't seem to do anything
4. showAll toggle state not remembered on leave/return to/from other tabs
5. Refresh icon does not seem to do anything

packages/run-it/src/utils/RunItSDK.ts
```diff
@@ -94,8 +95,14 @@ export class RunItSettings extends ApiSettings {
...
 
+const perfSDK = (settings: Partial<IApiSettings>) => {
+  const sdk = LookerBrowserSDK.init40(new RunItSettings(settings))
+  BrowserTransport.trackPerformance = true
+  return sdk
+}
+
 /** Initialized stand-alone API test runner */
-export const runItSDK = LookerBrowserSDK.init40(new RunItSettings(settings))
+export const runItSDK = perfSDK(settings)
```
jkaster, line 105: Explicitly turn on performance tracking

packages/sdk/src/rtl/browserTransport.ts
```diff
@@ -192,6 +79,106 @@ export class BrowserTransport extends BaseTransport {
...
+
+  private static _trackPerf = false
+
+  public static get trackPerformance() {
+    return this._trackPerf
+  }
+
+  public static set trackPerformance(value: boolean) {
+    this._trackPerf = value && BrowserTransport.supportsPerformance()
+  }
```
jkaster, line 99: Toggle to turn performance tracking on/off explicitly

packages/sdk/src/rtl/browserTransport.ts
```diff
@@ -192,6 +79,106 @@ export class BrowserTransport extends BaseTransport {
...
+  static markName(url: string) {
+    if (!this._trackPerf) return url
+
+    const entries = performance.getEntriesByName(url, 'resource')
+    if (entries.length > 0) {
+      const last = entries[entries.length - 1]
+      return `${url}-${last.startTime}`
+    }
+    return url
+  }
```
jkaster, line 134: This is the crux of accurate performance tracking issues: sometimes the resource timing performance entry for the URL is not complete at the time `markName()` is called, so it defaults to the more ambiguous _url as name_ pattern rather than including the start time used to uniquely identify duplicate URL requests.

Also, in `endMark()` sometimes there is a second resource timer that exists for the same url before `endMark()` is called, which leads to `markName()` finding the second one and losing the `startTime` tag.

This leads to `endMark()` not being able to associate with the correctly produced start mark to create the measure for the correct request.

There is some latency in the performance collections as well for retrieval and deleting.

So, I'm looking for an unambiguous, thread-safe way to track the performance metric. Maybe adding `startMark? :string` to `IRawResponse` and using that in `markEnd()` will work. I don't think https://developer.mozilla.org/en-US/docs/Web/API/Performance/now provides the level of timing precision we need since it rounds to the millisecond. I'll just need a thread-safe `startMark` tag creator.

packages/run-it/src/components/PerfTracker/perfUtils.ts
```diff
@@ -117,10 +139,10 @@ export class LoadTimes implements IResourceLoadTimes {
 export class PerfTimings {
   private _full = false
```
bryans99, line 140: isFull does not appear to be used. Just an observation

packages/run-it/src/components/ShowResponse/responseUtils.tsx
```diff
@@ -59,7 +59,7 @@ const ShowText = (response: IRawResponse) => (
  * @returns Image content
  */
 const imageContent = (response: IRawResponse) => {
-  let content
+  let content = ''
```
bryans99, line 62: Not sure that this change (the else also sets content) was needed but I'm okay with it

packages/sdk/src/rtl/browserTransport.ts
```diff
@@ -100,6 +79,93 @@ export class BrowserTransport extends BaseTransport {
     super(options)
   }
 
+  /** Does this browser have the necessary performance APIs? */
+  static supportsPerformance() {
+    return (
+      performance &&
+      performance.mark !== undefined &&
+      performance.measure !== undefined
+    )
```
bryans99, line 88: performance && !!performance.mark && !!performance.measure

whscullin:
I see yarn.lock got updated, but I don't see updated package versions in package.json

jkaster:
> I see yarn.lock got updated, but I don't see updated package versions in package.json

@whscullin I did `yarn upgrade` to update the packages .. I guess that didn't mod `package.json`. I'll make a pass through it.

packages/sdk/src/rtl/browserTransport.ts
jkaster: ok. Mozilla examples showed them explicitly checked, but maybe that was for javascript?

packages/run-it/src/components/ShowResponse/responseUtils.tsx
jkaster: the linter was complaining so I put that in to shut it up

packages/run-it/src/components/PerfTracker/perfUtils.ts
jkaster: Not currently. Originally I had it hooked up for enabling the `Clear` button in the perf table, but then I decided people probably want to clear whenever they want.

packages/run-it/src/components/PerfTracker/perfUtils.ts
```diff
@@ -97,11 +104,26 @@ export class LoadTimes implements IResourceLoadTimes {
...
         )
+      this.calcProcessTime()
     }
   }
 
-  get duration() {
-    return this.entry.duration
+  private calcProcessTime() {
+    this.duration = this.entry.duration
+    if (performance.getEntriesByName !== undefined) {
```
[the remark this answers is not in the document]
jkaster: According to Mozilla docs, you can't trust there will be a complete implementation of the Performance API in a given browser. As Opera answered for you ;)

packages/run-it/src/components/PerfTracker/perfUtils.ts
```diff
@@ -56,6 +59,7 @@ export const perfRound = (num: number) =>
 const diff = (end: number, start: number) => perfRound(end - start)
 
 export class LoadTimes implements IResourceLoadTimes {
+  duration = 0
```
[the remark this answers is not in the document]
jkaster: I was just adopting the terminology from the performance API and by the time I implemented this I was considering them "load times" so I chose that. If you feel strongly about it I can rename refactor.

packages/run-it/src/components/PerfTracker/perfTableUtils.tsx
```diff
@@ -121,6 +131,19 @@ const metric = (value: number, description: string) => (
   </>
 )
 
+// TODO get this work for trimming left side of url
```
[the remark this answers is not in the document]
jkaster: Thanks for helping me get this bit working.

packages/run-it/src/components/Loading/Loading.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { Flex, Spinner } from '@looker/components'
+
+interface LoadingProps {
+  loading: boolean
+  message?: string
+}
+
+export const Loading: FC<LoadingProps> = ({
+  loading,
+  message = 'Loading ...',
```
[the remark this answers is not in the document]
jkaster: Typically I think it's loading. Would *Working* be a better generic default?

packages/run-it/src/components/PerfTracker/PerfTracker.tsx
```diff
@@ -79,50 +76,54 @@ export const PerfTracker: FC<PerfTrackerProps> = ({
...
-          name="filtering"
-          options={filterOptions}
-          defaultValue={filterOption}
-          onChange={handleFilterChange}
-          inline
-        />
-      </Space>
+      <Heading>Load Times for {filter}</Heading>
+      <Flex>
+        <FlexItem>
```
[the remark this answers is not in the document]
jkaster: I am following the pattern from https://components.looker.com/components/layout/flex/

packages/run-it/src/RunIt.tsx
```diff
@@ -215,14 +219,10 @@ export const RunIt: FC<RunItProps> = ({
...
         <TabPanel key="response">
-          {loading && (
-            <>
-              <Flex>
-                <Spinner />
-                {`${httpMethod} ${pathify(endpoint, activePathParams)}`}
-              </Flex>
-            </>
-          )}
+          <Loading
```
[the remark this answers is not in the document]
jkaster: That component is used in multiple places in APIX. I'm happy to discuss the appropriate generic default message and corresponding component name with you, though.

bryans99:
Going to put some comments here but dont need to get fixed as part of this PR

1. Should https://self-signed.looker.com.* be https://self-signed.looker.com/*? Should we be showing the port number
image

2. The URL for the chart I am looking at does not standout

image

3. Start time does not make much sense (see BR/JK slack convo)
image

4. Show ULS with last run first
image

5. Consider filtering out API token (see above image)

6. Timings for a request should be 0 based (currently seems offset from time performance offset). Something like find the earliest start time and subtract from all start and end times. My theory is that this will give us a consistent tick display at the bottom of the chart.
image

7. Consider creating an issue for google charts to suppress tick display for timeline charts

packages/api-explorer/specs/Looker.3.1.oas.json
```diff
@@ -2,18 +2,15 @@
   "openapi": "3.0.0",
   "info": {
     "version": "3.1.0",
-    "x-looker-release-version": "7.13.0",
+    "x-looker-release-version": "7.15.0",
```
bryans99, line 5: Sorry, I'm gonna assume that all the changes to this particular file are valid. Might want to get another approval on this one.

packages/run-it/src/components/PerfTracker/PerfChart.tsx
```diff
@@ -84,33 +86,36 @@ export const PerfChart: FC<PerfChartProps> = ({ loadTimes }) => {
...
           textAlign="left"
         >
-          <span>Decoded body size: {entry.decodedBodySize}</span>
+          <span>Decoded body: {entry.decodedBodySize}</span>
         </Tooltip>
       </Space>
       <Chart
         width={'100%'}
-        height={'350px'}
+        height={`${8 * 41 + 10}px`}
```
bryans99, line 112: Still scrolling. 418 should do it. Also, a number is accepted.

packages/run-it/src/components/PerfTracker/PerfChart.tsx
```diff
@@ -84,33 +86,36 @@ export const PerfChart: FC<PerfChartProps> = ({ loadTimes }) => {
...
         chartType="Timeline"
-        loader={<div>Loading Chart</div>}
+        loader={<Loading loading={true} />}
         data={data}
+        options={{
```
bryans99, line 116: Upon further investigation, this has no effect. Remove

packages/run-it/src/components/PerfTracker/PerfTracker.tsx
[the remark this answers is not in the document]
bryans99: Fine by me. I mention because Scott has issues with FlexItem :D

packages/api-explorer/specs/Looker.3.1.oas.json
jkaster: yeah, I refreshed the spec files. These won't be in the final version at all because specs will be fetched directly from the Looker server. We've got some Looker server changes to land before that's possible, though.

packages/run-it/src/components/PerfTracker/PerfChart.tsx
jkaster: Sorry, I don't understand what _418 should do it_ means

packages/run-it/src/components/PerfTracker/PerfChart.tsx
jkaster: removed. Weird how it sometimes shows up and other times doesn't.

PR #289, opened 2020-08-12
Python SDK: enable models as dictionaries

joeldodge79:
The python SDK now accepts dictionaries as method bodies in addition to
the original model class instances. E.g. these are equivalent:

```
sdk.create_user({"first_name": "Jane", "last_name": "Smith"})
sdk.create_user(models.Writeuser(first_name="Jane", last_name="Smith"))
```

Note: static type checking fails on the first example but the assumption
is that users who opt for that form are not interested in type safety

Additionally, model class instances returned by the API act like
dictionaries. So you can do this:

```
me = sdk.me()
me["locale"] = "uk"
sdk.update_user(me["id"], me)
```

Note: the SDK will not do anything to prevent sending read-only
properties - the caller is responsible for the API's behavior in such
cases.

The dictionary-like behavior of model class instances is nearly
complete. `popitem()`, `clear()`, and `copy()` are the only methods not
supported. Another difference is setting a nested child property
to a dictionary: under the hood that gets converted to
the appropriate model instance and no reference to the original
dictionary is kept. The following demonstrates this:

```
my_email = {"email": "jane.smith@example.com"}
me = sdk.me()
me["credentials_api3"] = my_email
my_email["email"] = "jane.smith@foo.com"
print(me["credentials_api3"]["email"])  # jane.smith@example.com
me["credentials_api3"]["email"] = "jane.smith@foo.com"
print(me["credentials_api3"]["email"])  # jane.smith@example.com
```
(unrealistic example as the Looker API does not support updating email
 address via update_user)

### DIFF
--- a/python/looker_sdk/rtl/model.py
+++ b/python/looker_sdk/rtl/model.py
@@ -24,7 +24,16 @@
 """
 
 import collections
+import enum
+import keyword
 from typing import Any, cast, Iterable, Sequence, Optional, TypeVar
+
+try:
+    from typing import ForwardRef  # type: ignore
+except ImportError:
+    from typing import _ForwardRef as ForwardRef  # type: ignore
+
+import cattr
 
 
 EXPLICIT_NULL = cast(Any, "EXPLICIT_NULL")  # type:ignore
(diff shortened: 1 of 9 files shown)


python/looker_sdk/rtl/model.py
```diff
@@ -34,6 +43,124 @@ class Model:
...
+                f"'{self.__class__.__name__}' object has no attribute '{key}'"
+            )
+        annotation = self.__annotations__[key]
+        if isinstance(annotation, ForwardRef):
+            actual_type = eval(
+                annotation.__forward_arg__, self.__global_context, locals()
+            )
+            if isinstance(actual_type, enum.EnumMeta):
+
+                # untyped because mypy really doesn't like this enum internals stuff
```
jkaster, line 81: who does?

python/looker_sdk/rtl/model.py
```diff
@@ -34,6 +43,124 @@ class Model:
...
+                def err(val):
+                    valid = []
+                    for v in actual_type.__members__.values():
```
jkaster, line 84: could this be a class static collection so it's not looped every time?

python/looker_sdk/rtl/model.py
```diff
@@ -34,6 +43,124 @@ class Model:
...
+        return key in cattr.unstructure(self)
+
+    def keys(self):
+        return cattr.unstructure(self).keys()
+
+    def items(self):
+        return cattr.unstructure(self).items()
+
+    def values(self):
+        return cattr.unstructure(self).values()
```
jkaster, line 124: can/should `cattr.unstructure(self)` be evaluated once, or must it be evaluated every time?

python/looker_sdk/sdk/api40/methods.py
```diff
@@ -8442,8 +8442,6 @@ def create_user(
         transport_options: Optional[transport.PTransportSettings] = None,
     ) -> models.User:
         """Create User"""
-        if body:
-            assert isinstance(body, models.WriteUser)
```
jkaster, line 8446: I expected a bigger delta than this, for sure. Didn't you remove these assertions from all methods? Or was this the only one remaining that needed to be removed? Or was this added by hand then removed when you determined it wasn't needed for generated code?

python/tests/integration/api40/test_methods.py
```diff
@@ -21,11 +21,16 @@ def test_crud_user(sdk: mtds.Looker40SDK):
...
+            first_name="Rudolphontronix",
+            last_name="Doe",
+            is_disabled=False,
+            locale="fr",
+        )
     )
     assert isinstance(user, ml.User)
     assert isinstance(user.id, int)
-    assert user.first_name == "John"
+    assert user.first_name == "Rudolphontronix"
```
jkaster, line 33: Now there's a great first name.

python/tests/integration/api40/test_methods.py
```diff
@@ -73,6 +78,62 @@ def test_crud_user(sdk: mtds.Looker40SDK):
...
+    # Create user
+    new_user = sdk.create_user(
+        dict(
+            first_name="Rudolphontronix",
+            last_name="Doe",
+            is_disabled=False,
+            locale="fr",
+        )
+    )
+    assert new_user["first_name"] == "Rudolphontronix"
```
jkaster, line 93: ok, ok. really .. can't these values be variables defined once? (`test_first_name` and `test_last_name` perhaps?)

python/tests/rtl/test_serialize.py
```diff
@@ -230,7 +236,263 @@ def __init__(self, *, name2: str):
...
+    assert str(exc.value) == "'Model' object has no attribute 'foobar'"
+
+
+def test_dict_setitem_child(bm):
+    bm["model_no_refs1"]["name1"] = "model_no_refs1_another_name"
+    assert bm["model_no_refs1"]["name1"] == "model_no_refs1_another_name"
+
+    # creating new children from dictionaries instantiates a new child model
+    # object from that dict but does not maintain a reference.
+    child_dict = {"name1": "I used a dictionary"}
```
jkaster, line 309: Did you discover any cool new words in the dictionary?

python/looker_sdk/sdk/api40/methods.py
joeldodge79: oops,  my bad - I need to update codegen to remove them all, good catch!

python/looker_sdk/rtl/model.py
joeldodge79: python enums don't behave like regular classes so you can't really just add a static/class property - it just defines another enum member:

```
>>> class e(enum.Enum):
...     a = "a"
...     b = "b"
...     _values = ["a", "b"]

>>> e.a
<e.a: 'a'>

>>> e._values
<e._values: ['a', 'b']>

>>> e._values.value
['a', 'b']
```

we could do that (and access `actual_type._values.value`) but that sort of pollutes the enum... but we're already polluting it with `invalid_api_enum_value` so maybe that's ok?

If your concern is performance: the code inside this nested function is compiled but doesn't run unless called which only happens if the user tries to set in an invalid enum value. Also, our enums don't tend to be too big... I guess I favor this looping over polluting the codegen'd enums

python/looker_sdk/rtl/model.py
joeldodge79: I struggled with this a little bit - my mind went right to "how inefficient!" - My next thought was

> oh man, trying to cache this stuff and keep it in sync as they manipulate the "dictionary" is going to be really tedious and tricky

So I decided to opt for easy correctness for now: calling these methods will dump the data at that moment so it will be correct. I figure if people notice/complain about performance hits using dictionary mode then we can start to profile and figure it out. But probably not worth the extra investment without getting feedback first that people even use this stuff.

python/tests/integration/api40/test_methods.py
joeldodge79: fair enough, it's just copy/pasta from the `test_crud_user` test but I can remove all these magic strings

python/tests/rtl/test_serialize.py
joeldodge79: Last night I looked up the word "especial" on my kindle - didn't realize that was a word in English (just thought it was Spanish)

python/tests/integration/api40/test_methods.py
jkaster: Thanks!

python/looker_sdk/rtl/model.py
jkaster: Good plan

python/looker_sdk/rtl/model.py
jkaster: I think you made the right call

PR #291, opened 2020-08-13
Fixes to support embedding in other projects

kalenp:
This defines a project group and name which are necesssary so that the
kotlin sdk may be accessed as a project.  Until we have the project
published to a package distribution system, this allows us to use a
Gradle Source Dependency [1] to embed the project.

Also changes a few `compile` dependencies to their newer versions to
remove deprecation warnings from gradle.

[1] https://blog.gradle.org/introducing-source-dependencies

### DIFF
--- a/kotlin/build.gradle
+++ b/kotlin/build.gradle
@@ -10,6 +10,8 @@ buildscript {
     }
 }
 
+group = "com.looker.sdk"
+
 //plugins {
 //    id "org.jmailen.kotlinter" version "2.4.1"
 //}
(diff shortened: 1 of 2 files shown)


kotlin/build.gradle
```diff
@@ -25,14 +27,15 @@ repositories {
...
 sourceSets {
     main.kotlin.srcDirs += 'src/main/'
     test.kotlin.srcDirs += 'src/test'
 }
 
 dependencies {
-    compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
-    compile "org.ini4j:ini4j:0.5.4"
+    api "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+    implementation "org.ini4j:ini4j:0.5.4"
```
jkaster, line 38: Thanks for the updates!

kotlin/settings.gradle
```diff
@@ -0,0 +1 @@
+rootProject.name = 'looker-kotlin-sdk'
```
jkaster, line 1: Good to know. This should help streamline the Kotlin SDK package creation. Funny it's in `settings.gradle` rather than `build.gradle` but I guess `settings` is used for processing `build`?

kalenp: I don't know why they're in separate locations.  I tried putting them both in settings and gradle hated that and I found other examples suggesting this is the way to do it, so here we are.

PR #292, opened 2020-08-13
Change kotlin ApiSettings to be an interface

kalenp:
This allows for completely independent implementations which do not
expect any Ini file data at all.

### DIFF
--- a/kotlin/README.md
+++ b/kotlin/README.md
@@ -54,10 +54,14 @@ When the SDK is installed and the server location and API credentials are config
 Verify authentication works and that API calls will succeed with code similar to the following:
 
 ```kotlin
+import com.looker.rtl.ApiSettingsIniFile;
+import com.looker.rtl.AuthSession;
+import com.looker.sdk.LookerSDK;
+
 val localIni = "./looker.ini"
 val settings = ApiSettingsIniFile(localIni, "Looker")
-val session = UserSession(settings, Transport(settings))
-val sdk = Looker40SDK(session)
+val session = AuthSession(settings)
+val sdk = LookerSDK(session)
 // Verify minimal SDK call works
 val me = sdk.ok<User>(sdk.me())
 
--- a/kotlin/src/main/com/looker/rtl/ApiSettings.kt
+++ b/kotlin/src/main/com/looker/rtl/ApiSettings.kt
@@ -41,8 +41,10 @@ fun apiConfig(contents: String): ApiSections {
     return ret
 }
 
+interface ApiSettings : ConfigurationProvider {}
+
 open class ApiSettingsIniFile(var filename: String = "./looker.ini",
-                              section: String = "") : ApiSettings(File(filename).readText(), section) {
+                              section: String = "") : ApiSettingsIniText(File(filename).readText(), section) {
     override fun readConfig(): Map<String, String> {
         val file = File(filename)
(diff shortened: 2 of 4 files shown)


kotlin/src/main/com/looker/rtl/ApiSettings.kt
```diff
@@ -55,7 +57,7 @@ open class ApiSettingsIniFile(var filename: String = "./looker.ini",
 
 
 // TODO why no @JvmOverloads here?
-open class ApiSettings(contents: String, var section: String = "") : ConfigurationProvider {
+open class ApiSettingsIniText(contents: String, var section: String = "") : ApiSettings, ConfigurationProvider {
```
kalenp, line 60: I wanted to combine this into the ApiSettingsIniFile, perhaps as a named static constructor (companion object method in kotlin), but they both had `readConfig()` implementations, so I didn't want to rock the boat too much by reconciling those.

jkaster: the idea behind `readConfig()` is that it's supposed to support overriding from the base so you can get your configuration any way you want. As long as you're not breaking that "inherit and override" capability, changes should be fine.

kotlin/README.md
```diff
@@ -54,10 +54,14 @@ When the SDK is installed and the server location and API credentials are config
...
+import com.looker.rtl.ApiSettingsIniFile;
+import com.looker.rtl.AuthSession;
+import com.looker.sdk.LookerSDK;
+
 val localIni = "./looker.ini"
 val settings = ApiSettingsIniFile(localIni, "Looker")
-val session = UserSession(settings, Transport(settings))
-val sdk = Looker40SDK(session)
+val session = AuthSession(settings)
+val sdk = LookerSDK(session)
```
jkaster, line 64: wow. readme updates, even.

kotlin/src/main/com/looker/rtl/ApiSettings.kt
jkaster: My intent with `ApiSettings` was for it to be the base overrideable "config reader" implementation that happened to default to `.ini` text, and other inherited overrides would be more explicitly named. I'm wondering how many people would see `ApiSettingsIniText` or `ApiSettingsIniFile` and think "Oh, that's not the one I want to descend from."

kotlin/src/main/com/looker/rtl/Transport.kt
```diff
@@ -111,11 +111,11 @@ typealias Authenticator = (init: RequestSettings) -> RequestSettings
...
-    var baseUrl: String
-    var apiVersion: String
-    var verifySSL: Boolean
-    var timeout: Int
-    var headers: Map<String, String>
+    val baseUrl: String
+    val apiVersion: String
+    val verifySSL: Boolean
+    val timeout: Int
+    val headers: Map<String, String>
```
jkaster, line 118: Does making these immutable, does the transport's request still allow for option overrides to be passed in and used?

kalenp:
I realize I have a few more changes I need to make before this is ready.  I'm going to close this PR for now, and I'll reopen later.

kalenp:
Tried reopening this, but I can't seem to do that.  Maybe because I force-pushed after closing it.  Opened a new PR instead.  https://github.com/looker-open-source/sdk-codegen/pull/293

PR #293, opened 2020-08-14
Change ApiSettings to accept a Map instead of Ini text

kalenp:
This should allow for simpler overriding of behavior, allowing clients
to provide a custom dict rather than requiring that they use the INI
format.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+[*.kt]
+indent_style = space
+indent_size = 4
--- a/kotlin/src/main/com/looker/rtl/ApiSettings.kt
+++ b/kotlin/src/main/com/looker/rtl/ApiSettings.kt
@@ -41,21 +41,43 @@ fun apiConfig(contents: String): ApiSections {
     return ret
 }
 
-open class ApiSettingsIniFile(var filename: String = "./looker.ini",
-                              section: String = "") : ApiSettings(File(filename).readText(), section) {
-    override fun readConfig(): Map<String, String> {
-        val file = File(filename)
-        if (!file.exists()) return mapOf()
-        val contents = File(filename).readText()
-        val config = apiConfig(contents)
-        section = if (!section.isBlank()) section else config.keys.first()
-        return config[section] ?: mapOf()
-    }
-}
-
-
 // TODO why no @JvmOverloads here?
-open class ApiSettings(contents: String, var section: String = "") : ConfigurationProvider {
+// This takes a function returning a map so that the fromIniFile constructor
+// may lazily load the file data and not store the full, parsed file map in
+// memory long term.
+open class ApiSettings(val rawReadConfig: () -> Map<String, String>) : ConfigurationProvider {
+
+    companion object {
+        fun fromIniFile(filename: String = "./looker.ini", section: String = "") : ConfigurationProvider {
+            return ApiSettings({
+                val file = File(filename)
+                if (!file.exists()) {
+                    mapOf()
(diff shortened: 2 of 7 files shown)


kalenp:
@jkaster Could you re-review?  The prior version had broken the behavior where the file processing didn't hold the whole file data in memory long term.  This restores that by using a function to read the data in the base class.

.editorconfig
```diff
@@ -0,0 +1,7 @@
+[*]
+end_of_line = lf
```
jkaster, line 2: What OS are you using where this was causing an issue? Just curious. Or is this a Vim setting?

kotlin/src/main/com/looker/rtl/ApiSettings.kt
```diff
@@ -64,31 +86,22 @@ open class ApiSettings(contents: String, var section: String = "") : Configurati
...
-
-            settings["verify_ssl"].let { value ->
-                verifySSL = asBoolean(value) ?: verifySSL
-            }
-            settings["timeout"].let { value ->
-                timeout = if (value !== null) value.toInt() else timeout
-            }
+        settings["api_version"].let { value ->
+            apiVersion = unQuote(value ?: apiVersion)
+        }
```
jkaster, line 98: I think we don't need this any more at all, but that's not necessary to fix in this PR

kotlin/src/test/Utils.kt
```diff
@@ -34,10 +34,14 @@ typealias jsonDict = Map<String, Any>
...
-    : ApiSettingsIniFile(filename, section) {
+class TestSettingsIniFile(
+    filename: String = "./looker.ini",
+    section: String = "",
+    private val base: ConfigurationProvider = ApiSettings.fromIniFile(filename, section)
+) : ConfigurationProvider by base {
+
     override fun readConfig(): Map<String, String> {
-        var map = super.readConfig().toMutableMap()
+        var map = base.readConfig().toMutableMap()
```
jkaster, line 44: does `base` go to the deepest ancestor rather than whatever's closest in the inheritance tree?

.editorconfig
kalenp: I didn't hit it, but as an OSS project, good to protect against any potential future windows users.

kotlin/src/test/Utils.kt
kalenp: It actually refers to `private val base: ConfigurationProvider` in the constructor.  I'm using kotlin's interface delegation here, since I want to "inherit" from ApiSettings, but I don't want to directly call its constructor.  Sticking that in the TestSettingsIniFile constructor feels like a bit of a hack, but I needed to be able to reference the delegated instance from somewhere.

https://kotlinlang.org/docs/reference/delegation.html

.editorconfig
kalenp: The indent style and indent size settings I totally did use, though.  Apparently I need to get full kotlin support going in my vim instance.  This is just a shortcut for the most common settings.

kotlin/src/test/Utils.kt
jkaster: ah, that's where it is! I read that same link but didn't see a ref to `base`. I missed the variable assignment. Sorry about that.

kalenp: No problem.  It's a bit too subtle for my taste, but I couldn't think of a better way to get what I wanted here.  In the Mock settings, we don't need to access anything from the base instance, and that came out a bit cleaner IMO: https://github.com/looker-open-source/sdk-codegen/pull/293/files#diff-40b1d00a327afcf5a64170b78a2de656R21

PR #296, opened 2020-08-19
Base64 OAuth for Kotlin and Swift SDKs

jkaster:
This also includes making the rest of the Swift SDK RTL open and/or public for access outside of the forthcoming package


jkaster:
I also added the change for Python. I couldn't find a test for it, but I think it only needed to change in one place

PR #297, opened 2020-08-20
Create public constructors for Swift SDK models

jkaster:
Also generated against Looker 7.14+


packages/sdk-codegen/src/swift.gen.spec.ts
```diff
@@ -87,6 +87,42 @@ public enum PermissionType: String, Codable {
...
+    /**
+     * When true, the embed session will purge any residual Looker login state (such as in browser cookies) before creating a new login state with the given embed user info. Defaults to true.
+     */
+    public var force_logout_login: Bool?
+
+    public init(_ target_url: URI, session_length: Int64? = nil, force_logout_login: Bool? = nil) {
+        self.target_url = target_url
+        self.session_length = session_length
+        self.force_logout_login = force_logout_login
+    }
```
jkaster, line 124: Note that required parameters are positional rather than named. If this is not desirable, we can update so all parameters must be named.

swift/looker/Tests/lookerTests/methodsTests.swift
```diff
@@ -26,16 +26,16 @@ class methodsTests: XCTestCase {
     
     func simpleQuery() -> WriteQuery {
         return WriteQuery(
-            model: "system__activity",
-            view: "dashboard",
+            "system__activity",
+            "dashboard",
             fields: ["dashboard.id", "dashboard.title", "dashboard.count"],
             limit: "100")
```
jkaster, line 32: this is how the constructor call changes

rlm1023: Hmm, so it changes the call sites? Does this change just expose the constructors outside of the SDK package? If so, then I guess the change would be isolated to just sdk-codegen, so maybe that's okay.

jkaster: Both patterns are now supported. win/win.

PR #299, opened 2020-08-24
HelloWorld extension

josephaxisa:
TODO:
- [ ] Get the jest config to work for this package.
- [ ] Update readme


bryans99:
@jkaster 

No issues with the code and how it works within the sdk codegen repo. All great.

Missing stuff, README and sample manifest but there is bigger problem to be solved first.

The issue is how is a developer going to create an extension using it. The first thing I did was to copy the directory outside of the mono repo (if I was an extension developer, I don't need the rest of the mono repo, I just want to create my own extension which will go in my own repo). Unfortunately as soon as I did that,  neither yarn nor npm worked (understandably). 

I'm thinking that we should have some kind of script that basically "ejects" (stealing from react here) the package into its own directory external to the sdk codegen repo. I would not call it eject as eject has a different meaning in the react world, perhaps 'exile', 'uproot' or 'separate'. Basically they would run the script, it would prompt for a target directory and template and do its thing. I'm open to other suggestions as well. If we can get the template to a state where I can copy it and it works I'm okay with that too. 

For our development purposes I really like this format, especially when it comes to the kitchen sink. I just think we need to make a third parties life easier as well.

packages/extension-template-helloworld/webpack.dev.config.js
```diff
@@ -0,0 +1,42 @@
...
+  ...base,
+  devServer: {
+    historyApiFallback: true,
+    publicPath: 'http://localhost:8080/',
+    headers: {
+      'Access-Control-Allow-Origin': '*',
+      'Access-Control-Allow-Methods': 'GET, POST, PUT, DELETE, PATCH, OPTIONS',
+      'Access-Control-Allow-Headers':
+        'X-Requested-With, content-type, Authorization',
+    },
```
jkaster, line 39: Are these all required for an extension, or are some of these from the OAuth/CORS flow?

jkaster:
> I'm thinking that we should have some kind of script that basically "ejects" (stealing from react here) the package into its own directory external to the sdk codegen repo. I would not call it eject as eject has a different meaning in the react world, perhaps 'exile', 'uproot' or 'separate'. Basically they would run the script, it would prompt for a target directory and template and do its thing. I'm open to other suggestions as well. If we can get the template to a state where I can copy it and it works I'm okay with that too.

Good point, @bryans99. I'm wondering if we could create a base template to apply to `extract`-ed packages, similar to the way we've got linting, prettier, and babel rules that are shared across front-end repositories. Maybe the `extract` script could create or update an existing stand-alone repository that applies the common base templates to the extracted package.

josephaxisa:
A better home has been found for this.

PR #311, opened 2020-09-01
APIX Styling Round

miketorosian:
I was getting some setup issues re: Version # and also some Prettier issues so some of this may be a bit cluttered. Overall, I took a pass at setting some sensible max-widths on content areas, tweaked the header so that search aligns with the left edge of the main column but remains fairly flexible after adjust, adjusted type size color and the spacing between sections on the Method pages.


packages/api-explorer/src/components/DocReferences/DocReferences.tsx
```diff
@@ -54,8 +55,8 @@ export const DocReferences: FC<DocReferencesProps> = ({
   if (seeTypes.length === 0 && seeMethods.length === 0) return <></>
 
   return (
-    <CollapserCard heading="References">
-      <>
+    <Box mb="xlarge">
+      <CollapserCard heading="References">
```
josephaxisa, line 59: `CollapserCard` expects `children` to be a `ReactElement` so we still need the fragment around the elements here.

packages/api-explorer/src/components/DocSDKs/DocSDKs.tsx
```diff
@@ -50,8 +57,8 @@ export const DocSDKs: FC<LanguageSDKProps> = ({ api, method, type }) => {
   const [item] = useState(method ? noComment(method) : type!)
 
   return (
-    <CollapserCard heading="Language SDK declarations">
-      <>
```
josephaxisa, line 54: Same as above.

packages/api-explorer/src/components/DocResponses/DocResponses.tsx
```diff
@@ -47,8 +54,8 @@ export const DocResponses: FC<DocResponsesProps> = ({ responses }) => {
   const responseTree = buildResponseTree(responses)
 
   return (
-    <CollapserCard heading="Response Models">
-      <>
```
josephaxisa, line 51: Same as above.

packages/api-explorer/src/components/Header/ApiSpecSelector.tsx
```diff
@@ -55,15 +54,6 @@ export const ApiSpecSelector: FC<ApiSpecSelectorProps> = ({
...
-      defaultValue={spec.key}
-      options={options}
-      aria-label="Version"
-      label="Version"
-      mb="0"
-      mr="large"
-      onChange={handleChange}
-      inline
-    />
+    <Select defaultValue={spec.key} options={options} onChange={handleChange} />
```
josephaxisa, line 57: Screen Shot 2020-09-01 at 11 51 32

Was omitting the `Version` label intentional here? Unfortunately it breaks the unit test. `FieldSelect` was used because it has an associated label prop but if there's a preference for `Select` we can create a separate label here.

josephaxisa (changes requested):
Nice progress. Left one minor comment and a question below. I'll push a PR addressing the comment and the `@looker/sdk` version issue shortly.

packages/api-explorer/src/components/Header/ApiSpecSelector.tsx
miketorosian: @josephaxisa - It was, the FieldSelect that was there before was causing odd spacing issues in the header due to the inline attribute that places the label before the select. What if we just put Version before 4.0 in the Select? So in the example above, it would read Version 4.0 (experimental)

jkaster (commented):
I just have one question to be answered.

It's really useful to see some of the more advanced styling patterns implemented here. Thanks for that!

jkaster:
obsolete branch

PR #326, opened 2020-09-17
API Explorer: Fix stale state render bug

josephaxisa:
**The problem**: The default selected media type in `DocResponseTypes` was previously being set once, as a default state value. On navigation to another method, the component received a `responses` object but the default `mediaType` state did not change. As a result, there were cases where the new `responses` object did not have a key matching the default media type previously set on first render causing the app to fail to render the whole scene.

A bug of this nature is also manifesting itself in other components, most notably in `DocSDKs` where the content rendered stays the same as one switches from one method (or type) to another.

**Solution**: In components where there's a link between props and state, there should be a `useEffect` which listens to props changes and updates state accordingly.


jkaster (approved):
LGTM. Nice job tracking down the offending pattern

PR #332, opened 2020-09-24
Swift SDK: Added setToken top-level function for AuthSession

jkaster:
- See testMe() for sample working code
- Made Constants for various version variables public


rlm1023 (approved):
LGTM

PR #339, opened 2020-10-01
Changes to publish latest Typescript SDK

jkaster:
Had to skip a C# generator test for now so I could publish.

Also had to explicitly add the publish scope of `public` for every package to be published. I guess that's a new requirement.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "build:es": "yarn lerna exec --stream 'BABEL_ENV=build babel src --root-mode upward --out-dir lib --source-maps --extensions .ts,.tsx --no-comments'",
     "build:ts": "yarn lerna exec --stream --sort 'tsc -b tsconfig.build.json'",
     "clean": "rm -Rf packages/*/lib",
-    "prepublishOnly": "yarn test",
+    "prepublishOnly": "yarn jest packages/sdk",
     "deploy": "yarn workspace sdk deploy",
     "deploy:codegen": "lerna run deploy --scope sdk-codegen -c",
     "deploy:sdk": "lerna run deploy --scope sdk",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -33,7 +33,7 @@
     "build:es": "yarn lerna exec --stream 'BABEL_ENV=build babel src --root-mode upward --out-dir lib --source-maps --extensions .ts,.tsx --no-comments'",
     "build:ts": "yarn lerna exec --stream --sort 'tsc -b tsconfig.build.json'",
     "clean": "rm -Rf packages/*/lib",
-    "prepublishOnly": "yarn test",
+    "prepublishOnly": "yarn jest packages/sdk/",
```
josephaxisa, line 36: How come we are limiting the prepublish tests to just the sdk?

packages/extension-api-explorer/package.json
```diff
@@ -1,9 +1,10 @@
 {
   "name": "@looker/extension-api-explorer",
-  "version": "1.0.0",
+  "version": "1.0.4-alpha.0",
```
josephaxisa, line 3: What do you think about changing the extension version to match that of `@looker/api-explorer`?

josephaxisa (commented):
I have two minor questions

package.json
jkaster: Because everything else has to work correctly for the TS SDK to pass these tests and be deployable. We can extend it to all sdk packages by using `/sdk` instead of `/sdk/` due to pattern matching but I don't think we need to go that wide yet. All our other packages aren't currently being published. When api-explorer is ready to be published, then we can widen the scope.

packages/extension-api-explorer/package.json
jkaster: It would be attractive to version all the packages via lerna, or get rid of `-alpha`. Bigger issue than just this PR.

package.json
josephaxisa: Makes sense

josephaxisa (approved):

PR #340, opened 2020-10-01
Hackathon app bulk import update for hack@home

joeldodge79:


joeldodge79:
@izzymiller you should be able to try this script out by
1/ clone this repo
2/ make sure you have python 3.8.2 installed (using pyenv, ping me if you're not setup that way)
3/ `cd examples/python/hackathon_app/`
4/ `pipenv install --dev`
5/ follow the "Bulk import script" section of the readme

@jkaster was thinking you might want us to store "t-shirt-size" as a looker user attribute as well. Let me know if that's the case and I can add it here.

examples/python/hackathon_app/README.md
```diff
@@ -87,6 +87,27 @@ yarn start
 
 - this launches the frontend static server at http://127.0.0.1:3000 which proxies fetch requests to the backend server
 
+## Bulk import script
+
+The `bulk_import.py` script reads a CSV file that has "first_name", "last_name", and "email" column headers and adds users to the hack instance.
+It can be re-run safely w/out duplicating users (primary key is "email"). It will send the user an account setup email.
```
jkaster, line 93:
```suggestion
It can be re-run safely w/out duplicating users (primary key is "email"). It will send newly registered users an account setup email.
```

jkaster (changes requested):
Just a quick clarification in the doc. Looks good to me!

examples/python/hackathon_app/bulk_import.py
```diff
@@ -35,6 +33,7 @@ def main(filename: str, hackathon: str, limit: int):
         count += 1
         if limit and count == int(limit):
             break
+        time.sleep(60)
```
jkaster, line 36: 😢

jkaster (approved):
Looks good. Thanks for the quick rate limiting tweak

josephaxisa (approved):
A bit late to the party but this looks good to me!

PR #344, opened 2020-10-06
Route Param Validation

josephaxisa:
Added a `validateAndRender` method for validating route params before rendering a scene.

### DIFF
--- a/packages/api-explorer/src/components/DocMarkdown/DocMarkdown.tsx
+++ b/packages/api-explorer/src/components/DocMarkdown/DocMarkdown.tsx
@@ -56,7 +56,7 @@ export const DocMarkdown: FC<DocMarkdownProps> = ({ source, specKey }) => {
   const handleClick = (e: BaseSyntheticEvent) => {
     if (
       e.target.tagName === 'A' &&
-      e.target.origin === window.location.origin
+      e.target.pathname.startsWith(`/${specKey}`)
     ) {
       e.preventDefault()
       history.push(e.target.pathname)
--- a/packages/api-explorer/src/components/NotFound/NotFound.tsx
+++ b/packages/api-explorer/src/components/NotFound/NotFound.tsx
@@ -23,20 +23,18 @@
  SOFTWARE.
 
  */
-export { ApixHeading } from './common'
-export { DocActivityType } from './DocActivityType'
-export { DocCode } from './DocCode'
-export { DocMethodSummary } from './DocMethodSummary'
-export { DocMarkdown } from './DocMarkdown'
-export { DocPseudo } from './DocPseudo'
-export { DocRateLimited } from './DocRateLimited'
-export { DocReferences } from './DocReferences'
-export { DocResponses } from './DocResponses'
-export { DocSDKs } from './DocSDKs'
-export { DocSdkUsage } from './DocSdkUsage'
-export { DocStatus } from './DocStatus'
-export { DocTitle } from './DocTitle'
-export { GlobalStyles } from './GlobalStyles'
-export { Header } from './Header'
-export { NotFound } from './NotFound'
-export { SideNav } from './SideNav'
+import React, { FC } from 'react'
+import { Space } from '@looker/components'
+import { ApixHeading } from '../common'
+
+interface NotFoundProps {
+  message?: string
+}
+
+export const NotFound: FC<NotFoundProps> = ({ message }) => (
+  <Space p="xxxlarge" width="100%" height="50vh" around>
+    <ApixHeading as="h1">
+      {message || 'The page you requested could not be found.'}
+    </ApixHeading>
+  </Space>
+)
--- a/packages/api-explorer/src/routes/AppRouter.tsx
+++ b/packages/api-explorer/src/routes/AppRouter.tsx
@@ -27,9 +27,11 @@ import React, { FC, useContext } from 'react'
 import { Redirect, Route, Switch } from 'react-router-dom'
 import { ApiModel } from '@looker/sdk-codegen'
 import { OAuthScene, RunItContext } from '@looker/run-it'
-
 import { Looker40SDK } from '@looker/sdk/lib/browser'
+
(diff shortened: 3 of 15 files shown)


packages/api-explorer/src/components/NotFound/NotFound.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { ApixHeading } from '../common'
+
+interface NotFoundProps {
+  message?: string
+}
+
+export const NotFound: FC<NotFoundProps> = ({ message }) => (
+  <Space p="xxxlarge" width="100%" height="50vh" around>
+    <ApixHeading as="h1">
+      {message || 'The page you requested could not be found.'}
```
jkaster, line 37: Does this end up with the "not found" url being the URL that wasn't found? If not, we should show what the requested URL was, perhaps. It would also be cool to make recommendations for the correct URL, if feasible.

jkaster (approved):
Better handling than we had before

packages/api-explorer/src/components/NotFound/NotFound.tsx
josephaxisa: It does end up with the not found url as the url. I didn't register a route for not found, but we can do that too.

packages/api-explorer/src/routes/utils.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Validates a matched route's params and renders the scene if valid
+ * @param match A react router object containing route match information
+ * @param routeType The type of route to validate
+ * @param api Api spec
+ * @param scene Scene to render if route params are valid
+ */
+export const validateAndRender = (
+  match: any,
```
AprilArcus, line 63: react router has a generic type you can use here:

```
import { match } from 'react-router'
...
interface HomeRouteParams {
  specKey: string
}
interface TagRouteParams {
  specKey: string
  methodTag: string
}
interface MethodRouteParams {
  specKey: string
  methodTag: string
  methodName: string
}
interface TypeRouteParams {
  specKey: string
  typeName: string
}

type RouteParams =
  | HomeRouteParams 
  | TagRouteParams
  | MethodRouteParams
  | TypeRouteParams

...

export const validateAndRender = (
  match: match<Params>
  ...
```

packages/api-explorer/src/routes/utils.tsx
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Checks if spec key matches API spec version
+ */
+export const validSpecKey = (api: ApiModel, specKey: string) =>
+  api.version.startsWith(specKey)
+
+/**
+ * Checks if api spec contains tag
+ */
+export const validTag = (api: ApiModel, tag: string) => tag in api.tags
```
AprilArcus, line 40: If it's useful to you, validator methods that return booleans can make type assertions about their arguments if you want, e.g. `=> tag in api.tags as tag is TagList`

https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

packages/api-explorer/src/components/DocMarkdown/DocMarkdown.tsx
```diff
@@ -56,7 +56,7 @@ export const DocMarkdown: FC<DocMarkdownProps> = ({ source, specKey }) => {
   const handleClick = (e: BaseSyntheticEvent) => {
     if (
       e.target.tagName === 'A' &&
-      e.target.origin === window.location.origin
+      e.target.pathname.startsWith(`/${specKey}`)
```
AprilArcus, line 59: consider instantiating a history singleton in `index.ts` and exporting it:

```
import { createBrowserHistory } from 'history'
export const browserHistory = createBrowserHistory()
<Router history={browserHistory}>
  ...
</Router>
```

then you can

```
import { browserHistory } from '.'
...
browserHistory.matchPath(e.target.pathname, `/${specKey}`)
```

this could give you a little more type safety down the line, if e.g. you wanted to change specKey to an enum from a routes object.

packages/api-explorer/src/routes/AppRouter.tsx
```diff
@@ -47,18 +49,33 @@ export const AppRouter: FC<AppRouterProps> = ({ specKey, api }) => {
...
-      <Route path="/:specKey/methods/:methodTag" exact>
-        <TagScene api={api} />
-      </Route>
-      <Route path="/:specKey/methods/:methodTag/:methodName">
-        <MethodScene api={api} />
-      </Route>
-      <Route path="/:specKey/types/:typeName">
-        <TypeScene api={api} />
-      </Route>
+      <Route
```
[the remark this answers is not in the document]
bryans99: Alternatively, create specific validation methods for home, method, type and tag. I also think the method should return a boolean rather than render the component. Something like
```
      <Route
        path="/:specKey/(methods|types)?"
        render={({ match }) => {
          return validHomeRoute(match, api)  ? <HomeScene api={api} /> : <NotFound/>
        }
        exact
      />

File: packages/api-explorer/src/routes/utils.tsx [hunk 6]
```diff
@@ -0,0 +1,83 @@
...
+export const validateAndRender = (
+  match: any,
+  routeType: routeTypes,
+  api: ApiModel,
+  scene: ReactElement<any>
+) => {
+  let isValid = false
+
+  if (routeType === 'home') {
+    isValid = validSpecKey(api, match.params.specKey)
```
bryans99, line 71: Shouldn't you validate the spec for all routes? I didn't get notfound when I used a spec of 9.9

image

bryans99:
@jkaster 
A thought occurs to me. Would it not be better to add the validate to the API spec code? Accepts, spec, method/type and tag. Just a thought

jkaster:
We could add the validate to the API spec code since currently we're "registering" the allowed specs. When we serve the OpenAPI version of the spec directly from Looker we'll no longer be bundling them and will instead interrogate the Looker server to find out what specifications are available. But validation could *still* be performed where the specs are determined, and we should definitely throw a "not found" for an invalid spec.

PR #345, opened 2020-10-07
TS SDK oauth session code-verifier tweak

joeldodge79:
For now it must not be 2^n so that Looker correctly treats it as base64 encoded


joeldodge79:
`bin/smoke typescript` passes

jkaster (approved):
LGTM. Thanks for the fix!

PR #352, opened 2020-10-08
Restructured and added routing

josephaxisa:
hackapp

### DIFF
--- a/packages/hackathon/package.json
+++ b/packages/hackathon/package.json
@@ -48,7 +48,7 @@
     "react-router-dom": "^5.2.0",
     "redux": "^4.0.5",
     "redux-saga": "^1.1.3",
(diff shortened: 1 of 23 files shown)


packages/hackathon/package.json
```diff
@@ -47,7 +47,8 @@
     "react-redux": "^7.2.1",
     "react-router-dom": "^5.2.0",
     "redux": "^4.0.5",
-    "redux-saga": "^1.1.3"
+    "redux-saga": "^1.1.3",
+    "styled-components": "^5.2.0"
```
jkaster, line 51: This version of styled components doesn't conflict with components?

jkaster (changes requested):
I think you need to downgrade the styled components version

PR #371, opened 2020-10-29
Rename IRequest related variable and method names

joeldodge79:
Make the logic be more about existing language features than
needing to introduce crutches

### DIFF
--- a/packages/sdk-codegen/src/codeGen.ts
+++ b/packages/sdk-codegen/src/codeGen.ts
@@ -180,10 +180,11 @@ export interface ICodeGen {
   codeQuote: string
 
   /**
-   * Does this language require request types to be generated because it doesn't
-   * conveniently support named default parameters?
+   * Does this language support named parameters with default values? Otherwise
+   * request types for sdk method signatures may be generated:
+   * see IMethod.eligibleForRequestType
    */
-  needsRequestTypes: boolean
+  useNamedParameters: boolean
 
   /** Does this language have streaming methods? */
   willItStream: boolean
@@ -622,7 +623,7 @@ export interface ICodeGen {
 }
 
 export abstract class CodeGen implements ICodeGen {
-  needsRequestTypes = false
+  useNamedParameters = false
   willItStream = false
   codePath = './'
   packagePath = 'looker'
@@ -1065,8 +1066,8 @@ export abstract class CodeGen implements ICodeGen {
   }
 
   useRequest(method: IMethod) {
-    if (!this.needsRequestTypes) return false
-    return method.mayUseRequestType()
+    if (this.useNamedParameters) return false
+    return method.eligibleForRequestType()
   }
 
   // Looks up or dynamically creates the request type for this method based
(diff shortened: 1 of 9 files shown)


joeldodge79:
As I was adding `useNamedArguments` for python makeTheCall I realized that I like this gesture better. But happy to discuss if you disagree

jkaster (dismissed):
Good rename. It provides better clarity for what exactly that generation toggle means.

packages/sdk-codegen/src/codeGen.ts
```diff
@@ -622,7 +623,7 @@ export interface ICodeGen {
 }
 
 export abstract class CodeGen implements ICodeGen {
-  needsRequestTypes = false
+  useNamedParameters = false
```
joeldodge79, line 626: whoops, this should default true in keeping with prior default
```suggestion
  useNamedParameters = true
```

packages/sdk-codegen/src/typescript.gen.ts
```diff
@@ -62,7 +62,6 @@ export class TypescriptGen extends CodeGen {
 
   indentStr = '  '
   endTypeStr = '\n}'
```
joeldodge79, line 64:
```suggestion
  endTypeStr = '\n}'
  useNamedParameters = false
```

packages/sdk-codegen/src/csharp.gen.ts
```diff
@@ -137,7 +137,7 @@ export class CSharpGen extends CodeGen {
 
   indentStr = '  '
   endTypeStr = '\n}'
-  needsRequestTypes = false
+  useNamedParameters = true
```
joeldodge79, line 140:
```suggestion
```

packages/sdk-codegen/src/kotlin.gen.ts
```diff
@@ -55,7 +55,7 @@ export class KotlinGen extends CodeGen {
 
   indentStr = '  '
   endTypeStr = '\n) : Serializable'
-  needsRequestTypes = false
+  useNamedParameters = true
```
joeldodge79, line 58:
```suggestion
```

packages/sdk-codegen/src/python.gen.ts
```diff
@@ -50,6 +50,7 @@ export class PythonGen extends CodeGen {
   dataStructureDelimiter = ', '
   enumDelimiter = '\n'
   codeQuote = '"'
+  useNamedParameters = true
```
joeldodge79, line 53:
```suggestion
```

packages/sdk-codegen/src/swift.gen.ts
```diff
@@ -55,7 +55,7 @@ export class SwiftGen extends CodeGen {
 
   indentStr = '    '
   endTypeStr = `\n}`
-  needsRequestTypes = false
+  useNamedParameters = true
```
joeldodge79, line 58:
```suggestion
```

joeldodge79 (commented):
@jkaster I goofed on the default value for `useNamedParameters`. I updated it to `true` overrode to `false` for typescript, and removed unnecessary declaration in the other languages.

If that looks ok I'll commit these and need another review. Otherwise, if you'd prefer the default for `useNamedParameters` to be `false` then I'll leave it as is (but I think more languages support it than not in general)

josephaxisa (approved):
This all looks ok to me

PR #372, opened 2020-10-30
Python makeTheCall

joeldodge79:
Implemented python calling syntax generation. Also modified style of
typescript call syntax


jkaster (approved):
Definitely a step in the right direction.

josephaxisa (approved):
🚢  it!

PR #379, opened 2020-11-10
Add ktlint to Kotlin project

kalenp:
Added `ktlint` to the Kotlin project.  This required a few steps first.  `ktlint` didn't like the way we were including comments inline in function parameter lists, so I first had to pull those up into proper KDoc comments.  I made another little change to get the proper indentation in the generated code, but then stopped the process of aligning the generated code with the linted code once it was mostly in agreement.

The final step was to enable `ktlint` in the kotlin gradle project and apply it to all of the code.  I have not yet added automatic linting support for the generated files.  To manually apply it, run `./gradlew spotlessApply` in the `kotlin` directory.  This will reformat both the RTL and SDK files.


packages/sdk-codegen/src/kotlin.gen.ts
```diff
@@ -236,11 +236,10 @@ ${note}
   }
 
   encodePathParams(indent: string, method: IMethod) {
-    const bump = indent + this.indentStr
```
jkaster, line 231: this was outdated code, and should have been 
```ts
```

I'll have to think a bit more about using `indent` rather than the `bump indent` pattern. Indent may be getting used wrong here.

jkaster (approved):
Excellent code improvements and I'm really glad to see the linter tooling added. Thanks!

PR #380, opened 2020-11-10
API spec diff utility

josephaxisa:
- Added a `signature` method to `Method` and `Parameter`
- Added an `id` property to `Method` for lookup purposes
- Added a `summary` method to both `Property` and `Parameter`
- Added a `compareSpecs` utility which compares a Method's type and its params across two specs and computes a diff
- Added `yarn lint:spec` that runs a script which uses the `compareSpecs` utility to compare Looker API 3.1 beta endpoints with their 4.0 version and writes the result to csv
- Added `CHANGELOG.md` to `packages/sdk-codegen`


packages/sdk-codegen/src/specLinter.ts
```diff
@@ -0,0 +1,203 @@
...
+}
+
+/**
+ * Given two specs, compares each method's params and response type, and returns
+ * a string representing representing a csv containing the difference if any
+ * @param lSpec
+ * @param rSpec
+ */
+
+export type preDiffCb = ({ lMethod, rMethod }: ComputeDiffInputs) => boolean
```
jkaster, line 155: Not sure these params match to the type signature now, and the blank line should be removed

packages/sdk-codegen/src/specLinter.ts
```diff
@@ -0,0 +1,203 @@
...
+    if (preDiffCb && !preDiffCb({ lMethod, rMethod })) return
+
+    while (
+      lMethod.name !== rMethod.name &&
+      rMethod.name < lMethod.name &&
+      rIndex < rLength
+    ) {
+      /** Case when right method does not exist on the left */
+      if (preDiffCb && preDiffCb({ rMethod })) {
+        diff.push(computeDiff({ rMethod }))
```
jkaster, line 186: I'm having trouble following this. if `rMethod` doesn't exist, what is the diff that's computed? How is the `lMethod` value known for the diff?

packages/sdk-codegen/src/specLinter.ts
```diff
@@ -0,0 +1,203 @@
...
+      rMethod = rMethods[rIndex]
+    }
+
+    if (lMethod.name === rMethod.name) {
+      /** Case when method exists on both sides */
+      diff.push(computeDiff({ lMethod, rMethod }))
+      rIndex = rIndex + 1 < rLength ? rIndex + 1 : rIndex
+    } else {
+      /** Case when left method does not exist on the right */
+      diff.push(computeDiff({ lMethod }))
```
jkaster, line 198: same question as above. Guess I'll need to debug this to understand it, but that indicates code smell to me

jkaster (changes requested):
Some minor changes

packages/sdk-codegen/src/specLinter.ts
josephaxisa: My mistake. That's an extra comment that should be removed.

packages/sdk-codegen/src/specLinter.ts
josephaxisa: The logic for that is in [`computeDiff`](https://github.com/looker-open-source/sdk-codegen/blob/4efbc8e4414dfe6e97256ed4bbe9b211a4829e1c/packages/sdk-codegen/src/specLinter.ts#L119-L146). If only one of the left or right methods are passed in the `paramsDiff` and `typeDiff` are empty.

jkaster (approved):
LGTM

jkaster (approved):
Looks good! This will be great to add to API Explorer as a visual diff tool

PR #381, opened 2020-11-11
Hackathon app Looker extension

jkaster:
Getting close on the PR. This has the latest master with current Looker components merged.

The Hackathon app now uses DataTables instead of ActionLists and looks much better for it.

The HackerList.tsx still has some errors that show up in console logs to fix, but everything appears to be working.

Consider this a draft PR for now. I don't see a way to tag it as such. Maybe that's not available for OSS projects?


packages/hackathon/src/scenes/UsersScene/components/HackerList.tsx
```diff
@@ -80,9 +81,9 @@ export const HackerList: FC<HackerListProps> = ({
   const rows = hackers
     .slice(startIdx, startIdx + PAGE_SIZE)
     .map((hacker, idx) => (
-      <DataTableCell key={idx}>
+      <DataTableItem key={idx} id={idx.toString()}>
```
bryans99: Why is the id needed now?

Does id map to the html id? If there are multiple data table instances in a scene using this pattern for id do we end up with duplicate ids? Prefix with "Hacker-" maybe. I'm not sure and I dont know the purpose of this at this time

packages/hackathon/src/scenes/UsersScene/components/HackerList.tsx
```diff
@@ -81,7 +71,19 @@ export const HackerList: FC<HackerListProps> = ({
   const rows = hackers
     .slice(startIdx, startIdx + PAGE_SIZE)
     .map((hacker, idx) => (
-      <DataTableItem key={idx} id={idx.toString()}>
+      <DataTableItem
+        key={idx}
+        id={idx.toString()}
```
bryans99, line 76: See other comment regarding id

bryans99 (commented):
Overall I think its okay but I made a lot of the changes :D 

I do have some questions about id on data table item. 

I'll approve once I get the answer to those but at a minimum we should all approve I think

packages/hackathon/src/scenes/UsersScene/components/HackerList.tsx
lukelooker: `id` is used to power the `select` features of `DataTable` (demo at: https://components.looker.com/storybook/?path=/story/datatable--select).

Admittedly, in an ideal world we'd be able to use `key` but React doesn't let us have access to that for reasons I can't entirely argue with.

lukelooker: Oops... hit enter too fast. 

I suppose we could make it optional if there was a strong argument against it but I'd like `select` to "just work" when it's enabled and we'd have to build for a case where some rows are missing IDs and decide how to handle that if it wasn't required.

packages/hackathon/src/scenes/UsersScene/components/HackerList.tsx
lukelooker: I'm going to take it as a project to eventually support additional (simple) types besides `string` (`number` seems reasonable at the very least)

packages/hackathon/src/scenes/UsersScene/components/HackerList.tsx
bryans99: Thanks for the explanation. Wondering is selectId might be a better name to avoid confusion

bryans99 (approved):
LGTM

PR #383, opened 2020-11-18
Moving sdk-specific constants to the typescript SDK rather than the RTL package

jkaster:
- sdkVersion and environmentPrefix are now defined in the packages/sdk/constants.ts file
- the TS code generator now stamps this file
- these values are "pushed down" into the runtime package
- environment reading is now an opt-out by passing an empty environment prefix string
- all downstream files and packages have been updated to use the new initialization parameters
- all sdk-related tests are passing
- kotlin generator unit tests are also fixed
- change logs added and/or updated
- TS SDK readme updated


josephaxisa (commented):
Nicely done. I left a few minor suggestions below

packages/hackathon/src/authToken/extensionProxyTransport.ts
```diff
@@ -57,7 +56,7 @@ export class ExtensionProxyTransport extends BaseTransport {
     authenticator?: Authenticator,
     options?: Partial<ITransportSettings>
   ) {
-    const agentTag = options?.agentTag || `${agentPrefix} ${lookerVersion}`
+    const agentTag = options?.agentTag || agentPrefix
```
joeldodge79, line 59: does `options?.agentTag` already have the `lookerVersion renamed to sdkVersion` info embedded? Or are we ditching that piece from the agentTag?

joeldodge79 (changes requested):
Nice work John! I left a few comments/questions.

It feels a little odd to me that passing an empty string (and documenting its passage) down from `NodeSettings`'s constructor into various apiSettings.ts functions ultimately to `ApiConfigMap` which knows what to do with an empty string.

I could be convinced otherwise but I think a different approach in which somehow the code is re-organized such that passing an empty string to `NodeSettings`'s constructor explicitly by-passes any code path that reads from the env. Otherwise reading this "turn off env var reading" feature implementation is a bit scattered.

That being said, I'm ok with it as is and if you were to heed my suggestion that could be a future refactor.

packages/sdk-codegen/src/kotlin.gen.spec.ts
```diff
@@ -59,29 +59,26 @@ describe('Kotlin generator', () => {
...
-   * A hyphenated property name (read-only)
-   */
-  @JsonProperty("project-digest")
-  var project_digest: String? = null,
-  /**
-   * A spaced out property name (read-only)
-   */
-  @JsonProperty("computation time")
-  var computation_time: Float? = null
+      const expected = `/**
```
kalenp, line 71: Oh, what command should I have been running to catch these changes?

packages/sdk-rtl/src/nodeSession.ts
```diff
@@ -33,26 +33,19 @@ import {
...
 import { AuthSession, IAccessToken, IError } from './authSession'
-import { defaultApiVersion } from './constants'
-import { getenv } from './nodeSettings'
 
 const strPost: HttpMethod = 'POST'
 const strDelete: HttpMethod = 'DELETE'
 
 export class NodeSession extends AuthSession {
-  private readonly apiPath: string = ''
+  private readonly apiPath: string = '/api/3.1'
```
kalenp, line 43: Even though you've removed this from the RTL constants, seems like you would still want it to be configurable when being initialized as part of a particular SDK.

kalenp (commented):
At the very end, I see that you removed the two json files describing the APIs.  If we want to remove those, we should add `.gitignore` entries for them, so they don't accidentally creep back in.  

On the other hand, it seems like we should be able to run this without a running looker just based on those files, which is something that I'm still not sure how to do.  But if that's a goal, then perhaps we want to keep the json files present as the source that the rest of the SDK is built off of.  Then when we wanted to rev for a new Looker version, you would explicitly pull the json specs, rather than it always being based off of "whatever the dev has running".

packages/sdk-rtl/src/nodeSession.ts
jkaster: Agreed. I was deferring that until 3.1 is deprecated or we need something in a more recent API version. Can certainly come back to this.

packages/sdk-codegen/src/kotlin.gen.spec.ts
jkaster: If I removed the Looker.3.1.oas.json or Looker.4.0.oas.json that was a commit mistake.

`yarn jest packages/sdk-codegen/` would run those tests. Still need to work on our test suite scripts.

packages/hackathon/src/authToken/extensionProxyTransport.ts
jkaster: the extension proxy agentTag should be initialized at this point. If it isn't, it defaults.

packages/sdk-rtl/src/apiSettings.ts
```diff
@@ -25,30 +25,26 @@
...
-
-export const ApiConfigMap: IValueSettings = {
-  base_url: strLookerBaseUrl,
-  client_id: strLookerClientId,
-  client_secret: strLookerClientSecret,
-  timeout: strLookerTimeout,
-  verify_ssl: strLookerVerifySsl,
+/**
+ * Gets list of environment variable names for config values
+ * @param envPrefix Environment naming prefix to use. Leave blank to return no variable keys
```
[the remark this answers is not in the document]
jkaster: Agreed this should definitely go through another refactor for the environment setting options. I'm trying to keep the changes in this PR to a minimum, and it had lots to fix.

packages/sdk-rtl/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+- The SDK environment variable prefix can now be set to an empty string to avoid reading environment variables in [NodeSettingsIniFile() and NodeSettings()](src/nodeSettings.ts)
```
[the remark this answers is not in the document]
jkaster: Documenting it better in the readme would be the right place, I think.

packages/sdk-rtl/src/nodeSettings.spec.ts
```diff
@@ -150,34 +147,52 @@ timeout=30
 
   describe('NodeSettingsIniFile', () => {
     xit('settings default to the first section', () => {
```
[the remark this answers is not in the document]
jkaster: That's a test that's been skipped since before this PR. I could see if it passes now, but I'm trying to change only what needed to be changed.

packages/run-it/src/utils/RunItSDK.spec.ts
```diff
@@ -37,21 +33,19 @@ import { RunItSettings } from './RunItSDK'
 const settings = {
   ...DefaultSettings(),
   base_url: 'https://self-signed.looker.com:19999',
-  agentTag: `RunIt ${lookerVersion}.4.0`,
+  agentTag: `RunIt `,
```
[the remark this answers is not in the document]
jkaster: Actually, that shouldn't have a trailing space!

package.json
```diff
@@ -156,7 +156,13 @@
         "config/license-header.js"
       ],
       "import/default": "off",
-      "sort-keys": "off"
+      "sort-keys": "off",
+      "spaced-comment": [
+        "error",
+        "always", {
+          "markers": ["#"]
```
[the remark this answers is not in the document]
jkaster: It's an attempt to get region markers to pass linting, which isn't yet working. I have an open question about that.

packages/sdk/src/node.ts
```diff
@@ -24,11 +24,11 @@
 
  */
 
-import * as models31 from './sdk/3.1/models'
-export * from './sdk/3.1/methods'
+import * as models31 from './3.1/models'
+export * from './3.1/methods'
 export { models31 }
```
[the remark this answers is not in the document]
jkaster: Probably because I didn't know that syntax at the time I was writing this the first time 😉

packages/sdk-codegen/src/kotlin.gen.spec.ts
jkaster: I'm not seeing removal of the needed spec files. That folder is git ignored by default. When API Explorer no longer needs to bundle the OpenAPI spec files @josephaxisa or I will again remove the `*.oas.json` files from the repo.

jkaster:
@bryans99 you're a reviewer because this low-level change breaks the extension framework a bit

packages/sdk-codegen/src/kotlin.gen.spec.ts
joeldodge79: @jkaster, regarding @kalenp's comment about removed files (and hence the -33k loc) you need to:

```
git reset origin/master spec/Looker.3.1.oas.json spec/Looker.4.0.oas.json
git checkout origin/master spec/Looker.3.1.oas.json spec/Looker.4.0.oas.json
git commit -am 'undeleting oas spec files
```

joeldodge79 (changes requested):
still seeing the OAS spec files (`spec/Looker.3.1.oas.json` and `spec/Looker.4.0.oas.json`) deleted in this PR

jkaster:
Thanks @joeldodge79 I'll see if I can figure out why my local view is different than the GH view

jkaster:
Thanks @joeldodge79 I fixed the inadvertent deletion my IDE was lying about ;)

joeldodge79:
I'm ready to give an LGTM when you're ready for it

joeldodge79 (approved):
👍

PR #384, opened 2020-11-19
Refactor Kotlin RTL

kalenp:
The goals of this PR are twofold:
- Improve the handling of errors
- Improve the typing of responses

This mostly revoles around replacing SDKResponse with
SdkResult<TSuccess, TFailure>.  This allows us to track the type
information with the SdkResult and easily get out the expected type on
success or failure.

For now, all errors are expected to be of type com.looker.sdk.Error,
even though we do have some methods which return more precise errors.
If for some reason you need to escape the default typing, such as to get
a finer Error type, you can use `bodyAs<T>()` on both SuccessResponse
and FailureResponse.

For clients, this commit keeps the concept of the `ok()` method that
returns the success type or throws an error.  This moved from being a
method on the sdk itself to an extension function on the SdkResult,
so you can do `val foo: Foo = sdk.get_foo().ok()`.  If you want to do
more precise handling, SdkResult is a `sealed` class, so you can use a
`when` statement to check for the various type (SuccessResonse,
FailureResponse, Error) and get smart casting to access the appropriate
types and Response metatdata.

### DIFF
--- a/kotlin/src/main/com/looker/rtl/SdkResult.kt
+++ b/kotlin/src/main/com/looker/rtl/SdkResult.kt
@@ -1,0 +1,104 @@
+package com.looker.rtl
+
+import io.ktor.client.call.receive
+import io.ktor.client.response.HttpResponse
+import io.ktor.http.isSuccess
+import kotlinx.coroutines.runBlocking
+
+class FailureResponseError(val result: SdkResult<*, *>) : Exception()
+
+interface SdkResponse<T> {
+    val success: Boolean
+    val statusCode: Int
+    val method: HttpMethod
+    val path: String
+    fun body(): T
+}
+
+sealed class SdkResult<out TSuccess, out TFailure> {
+    abstract val success: Boolean
+    abstract val method: HttpMethod
+    abstract val path: String
+
+    data class SuccessResponse<TSuccess>(
+        val response: HttpResponse,
+        override val method: HttpMethod,
+        override val path: String,
+        private val body: TSuccess
+    ) : SdkResponse<TSuccess>, SdkResult<TSuccess, Nothing>() {
+        override val success: Boolean = true
+        override val statusCode: Int = response.status.value
+
+        override fun body(): TSuccess = body
+
+        inline fun <reified T> bodyAs(): T {
+            return runBlocking { response.receive<T>() }
+        }
+    }
+
+    data class FailureResponse<TFailure>(
+        val response: HttpResponse,
+        override val method: HttpMethod,
+        override val path: String,
+        private val body: TFailure
+    ) : SdkResponse<TFailure>, SdkResult<Nothing, TFailure>() {
+        override val success: Boolean = false
+        override val statusCode: Int = response.status.value
+
+        override fun body(): TFailure = body
+
+        inline fun <reified T> bodyAs(): T {
+            return runBlocking { response.receive<T>() }
+        }
+    }
+
+    data class Error(
+        val error: Throwable,
+        override val method: HttpMethod,
+        override val path: String
+    ) : SdkResult<Nothing, Nothing>() {
+        override val success: Boolean = false
+    }
+
+    companion object {
+        inline fun <reified TSuccess, reified TFailure> response(
+            response: HttpResponse,
+            method: HttpMethod,
+            path: String
+        ): SdkResult<TSuccess, TFailure> {
+            try {
+                if (response.status.isSuccess()) {
+                    val body = runBlocking { response.receive<TSuccess>() }
+                    return SuccessResponse<TSuccess>(response, method, path, body)
+                } else {
+                    val body = runBlocking { response.receive<TFailure>() }
+                    return FailureResponse<TFailure>(response, method, path, body)
+                }
+            } catch (ex: Exception) {
+                return error(ex, method, path)
+            }
+        }
+
+        fun <TSuccess, TFailure> error(
+            error: Throwable,
+            method: HttpMethod,
+            path: String
+        ): SdkResult<TSuccess, TFailure> {
+            return SdkResult.Error(error, method, path)
+        }
+    }
+}
+
+fun <TSuccess> SdkResult<TSuccess, *>.ok(): TSuccess {
+    return when (this) {
+        is SdkResult.SuccessResponse<TSuccess> -> {
+            this.body()
+        }
+        is SdkResult.FailureResponse<*> -> {
+            throw FailureResponseError(this)
+        }
+        is SdkResult.Error -> {
+            throw this.error
+        }
+    }
+}
--- a/kotlin/src/main/com/looker/rtl/Transport.kt
+++ b/kotlin/src/main/com/looker/rtl/Transport.kt
@@ -26,7 +26,6 @@ package com.looker.rtl
 
 import io.ktor.client.HttpClient
 import io.ktor.client.call.call
-import io.ktor.client.call.receive
 import io.ktor.client.engine.okhttp.OkHttp
 import io.ktor.client.features.json.JacksonSerializer
 import io.ktor.client.features.json.JsonFeature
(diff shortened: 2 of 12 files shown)


kotlin/src/main/com/looker/rtl/Transport.kt
```diff
@@ -266,38 +228,29 @@ class Transport(val options: TransportOptions) {
...
 
-    inline fun <reified T> request(
+    inline fun <reified TSuccess, reified TFailure> request(
         method: HttpMethod,
         path: String,
         queryParams: Values = mapOf(),
         body: Any? = null,
         noinline authenticator: Authenticator? = null
-    ): SDKResponse {
-        // TODO get overrides parameter to work without causing compilation errors in UserSession
```
kalenp, line 276: Do I need to preserve these TODOs?

packages/sdk-codegen/src/kotlin.gen.ts
```diff
@@ -58,6 +58,8 @@ export class KotlinGen extends CodeGen {
   needsRequestTypes = false
   willItStream = true
 
+  tFailure = 'com.looker.sdk.Error'
```
kalenp, line 61: I used the fully qualified name here, since Kotlin also defines an `Error` class.  Maybe the API response type should be clarified to use something that doesn't conflict.

packages/sdk-codegen/src/kotlin.gen.ts
```diff
@@ -404,27 +410,31 @@ ${props.join(this.propDelimiter)}
...
-    }>(${this.httpPath(method.endpoint, request)}${args ? ', ' + args : ''})`
+    return `${indent}return ${methodName}<${tSuccess}, ${this.tFailure}>(
+${bump}${functionArgs.join(', ')}
+${indent})`
   }
 
   streamCall(indent: string, method: IMethod) {
     const request = this.useRequest(method) ? 'request.' : ''
+    const methodName = this.it(method.httpMethod.toLowerCase())
+    const tSuccess = 'ByteArray'
```
kalenp, line 430: For streams, we always return ByteArray for TSuccess.  I kept `com.looker.sdk.Error` for TFailure, but maybe that should be ByteArray as well?  Or maybe the streaming methods should also do deserialization?  I don't know much about how we're handling streaming methods in the other SDKs, so I tried to change as little here as possible.

kalenp (commented):
Pushing this up to get some feedback while I do a final cleanup pass.

kotlin/src/main/com/looker/rtl/SdkResult.kt
```diff
@@ -0,0 +1,104 @@
...
+
+interface SdkResponse<T> {
+    val success: Boolean
+    val statusCode: Int
+    val method: HttpMethod
+    val path: String
+    fun body(): T
+}
+
+sealed class SdkResult<out TSuccess, out TFailure> {
```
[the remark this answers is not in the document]
joeldodge79: I know you're just augmenting the existing pattern of an sdk method returns a "meta" object that you can get the actual SDK response type from (on success) with `.body()` or use `.ok()` to return that or throw. And this structure seems reasonable to me.

My question is at a higher design level for you and @jkaster - searching through some pretty popular python sdks (github, azure, google cloud sdks) - the idiomatic approach for these python sdks is for a method to behave as if the `sdk.ok()` (or `response.ok()` here) method is automatically called (i.e. the sdk method either raises or returns the `TSuccess` object directly)

For python I'd actually propose adding a method on the SDK response types that exposes the success/status_code/method/path/raw_body metadata (and for the error case throwing an exception that has this same metadata access method).

Is it ok from our design perspective that the python sdk behaves so differently? We could make it behave like the others but then I think we lose the idiomatic nature that other popular python sdk libraries are following.

thoughts?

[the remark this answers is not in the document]
kalenp: It's definitely simpler if the sdk response is always just TSuccess or an exception.  I guess since my motivation for this was better error handling, we could limit the extra functionality to just the error path.  That would limit the client's ability to introspect on the response in the success case, but maybe we don't care about that.  Also, I remember @jkaster talking about how he didn't want to throw in the failure/error case until you actually went to get the value.  Something about working better in an async flow, so maybe we should look at what node or something else natively async does here.

I'd be up for a quick group chat about this topic.

kotlin/src/main/com/looker/rtl/SdkResult.kt
```diff
@@ -0,0 +1,104 @@
...
+            response: HttpResponse,
+            method: HttpMethod,
+            path: String
+        ): SdkResult<TSuccess, TFailure> {
+            try {
+                if (response.status.isSuccess()) {
+                    val body = runBlocking { response.receive<TSuccess>() }
+                    return SuccessResponse<TSuccess>(response, method, path, body)
+                } else {
+                    val body = runBlocking { response.receive<TFailure>() }
```
kalenp, line 74: Pondering this some more, I'm wondering if the FailureResponse case should actually defer on receiving the data until requested.  The issue I'm running into while thinking this through is that I want to be able to deal with a single API being able to return multiple different types of error response.  In that case, we're going to have to do some preliminary parsing or some try/catch(ParseError) for the different types.  With ktor, you can only do `response.receive()` a single time, so that's not going to work.

I'm thinking that maybe we do `response.receive<ByteArray>()` or `response.receive<String>()` to get the "raw" data, and then parse that into whatever type is requested.  Of course that'll mean that we have to manage our own object mapper in the RTL instead of using ktor's registered one.  All around, it ends up being a lot more to manage here in the RTL layer, but I'm not seeing a good way around that.

Looking at the error responses and some of the libraries like Jackson or kotlinx.serialization, it would be nice to have a discriminator field to use to select the type to parse into, but our API doesn't really provide that.  

Just brain dumping into this comment for further discussion.

kotlin/src/main/com/looker/rtl/Transport.kt
jkaster: We need to functionality that TODO was for, but not necessarily in that place

packages/sdk-codegen/src/kotlin.gen.ts
jkaster: We rename the spec's `Error` class for Swift, which is less flexible. We may need to do the same kind of thing for the Kotlin generator.

packages/sdk-codegen/src/kotlin.gen.ts
jkaster: The kotlin "streaming" support was a total hack just to support binary payload downloading. Same is true for the Swift SDK. We still need real streaming functionality.

kotlin/src/main/com/looker/rtl/SdkResult.kt
joeldodge79: well, it will be a comfort python isn't the only language with a separate [de]serialization layer. I always thought it was unfair that the other languages just somehow got their models/types hydrated for free from the transport.

kotlin/src/main/com/looker/rtl/SdkResult.kt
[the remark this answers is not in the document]
joeldodge79: happy to participate - I'll let you and our fearless leader @jkaster coordinate :-)

kotlin/src/main/com/looker/rtl/SdkResult.kt
jkaster: ## Updated feedback

- the Kotlin SDK is now using GSON to ignore serializing null values into the request payload (so far, no complaints from the few users we have of that SDK, and it resolves some use cases with some endpoints). 
- It also appears that being able to deserialize a JSON payload to a Kotlin type outside of the request/response flow is a desirable feature as well, so we may want to separate that out into an independent function

kabat87 (approved):

PR #386, opened 2020-11-19
ApiExplorer: Modify app config to reflect dev server config changes

josephaxisa:
`devServer` is no longer running on `https` and `appconfig.json` needed to be updated.


jkaster (approved):
Is this because of the change made for the Components update?

josephaxisa:
That's it. We're no longer running the dev server with the `https` flag

PR #388, opened 2020-11-20
Python RTL transport interface headers cleanup

### LINKED ISSUE #387: Python RTL transport interface: headers -> transport_options
The other languages have headers as part of the transport options bag, python should do the same:

This way sdk callers can pass in headers as part of the transport_options to individual calls

joeldodge79:
Moved headers down into the transport_options bag.
Called out the explicit "per-request" transport options shape.
Fixed example that broke a while ago :-()

closes #387

### DIFF
--- a/examples/python/Pipfile
+++ b/examples/python/Pipfile
@@ -12,7 +12,7 @@ verify_ssl = true
 
 [packages]
 looker-sdk = "*"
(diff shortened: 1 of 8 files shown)


josephaxisa (approved):
LGTM

jkaster (approved):
LGTM

PR #390, opened 2020-11-23
Completed smoke test pattern tests for TS SDK

jkaster:
- created `bin/unit_test` script to run as part of and before `bin/smoke`
- `bin/unit_test typescript` runs the `packages/sdk-rtl` tests
- `bin/smoke typescript` runs the unit tests and the `packages/sdk/test` tests
- JUnit XML-formatted results are named `results/ts_sdk_unit.xml` and `results/ts_sdk_smoke.xml`

### DIFF
--- a/bin/unit_test
+++ b/bin/unit_test
@@ -31,24 +31,24 @@ fi
 if [[ $tests == *python* ]]
 then
   cd python || exit
-  echo "Unit testing Python SDK ..."
-  pipenv run tox
+  echo "TODO: Unit testing Python SDK ..."
+  #pipenv run tox
   cd ..
 fi
 
 if [[ $tests == *kotlin* ]]
 then
   cd kotlin || exit
-  echo "Unit testing Kotlin SDK ..."
-  ./gradlew test
+  echo "TODO: Unit testing Kotlin SDK ..."
+  # ./gradlew test
   cd ..
 fi
 
 if [[ $tests == *swift* ]]
 then
   cd swift/looker || exit
-  echo "Unit testing Swift SDK ..."
-  swift test
+  echo "TODO: Unit testing Swift SDK ..."
+  # swift test
   cd ../..
 fi
 
--- a/bin/unit_test
+++ b/bin/unit_test
@@ -17,20 +17,21 @@ else
   tests="$*"
 fi
 
-./bin/unit_test "${tests}"
+yarn jsonify test/data.yml
+rm -rf results/
 
-echo "Smoking SDKs: ${tests} ..."
+echo "Unit testing SDKs: ${tests} ..."
 
 if [[ $tests == *typescript* ]]
 then
-  echo "Smoking Typescript SDK ..."
-  JEST_JUNIT_OUTPUT_DIR=results JEST_JUNIT_OUTPUT_NAME=ts_sdk_smoke.xml yarn jest packages/sdk/test --reporters=default --reporters=jest-junit
+  echo "Unit testing Typescript SDK ..."
+  JEST_JUNIT_OUTPUT_DIR=results JEST_JUNIT_OUTPUT_NAME=ts_sdk_unit.xml yarn jest packages/sdk-rtl --reporters=default --reporters=jest-junit
 fi
 
 if [[ $tests == *python* ]]
 then
   cd python || exit
-  echo "Smoking Python SDK ..."
+  echo "Unit testing Python SDK ..."
   pipenv run tox
   cd ..
 fi
@@ -38,7 +39,7 @@ fi
 if [[ $tests == *kotlin* ]]
 then
   cd kotlin || exit
-  echo "Smoking Kotlin SDK ..."
+  echo "Unit testing Kotlin SDK ..."
   ./gradlew test
   cd ..
 fi
@@ -46,9 +47,9 @@ fi
 if [[ $tests == *swift* ]]
 then
   cd swift/looker || exit
-  echo "Smoking Swift SDK ..."
+  echo "Unit testing Swift SDK ..."
   swift test
   cd ../..
 fi
 
-echo "Smoke tests completed"
+echo "Unit tests completed"
--- a/packages/sdk/test/methods.spec.ts
+++ b/packages/sdk/test/methods.spec.ts
@@ -36,9 +36,9 @@ import {
   NodeSession,
   ApiConfigMap,
 } from '@looker/sdk-rtl'
-import { TestConfig } from '../../../sdk-rtl/src/testUtils'
-import { LookerNodeSDK } from '../nodeSdk'
-import { Looker40SDK as LookerSDK } from '../4.0/methods'
+import { TestConfig } from '../../sdk-rtl/src/testUtils'
+import { LookerNodeSDK } from '../src/nodeSdk'
+import { Looker40SDK as LookerSDK } from '../src/4.0/methods'
 import {
   ICreateQueryTask,
   IQuery,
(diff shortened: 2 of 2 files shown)


packages/sdk/test/methods.spec.ts
```diff
@@ -182,6 +206,59 @@ describe('LookerNodeSDK', () => {
...
+      } else {
+        const dashboards = await sdk.ok(sdk.search_dashboards({ limit: 1 }))
+        expect(dashboards).toBeDefined()
+        if (dashboards.length > 0) {
+          type = 'dashboards'
+          id = dashboards[0].id!
+        }
+      }
+      expect(type).toBeDefined()
+      expect(id).toBeDefined()
```
joeldodge79, line 228: let me know if you have trouble getting the docker container running. I'm torn about this "defensive" testing :-) On the one hand assuming a look exists makes the test easier to read which I'd prefer. However, a previous test might have deleted all the looks? I guess in that case I'd rather we created a look than falling back to trying to find a dashboard all of which may have also been deleted?

IMHO we should just assume other test writers will not go around deleting pre-existing data that (unless it's critical and there's no way to create an object for deletion first). I'd say just assume a look exists and delete this fallback behavior.

packages/sdk/test/methods.spec.ts
```diff
@@ -182,6 +206,59 @@ describe('LookerNodeSDK', () => {
...
+        sdk.content_thumbnail({ type: type, resource_id: id, format: 'png' })
+      )
+      expect(image).toBeDefined()
+      expect(mimeType(image)).toEqual('image/png')
+      const svg = await sdk.ok(
+        sdk.content_thumbnail({ type: type, resource_id: id, format: 'svg' })
+      )
+      expect(svg).toBeDefined()
+      expect(svg).toMatch(/^<\?xml/)
+    })
```
joeldodge79, line 239: hurray for `content_thumbnail()`!

packages/sdk/test/methods.spec.ts
```diff
@@ -182,6 +206,59 @@ describe('LookerNodeSDK', () => {
...
+  describe('PUT smoke test', () => {
+    it('set default color collection', async () => {
+      const sdk = new LookerSDK(session)
+      const current = await sdk.ok(sdk.default_color_collection())
+      expect(current).toBeDefined()
+      const cols = await sdk.ok(sdk.all_color_collections())
+      const other = cols.find((c) => c.id !== current.id)
+      expect(other).toBeDefined()
+      // tests to stop lint from complaining
+      if (other && other.id && current.id) {
```
joeldodge79, line 251: rather than `!` everywhere? 
a/ I wish lint would recognize `expect(other).toBeDefined()` as verifying that `other` is not undefined/null
b/ dead horse but can't resist: wish our linting/typing recognized that `.id` must be non-undefined/null

all that being said, I think I'd prefer a `!` to silence the warning because reading this conditional I'm more likely to think "oh, maybe these things could actually be null?". Using `!` seems to more explicitly call out the problem we're working around?

I don't really feel strongly either way though, so ok as is

packages/sdk/test/methods.spec.ts
```diff
@@ -182,6 +206,59 @@ describe('LookerNodeSDK', () => {
...
+        const actual = await sdk.ok(sdk.set_default_color_collection(other.id))
+        expect(actual).toBeDefined()
+        expect(actual.id).toEqual(other.id)
+        const updated = await sdk.ok(sdk.default_color_collection())
+        expect(updated.id).toEqual(actual.id)
+        await sdk.ok(sdk.set_default_color_collection(current.id))
```
joeldodge79, line 257: no follow on assertions about the result of this line?

bin/smoke
```diff
@@ -17,15 +17,14 @@ else
...
-rm -rf results/
+./bin/unit_test "${tests}"
 
 echo "Smoking SDKs: ${tests} ..."
 
 if [[ $tests == *typescript* ]]
 then
   echo "Smoking Typescript SDK ..."
-  JEST_JUNIT_OUTPUT_DIR=results JEST_JUNIT_OUTPUT_NAME=smoketypescript.xml yarn jest packages/sdk --reporters=default --reporters=jest-junit
+  JEST_JUNIT_OUTPUT_DIR=results JEST_JUNIT_OUTPUT_NAME=ts_sdk_smoke.xml yarn jest packages/sdk/test --reporters=default --reporters=jest-junit
```
joeldodge79, line 27: curious why `packages/sdk/` has a separate `test/` dir? It seems like the convention is to have spec files alongside the source files? I can see the argument that `methods.spec.ts` doesn't unittest a particular file (in which case it might be better named something like `integration.spec.ts` or `smoke.spec.ts`) but still seems like the odd ball wrt to all the other `packages/` having their .spec.ts files side-by-side

bin/unit_test
```diff
@@ -0,0 +1,55 @@
...
+    then
+        echo "Test options: all, typescript, kotlin, swift, python"
+        echo "Multiple languages can be specified at the same time"
+        echo "'all' is the default"
+    fi
+done
+
+if [ $# == 0 ] || [ "$1" == "all" ]
+then
+  tests="typescript kotlin swift python"
```
joeldodge79, line 15: It would be nice to have an entry point that unittests the `packages/sdk-codegen*` packages as well. Maybe add a `"codegen"` argument here if can we combine the tests/results or `"codegen codegen-utils codegen-scripts"` arguments if we can't?

joeldodge79 (approved):
sweet! glad to see the start of separating out of unit and integration test entry points. I left several suggestions which I'm ok with you ignoring.

josephaxisa (dismissed):

bin/unit_test
```diff
@@ -0,0 +1,55 @@
...
+then
+  echo "Unit testing Typescript SDK ..."
+  JEST_JUNIT_OUTPUT_DIR=results JEST_JUNIT_OUTPUT_NAME=ts_sdk_unit.xml yarn jest packages/sdk-rtl --reporters=default --reporters=jest-junit
+fi
+
+if [[ $tests == *python* ]]
+then
+  cd python || exit
+  echo "Unit testing Python SDK ..."
+  pipenv run tox
```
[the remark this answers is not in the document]
jkaster: > seems like the same splitting still needs to happen for kotlin + swift?

Yes, this PR is only for TS

packages/sdk/test/methods.spec.ts
jkaster: When we have a consistent, known setup we can update this test. For now, I want to ensure it's testing the pattern we want, which is both binary and string payloads.

packages/sdk/test/methods.spec.ts
jkaster: Using the imperative makes the lint checker produce warnings. I don't like this either, but there's no other way (other than completely disabling the lint rule) to remove the warnings.

packages/sdk/test/methods.spec.ts
jkaster: I suppose I could, but it seems the only failure condition you might have at that point would be if the server goes away before or during that last "reset" call, and if that happens, you'll have other issues to deal with anyway?

bin/smoke
jkaster: You're right that the TS convention is to have the test file next to the file being tested. 

I thought it best to violate that rule for tests that use our generated SDK files. I think the only thing that should be in the generated directories is the generated code. Because of the way jest tests, we need a distinct path, because `yarn jest packages/sdk` would result in running the tests for:

- packages/sdk
- packages/sdk-rtl
- packages/sdk-codegen
- packages/sdk-codegen-scripts

which is definitely not what we want.

We could rename the test source file, but it _is_ testing the methods, after all.

bin/unit_test
jkaster: Agreed. That would be another PR.

joeldodge79 (approved):
LGTM!

PR #392, opened 2020-11-24
Python SDK smoke tests

josephaxisa:
- Removed duplicate 3.1 tests. The same tests exist for 4.0
- Move 4.0 `test_methods.py` from `/tests/integration/api40/` to `/tests/integration`
- Added a few more tests to meet are coverage requirements


python/tests/integration/test_methods.py
```diff
@@ -357,6 +379,47 @@ def test_crud_look(sdk: mtds.Looker40SDK, looks):
         assert not look.deleted
 
 
+def test_png_svg_downloads(sdk: mtds.Looker40SDK):
+    """content_thumbnail() should return a binary or string response based on the specified format."""
+    looks = sdk.search_looks(limit=1)
+    _id: str
+    if looks:
+        _type = "look"
+        _id = str(looks[0].id)
```
joeldodge79, line 388: maybe `resource_id` and `resource_type` ?
If you want to avoid keyword collision then the convention is trailing underscore (`id_` and `type_`):
> single_trailing_underscore_: used by convention to avoid conflicts with Python keyword

from [PEP8](https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles)

joeldodge79 (changes requested):
awesome, much cleaner! one small nit

jkaster (commented):
Looks good to me when you address Joel's feedback!

python/tests/integration/test_methods.py
josephaxisa: `resource_id` and `resource_type` it is

joeldodge79 (approved):
Thank you for being patient with my little mind's hobgoblin :-) maybe `black` will be smart enough one day to have `s/_id/id_/` for you automatically and then none of us has to keep this stupid trivia occupying space in our brains

PR #393, opened 2020-11-24
Upgrade kotlin version to 1.4.20

kalenp:
We want to play with kotlinx.serialization, which requires Kotlin 1.4, so time to upgrade.  While I was at it, I upgraded the gradle wrapper version and did a bit of cleanup of the gradle build file.


jkaster (approved):
Works great!

PR #396, opened 2020-11-25
Upgrade to ktor 1.4.2

kalenp:
This is a prerequisite for using kotlinx.serialization with the ktor
client, so time to upgrade before we experiment with that.


jkaster (approved):
Looks great. Thanks for the update!

PR #397, opened 2020-11-25
Kotlin smoke tests

jkaster:
Kotlin smoke tests

- separated smoke tests into TestSmoke.kt
- Added missing smoke test patterns
- Added `is...` `get...` and `set...` JsonProperty attributes to fix hidden serialization bug in Jackson
- updated methods and models to latest version
- `bin/smoke kotlin` now runs *only* the Kotlin smoke tests, not all Kotlin tests


packages/sdk-codegen/src/kotlin.gen.ts
```diff
@@ -162,9 +162,14 @@ import java.util.*
   declareProperty(indent: string, property: IProperty) {
     const optional = !property.required ? '? = null' : ''
     const type = this.typeMap(property.type)
-    const attr = property.hasSpecialNeeds
-      ? `${indent}@JsonProperty("${property.jsonName}")\n`
-      : ''
+    // Bypass jackson serialization bug https://stackoverflow.com/questions/32270422/jackson-renames-primitive-boolean-field-by-removing-is
```
joeldodge79, line 165: that's a good one!

kotlin/src/main/com/looker/sdk/3.1/methods.kt
```diff
@@ -3929,6 +3934,30 @@ class Looker31SDK(authSession: AuthSession) : APIMethods(authSession) {
...
+    /**
+     * ### Discover information about derived tables
+     *
+     * @param {String} model The name of the Lookml model.
+     * @param {String} format The format of the graph. Valid values are [dot]. Default is `dot`
+     * @param {String} color Color denoting the build status of the graph. Grey = not built, green = built, yellow = building, red = error.
+     *
+     * GET /derived_table/graph/model/{model} -> DependencyGraph
+     */
+    @JvmOverloads fun graph_derived_tables_for_model(
```
joeldodge79, line 3946: maybe you generated against master instead of 7.20? I think this is in 21.0 (MOD-2999)

joeldodge79 (changes requested):
looks good to me except I think you didn't mean to generate against master?

kotlin/src/main/com/looker/sdk/3.1/methods.kt
jkaster: yup. regenerated now.

joeldodge79 (approved):
LGTM

PR #398, opened 2020-11-25
Python RTL refactor for auth_{session,token}

josephaxisa:


jkaster (commented):
This looks OK to me, but I'll wait for Joel to provide feedback.

joeldodge79 (approved):
wow, so much cleaner! And this was so quick and easy to review, thanks for breaking up the PRs!!

PR #399, opened 2020-11-25
Added smoke and unit tests for Swift SDK

jkaster:
- added `smokeTests.swift` and moved some existing tests into it
- `bin/unit_test swift` runs the Swift unit tests
- `bin/smoke swift` runs only the Swift unit and smoke tests


bin/unit_test
```diff
@@ -48,7 +48,7 @@ if [[ $tests == *swift* ]]
 then
   cd swift/looker || exit
   echo "TODO: Unit testing Swift SDK ..."
```
joeldodge79, line 50:
```suggestion
  echo "Unit testing Swift SDK ..."
```

bin/unit_test
```diff
@@ -48,7 +48,7 @@ if [[ $tests == *swift* ]]
...
-  # swift test
+  swift test --filter "api|authToken|constants|models|transport"
```
joeldodge79, line 51: too bad there's not a more general filter that would catch these all

joeldodge79 (changes requested):
tiny bit of cleanup

bin/unit_test
jkaster: we can _probably_ move those into a `unit` subfolder in a subsequent PR. I think the pattern match will still work.

joeldodge79 (approved):
LGTM

PR #400, opened 2020-11-26
Python SDK: Split sdk-level constants from RTL

josephaxisa:


python/looker_sdk/rtl/auth_session.py
```diff
@@ -46,14 +46,14 @@ def __init__(
         settings: api_settings.PApiSettings,
         transport: transport.Transport,
         deserialize: serialize.TDeserialize,
-        version: str,
+        api_version: str,
```
josephaxisa, line 49: Renamed this to match `ApiMethods`

python/looker_sdk/rtl/api_settings.py
```diff
@@ -49,7 +48,13 @@ def read_config(self) -> Dict[str, str]:
 class ApiSettings(PApiSettings):
     deprecated_settings: Set[str] = {"api_version", "embed_secret", "user_id"}
 
-    def __init__(self, filename: str = _DEFAULT_INI, section: Optional[str] = None):
+    def __init__(
+        self,
+        sdk_version: str,
```
josephaxisa, line 53: Perhaps `sdk_version` should be optional. If the RTL is being used on its own (without the SDK), it might also make sense to rename it, maybe `agent_postfix` would be a more appropriate name. Thoughts?

jkaster: in Typescript, ApiMethods sets the AgentTag in its entirety and pushes that as a property down to the request processor. Maybe the same kind of thing can be used here?

joeldodge79 (changes requested):
once you merge`jax/auth-session-and-token-refactor` then you can re-target this PR to master.

Just one change to make the `sdk_version` parameter optional.

jkaster:
@joeldodge79 At the very least we want `TS-SDK` (or `PY-SDK`) as the AgentTag/x-looker-appid. Where feasible/possible, we push down the Looker version constant to the ApiMethods constructor so it becomes `TS-SDK 7.20` for example. We don't need to track the API version for the SDK, because the url being hit tells us the API version.

jkaster (approved):
LGTM, but I'm happy to let @joeldodge79 confirm

joeldodge79 (approved):
👍

PR #401, opened 2020-12-01
Moved non-file I/O spec conversion functions into sdk-codegen

jkaster:
- Minor cleanup for testUtils.ts
- Spec file I/O is still in sdk-codegen-scripts
- Fixed some miscellaneous test issues
- Avoids node infection for pending browser-based spec conversion


packages/sdk-codegen-scripts/src/fetchSpec.ts
```diff
@@ -186,27 +186,16 @@ export const login = async (props: ISDKConfigProps) => {
...
+  const accessToken = await response.access_token
 
-    if (accessToken) {
-      return accessToken
-    } else {
-      log(`Server Response: ${JSON.stringify(response)}`)
-      throw new Error('Access token could not be retrieved.')
-    }
-  } catch (err) {
-    throw err
```
joeldodge79, line 209: 😸

packages/sdk-codegen/src/specConverter.ts
```diff
@@ -0,0 +1,194 @@
+/*
```
joeldodge79, line 1: I'm assuming there's no major re-write in here, just ported from sdk-codegen-scripts?

packages/sdk-codegen/src/testUtils/testUtils.ts
```diff
@@ -59,17 +60,20 @@ export interface ITestConfig {
   testSection: IKeyAny
 }
 
+const homeToRoost = '../../../../'
+
+export const getRootPath = () => path.join(__dirname, homeToRoost)
```
joeldodge79, line 65: TIL `__{dir,file}name` - python version is `__file__` - I find the choice of leading and trailing double underscores between languages ... entertaining

joeldodge79 (approved):
LGTM

josephaxisa (approved):
👍

packages/sdk-codegen-scripts/src/fetchSpec.ts
jkaster: yeah, I think that was left over from some testing I was doing when first writing that code.

packages/sdk-codegen/src/specConverter.ts
jkaster: Right. This is prep work for full Swagger -> OpenApi on-demand conversion in the browser

PR #402, opened 2020-12-01
Remove "../looker.ini" dependence for Python tests

joeldodge79:
CI will use LOOKERSDK env vars. Currently, passing "../looker.ini" in
the tests breaks CI because there is no such file. This change adds
"../looker.ini" to the allowlist (_DEFAULT_INIS) of paths that can be
specified and not bomb out if they don't exist.

misc cleanup:
- ApiSettings constructor to take keyword args only (for readability)
- DRYed out `init31/40` settings creation
- added pass through env vars for pipenv -> tox (note: CI will not rely
    on this since it does not run tox in pipenv, this is just for
    bin/smoke usage)


python/looker_sdk/rtl/api_settings.py
```diff
@@ -72,6 +73,10 @@ def __init__(
             section (str): section in config file. If not supplied default to
                 reading first section.
         """
+        if not os.path.isfile(filename):
+            if filename and filename not in _DEFAULT_INIS:
+                raise FileNotFoundError(f"No config file found: '{filename}'")
+
```
joeldodge79, line 79: I made config file existence checking eager to simplify the config reading below

python/looker_sdk/rtl/api_settings.py
```diff
@@ -123,26 +122,25 @@ def _bool(val: str) -> bool:
             raise TypeError
         return converted
 
-    @staticmethod
```
joeldodge79, line 126: I can't remember why this was a `@staticmethod` but it doesn't need to be

python/looker_sdk/rtl/transport.py
```diff
@@ -63,7 +64,8 @@ class PTransportSettings(Protocol):
     headers: Optional[MutableMapping[str, str]]
 
     def is_configured(self) -> bool:
-        return bool(self.base_url)
+        if not self.base_url:
+            raise error.SDKError("Missing required configuration value: base_url")
```
joeldodge79, line 68: I've gone back and forth on this but my latest thinking is that raising here rather than everywhere else is cleaner

python/tests/integration/.env
```diff
@@ -1 +1,5 @@
+LOOKERSDK_CLIENT_ID=${LOOKERSDK_CLIENT_ID}
+LOOKERSDK_CLIENT_SECRET=${LOOKERSDK_CLIENT_SECRET}
+LOOKERSDK_BASE_URL=${LOOKERSDK_BASE_URL}
+LOOKERSDK_VERIFY_SSL=${LOOKERSDK_VERIFY_SSL}
```
joeldodge79, line 4: these are so that if you run `pipenv run tox` and don't have a "../looker.ini" and use LOOKERSDK env vars instead, they'll actually make it through so the code under test can read them!

jkaster (approved):
LGTM

python/looker_sdk/__init__.py
```diff
@@ -37,20 +37,22 @@
...
 
-class InitError(Exception):
-    pass
+def _settings(config_file: str, section: Optional[str] = None):
+    return api_settings.ApiSettings(
+        filename=config_file,
+        section=section,
+        sdk_version=constants.sdk_version,
+        env_prefix=constants.environment_prefix,
+    )
```
josephaxisa, line 46: So much better

josephaxisa (commented):
I know this has been merged but I just had a look anyway and it LGTM :)

PR #403, opened 2020-12-01
Getting roles for each user was very slow

bryans99:
This fix falls back to using all roles for all users. If all roles fails, thats okay as the kind of user that does not have access to all roles will not care about the roles of other users.


jkaster (approved):
I think this works. Thanks for the improvement.

PR #405, opened 2020-12-04
[Part 1 - APIX Declaration Links] - Declaration Miner

josephaxisa:


packages/sdk-codegen-scripts/src/miner.ts
```diff
@@ -74,13 +74,13 @@ export const filterAllFiles = (fileName: string) => fileName.trim().length > 0
...
  */
 export const filterCodeFiles = (fileName: string) => {
-  if (/(^|\/)node_modules|lib\//gi.test(fileName)) return false
+  if (/(^|\/)node_modules\//gi.test(fileName)) return false
   const ext = path.extname(fileName).toLocaleLowerCase()
   // eslint-disable-next-line @typescript-eslint/no-use-before-define
   return ext in fileMiners
 }
 
-const skipFolder = (name: string) => /node_modules|lib/gi.test(name)
```
jkaster, line 83: Does this mean `lib` paths are now processed for example mining?

jkaster (commented):
I have a question about how this may change example mining

packages/sdk-codegen-scripts/src/miner.ts
josephaxisa: My mistake. I just modified `skipFolder` to accept an exclude list

jkaster (approved):
LGTM

PR #407, opened 2020-12-04
Python CI workflow

joeldodge79:
Run unit and integration tests using Github Actions.

Some notes for future consideration:
* It would be really nice if GA supported yaml anchors/aliases to DRY
  out the configuration. An alternative could be to use [composite run
  steps](https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/creating-a-composite-run-steps-action)
  but keeping it simple and slightly duplicated for now.
* See TODOs about integration tests on windows and macos
* I wanted to include a "LOOKER_LATEST" version variable so that the
  windows and macos job config blocks could reference it and not have
  to be changed each time a new looker version is added but currently
  the "env" and "secrets" contexts aren't available and job level
  configuration syntax

### DIFF
--- a/.github/scripts/wait_for_looker.py
+++ b/.github/scripts/wait_for_looker.py
@@ -1,0 +1,34 @@
+import ssl
+import sys
+import time
+import urllib.request
+
+# turn off ssl checking
+ctx = ssl.create_default_context()
+ctx.check_hostname = False
+ctx.verify_mode = ssl.CERT_NONE
+
+MAX_RETRIES = 160
+ATTEMPTS = 1
+while MAX_RETRIES:
+    retry_msg = f"after {ATTEMPTS} attempts: {MAX_RETRIES} retries remaining."
+    try:
+        status = urllib.request.urlopen(
+            "https://localhost:10000/login", context=ctx
+        ).getcode()
+    except urllib.error.URLError:
+        msg = f"Looker server connection rejected {retry_msg}"
+    else:
+        if status == 200:
+            print(f"Looker ready after {ATTEMPTS} attempts.")
+            sys.exit(0)
+        else:
+            msg = f"Received status({status}) from Looker {retry_msg}"
+
+    ATTEMPTS += 1
+    MAX_RETRIES -= 1
+    time.sleep(2)
+    print(msg)
+
+print(f"Looker took too long to start {retry_msg}")
+sys.exit(1)
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -1,0 +1,181 @@
+name: Python CI
+on:
+  pull_request_target:
+    paths:
+      - python/**
+      - packages/sdk-codegen*/**
+
+  push:
+    branches:
+      - $default-branch
+    paths:
+      - python/**
+      - packages/sdk-codegen*/**
+
+env:
+  LOOKERSDK_BASE_URL: https://localhost:20000
+  LOOKERSDK_VERIFY_SSL: false
+  TOX_JUNIT_OUTPUT_DIR: results
+
+defaults:
+  run:
+    shell: bash
+    working-directory: python/
+
+jobs:
+  unit:
+    name: Unit - ${{ matrix.os }} / py${{ matrix.python-version }}
+    env:
+      TOX_JUNIT_OUTPUT_NAME: ${{ matrix.os }}.py${{ matrix.python-version }}
+    runs-on: ${{ matrix.os }}-latest
+    strategy:
+      matrix:
+        os:
+          - ubuntu
+          - macos
+          - windows
+        python-version:
+          - '3.6'
+          - '3.7'
+          - '3.8'
+          - '3.9'
+
+    steps:
+      - name: Repo Checkout
+        uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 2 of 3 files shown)


.github/workflows/python-ci.yml
```diff
@@ -0,0 +1,181 @@
+name: Python CI
+on:
+  pull_request_target:
+    paths:
+      - python/**
+      - packages/sdk-codegen*/**
```
joeldodge79, line 6: monorepo - python CI should only run for files changed in these dirs... i think?

.github/workflows/python-ci.yml
```diff
@@ -0,0 +1,181 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install Tox and any other packages
+        run: |
+          python -m pip install --upgrade pip
+          pip install tox
+      - name: Set LOOKERSDK_CLIENT_{ID,SECRET}
+        run: |
+          echo "LOOKERSDK_CLIENT_ID=${{ secrets[format('LOOKERSDK_CLIENT_ID__{0}', matrix.looker)]}}" >> $GITHUB_ENV
```
joeldodge79, line 56: Each Looker image has different ID/SECRET creds - this code retrieves them programmatically from secrets.
In order for this workflow to function we'll need someone who can see the "Settings" tab to add these secrets:
`LOOKERSDK_CLIENT_ID__7_20`
`LOOKERSDK_CLIENT_SECRET__7_20`
`LOOKERSDK_CLIENT_ID__7_18`
`LOOKERSDK_CLIENT_SECRET__7_18`
`GCP_PROJECT_ID`
`GCP_AR_READER_SA_KEY`

I'll create an internal doc describing how to get the values. Happy to do it myself if someone can give me the access

joeldodge79 (commented):
some inline notes. happy to meet and chat about this initial CI workflow

josephaxisa (commented):
This all looks perfectly fine to me but I'm not very familiar with Github actions so I'll let John give the final approval

.github/workflows/python-ci.yml
jkaster: That's a good start, I think. We may need `packages/sdk-codegen-scripts/**` as well but probably not since `sdk-codegen` is where any changes to the actual generation syntax would appear.

.github/scripts/wait_for_looker.py
```diff
@@ -0,0 +1,34 @@
...
+ctx.check_hostname = False
+ctx.verify_mode = ssl.CERT_NONE
+
+MAX_RETRIES = 160
+ATTEMPTS = 1
+while MAX_RETRIES:
+    retry_msg = f"after {ATTEMPTS} attempts: {MAX_RETRIES} retries remaining."
+    try:
+        status = urllib.request.urlopen(
+            "https://localhost:10000/login", context=ctx
```
jkaster, line 17: Is this just to see if the looker server is alive?
```suggestion
            "https://localhost:10000/alive", context=ctx
```

.github/workflows/python-ci.yml
```diff
@@ -0,0 +1,181 @@
...
+defaults:
+  run:
+    shell: bash
+    working-directory: python/
+
+jobs:
+  unit:
+    name: Unit - ${{ matrix.os }} / py${{ matrix.python-version }}
+    env:
+      TOX_JUNIT_OUTPUT_NAME: ${{ matrix.os }}.py${{ matrix.python-version }}
```
jkaster, line 29: Nice to have the JUnit output right away

.github/workflows/python-ci.yml
jkaster: Who can give you the access you're needing?

.github/workflows/python-ci.yml
```diff
@@ -0,0 +1,181 @@
...
+          # Regarding docker pull ... it would be nice if there's a way to cache
+          # only some layers of the image on the runner but we don't want to cache
+          # the final layer(s) with Looker IP. This would speed up docker pull on
+          # all OSs.
+          #- os: macos
+          #  python-version: '3.x'
+          #  looker: '7_20'
+          # TODO: currently can't run linux containers on windows.
+          # Pending new windows server version
+          # https://github.com/actions/virtual-environments/issues/1143#issuecomment-698797524
```
jkaster, line 96: Good callout

jkaster (approved):
This looks good. I did make a suggestion for the "alive?" endpoint to use. Happy to approve again if you want to take that change.

.github/workflows/python-ci.yml
joeldodge79: actually, the trailing `*/**` should capture `sdk-codegen` and `sdk-codegen-{utils,scripts}`

.github/scripts/wait_for_looker.py
joeldodge79: I'll play with that.

.github/workflows/python-ci.yml
jkaster: Doh. You're right. I didn't notice that `*` before the `/` ;)

.github/workflows/python-ci.yml
joeldodge79: all set. I've started a document that describes how these secrets are managed.

PR #412, opened 2020-12-08
Creating modules for sdk packages

jkaster:
Adds modules for Typescript SDKs. If the canary build works, I'll publish the update.


bryans99 (approved):
LGTM

Don't lose track of fixing the test :D 

I assume anyone wanting streams has to instantiate it directly

jkaster:
@bryans99 `methods.spec.ts` and `downloadTile.ts` have already been updated for the new stream sdk initialization requirements

PR #432, opened 2021-01-05
Do not auto run CI on forked PRs

joeldodge79:
I misunderstood the security boundary for the pull_request_target event.
Apparently we should not be running code from forked PRs as it can gain
access to the github token which has access to our github secrets.

https://securitylab.github.com/research/github-actions-preventing-pwn-requests

### DIFF
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -1,6 +1,6 @@
 name: Python CI
 on:
-  pull_request_target:
+  pull_request:
     types:
       - opened
       - synchronized
@@ -37,11 +37,11 @@ jobs:
     if: >
       github.event_name == 'push' ||
       (
-        github.event_name == 'pull_request_target' &&
+        github.event_name == 'pull_request' &&
         github.event.action != 'labeled'
       ) ||
       (
-        github.event_name == 'pull_request_target' &&
+        github.event_name == 'pull_request' &&
         github.event.action == 'labeled' &&
         github.event.label.name == 'CI:TEST'
       )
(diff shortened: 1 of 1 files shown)


joeldodge79:
I've disabled github actions on this repo until this PR goes through

jkaster (approved):
This looks ok to me but I don't have full context yet to completely understand the difference this makes

.github/workflows/python-ci.yml
```diff
@@ -1,6 +1,6 @@
 name: Python CI
 on:
-  pull_request_target:
+  pull_request:
```
joeldodge79, line 3: now forked repo PR CI runs will get a github access token that has read-only access and no secrets access.

this means for now that the integration tests will fail for forked PRs. We'll come up with a workflow where we can manually trigger integration test runs for external contributors.

.github/workflows/python-ci.yml
```diff
@@ -69,9 +69,6 @@ jobs:
           access_token: ${{ github.token }}
       - name: Repo Checkout
         uses: actions/checkout@v2
-        with:
-          ref: ${{ github.event.pull_request.head.ref }}
-          repository: ${{ github.event.pull_request.head.repo.full_name }}
```
joeldodge79, line 74: These settings were required to checkout forked PR's code because the default for `pull_request_target` was the base repo master HEAD. Now that we're using `pull_request` whose default is the branch's head we don't need it.

joeldodge79 (commented):
some explanatory comments for posterity

PR #433, opened 2021-01-08
Fixes issue with default arguments in golang

panoskoug:
Adds one more golang example that gets all the LookML projects.
This example would have failed with this panic before this change.
`panic: reflect: call of reflect.Value.IsNil on string Value`


kppk (approved):
lgtm :+1:

PR #435, opened 2021-01-11
Update package.jsons for security issues

jkaster:


examples/access-token-server/package.json
```diff
@@ -11,7 +11,7 @@
     "create-status-json": "ts-node script/create_status_json.ts"
   },
   "dependencies": {
-    "@looker/sdk": "^0.3.8-beta.0",
+    "@looker/sdk": "^7.20.0",
```
bryans99, line 14: It's premature to update this version. Just tested it out and it still fails

bryans99 (changes requested):
access token server still doesnt work with latest sdk

bryans99 (approved):

PR #437, opened 2021-01-12
Improve npm packaging

bryans99:
published npm packages now included commonjs and esm distributions.

further work needs to be done to expose both commonjs and esm but the packages that need to run in both node and the browser either expose both or commonjs rather than esm.


jkaster (approved):
LGTM

PR #438, opened 2021-01-13
chore: python ci/release workflow tweaks

joeldodge79:
Python CI
* remove "if: false" for integration tests to re-enable
* only trigger on files under python/
* add "Twine upload check" as part of CI so we avoid that pitfal when
  publishing
* More intelligent "Remove CI:TEST Label" conditional that will skip the
  whole job rather than start the job and skip the single step

Python Release
* trigger on release/edited so that marking a release as non-draft will be
  the condition for publishing to PyPI

Release Please
* stop using secrets.RELEASE_PLEASE_TOKEN - we no longer want to have
the release-please-action's side-effects trigger other workflows.
  * We'll have a separate solution for running CI on a Release PR prior
  to merging it
  * we're no longer publishing on "release/published".

### DIFF
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -10,14 +10,12 @@ on:
       - labeled
     paths:
       - python/**
-      - packages/sdk-codegen*/**
 
   push:
     branches:
       - master
     paths:
       - python/**
-      - packages/sdk-codegen*/**
 
 env:
   LOOKERSDK_BASE_URL: https://localhost:20000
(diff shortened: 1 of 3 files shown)


.github/workflows/python-ci.yml
```diff
@@ -33,6 +31,7 @@ jobs:
   unit:
     # 1st condition: push event (restricted to master above)
     # 2nd condition: PR opened/synchronized/reopened (types filter above)
+    #                except Release PRs, those need a manual CI:TEST
```
joeldodge79, line 34: I'll come up with an automated solution that will run CI on a Release PR just prior to merging it.

joeldodge79 (commented):
@jkaster mostly just re-enabling Python CI integration tests. the release-please and publish workflows are disabled on sdk-codegen since those are still WIP

jkaster (approved):
LGTM. Exciting we're getting so close.

PR #439, opened 2021-01-13
Add typescript declarations to esm directory

bryans99:
Provide typescript in esm directory to allow extension sdk to reach into the esm directory to reduce final code size.

Note, sdk-codegens-scripts is excluded as an error ocurred running tsc. Not an issue as the only package that really requires this is sdk.


bryans99:
Deferring this change until extension-sdk and extension-sdk-react are in the mono repo

PR #440, opened 2021-01-14
adding examples in Golang

DucarrougeR:


examples/go/get_all_dashboards.go
```diff
@@ -0,0 +1,32 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/looker-open-source/sdk-codegen/go/rtl"
+	v4 "github.com/looker-open-source/sdk-codegen/go/sdk/v4"
```
kppk, line 7: super nit: seems the alias `v4` is not needed?

kppk (dismissed):
lgtm, thanks! :+1:

examples/go/get_all_dashboards.go
DucarrougeR: Great, thanks I am updating these

DucarrougeR: Closing this comment here, I've updated all 4 files to remove the alias. Thanks again

kppk (approved):
lgtm

PR #453, opened 2021-01-30
Typescript SDK for Looker 21.0

jkaster:
Updated deployment config for the Typescript SDK to set version to 21.0


packages/sdk-codegen/package.json
```diff
@@ -9,7 +9,7 @@
   ],
   "author": "Looker",
   "license": "MIT",
-  "private": false,
+  "private": true,
```
josephaxisa, line 12: Interesting. Didn't know sdk-codegen is not meant to be published on npm

jkaster: It will be after we fix the package versioning for CD. It shouldn't follow looker release versioning like the SDKs do

packages/sdk-rtl/src/nodeServices.ts
```diff
@@ -33,6 +33,7 @@ import { NodeCryptoHash, NodeTransport } from './nodeTransport'
 export class NodeServices implements IPlatformServices {
   /** Cryptography service interface */
   crypto: ICryptoHash
+  /** SDK configuration interface */
```
bryans99, line 36: Ah crap - gonna have a merge issue with this :D

bryans99 (approved):
LGTM

PR #457, opened 2021-02-03
Fixes an issue with Golang API requests containing strings

panoskoug:
It turns out that the string was marshalled with quotes
for instance https://test.looker.com:443/api/4.0/users/search?email="foo@bar.com"
instead of https://test.looker.com:443/api/4.0/users/search?email=foo@bar.com
causing the server to ignore it.

### DIFF
--- a/go/example/main.go
+++ b/go/example/main.go
@@ -35,6 +35,23 @@ func printAllProjects(sdk *v4.LookerSDK) {
 	}
 }
 
+func printAboutMe(sdk *v4.LookerSDK) {
+
+	me,err := sdk.Me("",nil)
+	check(err)
+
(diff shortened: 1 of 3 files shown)


kppk (approved):
lgtm

Looking at other sdks:

https://github.com/looker-open-source/sdk-codegen/blob/master/kotlin/src/main/com/looker/rtl/Transport.kt#L139

https://github.com/looker-open-source/sdk-codegen/blob/0c736bca54d62a94b6861c7bf9d7c644d4809c74/csharp/rtl/SdkUtils.cs#L100

looks like we might need to handle date too.

panoskoug:
OK I will keep Date in mind as well, we might need that at some point

PR #462, opened 2021-02-05
Brynryans/lint fix copyright headers

bryans99:
Year is now 2021.

For generated ts files, comment prefix is now /* rather than jsdoc style /**. Minor change to ts generator.

Only impacts ts, tsx and js files. 

Other code generators will use 2021 when then are run next.

### DIFF
--- a/packages/sdk-codegen/src/typescript.gen.ts
+++ b/packages/sdk-codegen/src/typescript.gen.ts
@@ -143,14 +143,16 @@ export interface IDictionary<T> {
 
   commentHeader(indent: string, text: string | undefined, commentStr = ' * ') {
     if (!text) return ''
+    const commentPrefix =
+      text.includes(' License') && text.includes('Copyright (c)') ? '/*' : '/**'
     if (commentStr === ' ') {
-      return `${indent}/**\n\n${commentBlock(
+      return `${indent}${commentPrefix}\n\n${commentBlock(
         text,
         indent,
         commentStr
       )}\n\n${indent} */\n`
     }
-    return `${indent}/**\n${commentBlock(
+    return `${indent}${commentPrefix}\n${commentBlock(
       text,
       indent,
       commentStr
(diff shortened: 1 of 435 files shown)


packages/sdk-node/src/index.ts
```diff
@@ -25,7 +25,7 @@
  */
 
 // Version 21.0.4
- export * from './nodeServices'
+export * from './nodeServices'
```
bryans99, line 28: This was reported by John from a previous PR and I said I would fix in the next PR I put up.

packages/sdk-rtl/src/index.ts
```diff
@@ -24,7 +24,7 @@
 
  */
 
- // Version 21.0.4
+// Version 21.0.4
```
bryans99, line 27: Another unrelated fix but similar to the one above

packages/wholly-sheet/package.json
```diff
@@ -37,6 +37,7 @@
     "uuid": "^8.3.1"
   },
   "devDependencies": {
+    "@looker/sdk-node": "^21.0.4",
```
bryans99, line 40: This one needed to be fixed as well. Needed by a test that still worked (I guess through the magic of lerna) but I felt the eslint error needed to be fixed sooner rather than later

bryans99:
Note the three unrelated changes that slipped through (there were actually others but I backed those out). The above I felt warranted inclusion. All other changes are header related in the following categories
1. Change of year <- vast majority of changes this
2. Header missing
3. /** -> /*

packages/sdk-codegen/src/typescript.gen.ts
```diff
@@ -143,14 +143,16 @@ export interface IDictionary<T> {
 
   commentHeader(indent: string, text: string | undefined, commentStr = ' * ') {
     if (!text) return ''
+    const commentPrefix =
+      text.includes(' License') && text.includes('Copyright (c)') ? '/*' : '/**'
```
jkaster, line 147: nice little tweak

jkaster (approved):
LGTM.

Next year I get to do this so I can have an even bigger PR.

PR #463, opened 2021-02-06
Eslint fixes

bryans99:
This fixes some but not all eslint issues. In some cases eslint disable has been used. Reasons any one of the following
1. lint rule does not make sense for code. example prefer-required, prefer to have class, prefer in document and import no extraneous dependencies (this one is weirs suspect eslint and lerna at cross purposes)
2. fix breaks test - example DocMarkdown.spec.ts
3. fix change to impactful and no really good test (api-explorer/scripts/utils.ts for example)

Note one fix removed a duplicate case statement (data/projects/reducer.ts). Probably not an issue but could be the wrong action or could be that the value should be cleared. Will wait for a bug to be reported.

Once this is merged I'm going to turn on lint:es as part of the CI build

### DIFF
--- a/packages/api-explorer/scripts/utils.ts
+++ b/packages/api-explorer/scripts/utils.ts
@@ -118,7 +118,9 @@ const brokenPromise = (message: string) => Promise.reject(new Error(message))
 export const registerApp = async () => {
   const args = process.argv.slice(2)
   const total = args.length
+  // eslint-disable-next-line node/no-path-concat
   const iniFile = total < 1 ? `${__dirname}/../../../looker.ini` : args[0]
+  // eslint-disable-next-line node/no-path-concat
   const configFile = total < 2 ? `${__dirname}/appconfig.json` : args[1]
   let result = ''
   console.log(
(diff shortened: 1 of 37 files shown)


bryans99:
Actually, decided to turn on lint now. Will back on request though

bryans99:
CI:TEST

packages/api-explorer/scripts/utils.ts
```diff
@@ -118,7 +118,9 @@ const brokenPromise = (message: string) => Promise.reject(new Error(message))
 export const registerApp = async () => {
   const args = process.argv.slice(2)
   const total = args.length
+  // eslint-disable-next-line node/no-path-concat
   const iniFile = total < 1 ? `${__dirname}/../../../looker.ini` : args[0]
+  // eslint-disable-next-line node/no-path-concat
   const configFile = total < 2 ? `${__dirname}/appconfig.json` : args[1]
```
jkaster, line 124: I can update these to path.join() in a future PR

jkaster (approved):
LGTM. Nice cleanup!

PR #466, opened 2021-02-09
Fixed dev environment

josephaxisa:
Corrected babel's out dir when running in watch mode. Prior to this, webpack was not recompiling when for example APIX's dev server was running and a change was made in one of APIX's dependencies, such as RunIt.


jkaster (approved):
LGTM

PR #469, opened 2021-02-10
yarn watch changes

bryans99:
There are now two watch commands
yarn watch - watches for changes and builds for browser based development
yarn watch:cjs - watches for changes and builds for node based development

In addition, updated extension-sdk-react package for extensions (fixes an issue with browser backward/forward buttons)


jkaster (approved):
LGTM

PR #472, opened 2021-02-11
Fixed Python makeTheCall dictionary syntax

josephaxisa:
### The Problem

Currently the Python generator defines dictionaries using the `dict()` function, e.g. `dict(k1 = "v1", k2 = "v2")`. 

This generates incorrect syntax when the key contains periods or other special characters like hyphens, as it requires it to be quoted, e.g. `dict("foo.bar" = "v1")`. Additionally, passing keyword arguments into `dict()` is risky if the keys are not valid Python identifiers, e.g. `import`. 

### Solution
Initialize dictionaries using curly braces as it works for all kinds of keys. 
- Added `hashKeyQuote` and `hashSetSep` attributes to `ICodeGen`
- Removed overriding `argSet` method in PythonGen

Note that currently we do not capture the key types and all keys are quoted. This is a bug that is unlikely to be experienced but might need fixing later on. See [here](https://github.com/looker-open-source/sdk-codegen/blob/e00f8f7d2d6d138232ed21be231788c7898b9184/packages/sdk-codegen/src/python.gen.spec.ts#L1028-L1032).

### DIFF
--- a/packages/sdk-codegen/src/codeGen.ts
+++ b/packages/sdk-codegen/src/codeGen.ts
@@ -167,8 +167,14 @@ export interface ICodeGen {
   /** argument separator string. Typically ', ' */
   argDelimiter: string
 
-  /** type properties/args expression separater. E.g ': ' for Typescript */
+  /** type properties/args expression separator. E.g ': ' for Typescript */
   argSetSep: string
+
+  /** hash type properties/args expression separator. E.g ': ' for Typescript */
+  hashSetSep: string
+
+  /** hash key quotes E.g '' for Typescript and '"' for Python */
+  hashKeyQuote: string
 
   /** parameter delimiter. Typically ",\n" */
   paramDelimiter: string
@@ -642,6 +648,7 @@ export abstract class CodeGen implements ICodeGen {
   fileExtension = '.code'
   argDelimiter = ', '
   argSetSep = ': '
+  hashSetSep = ': '
   paramDelimiter = ',\n'
   propDelimiter = '\n'
   enumDelimiter = ',\n'
@@ -652,6 +659,7 @@ export abstract class CodeGen implements ICodeGen {
   arrayClose = ']'
   hashOpen = '{'
   hashClose = '}'
+  hashKeyQuote = ''
   typeOpen = '{'
   typeClose = '}'
   useModelClassForTypes = false
@@ -934,7 +942,8 @@ export abstract class CodeGen implements ICodeGen {
     const bump = this.bumper(indent)
     Object.entries(val).forEach(([name, v]) => {
       const exp = this.anyValue(bump, v)
-      args.push(this.argSet(name, this.argSetSep, exp))
+      const key = `${this.hashKeyQuote}${name}${this.hashKeyQuote}`
+      args.push(this.argSet(key, this.hashSetSep, exp))
     })
     return this.argIndent(indent, args, this.hashOpen, this.hashClose)
   }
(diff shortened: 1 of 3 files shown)


packages/sdk-codegen/src/python.gen.spec.ts
```diff
@@ -1025,33 +1025,43 @@ class MergeFields(model.Model):
         const oneItem = [1]
         const threeItems = ['Abe', 'Zeb', token]
         const inputs: IDictionary<any> = {
+          /**
+           * The below key is quoted in the generated dictionary. This is a bug
+           * that might need addressing later on.
+           */
```
jkaster, line 1031: This can wait for another PR, but can you explain what's going on here?

jkaster (dismissed):
Good for this PR. Just one question for follow-up if it's not a quick fix.

packages/sdk-codegen/src/python.gen.spec.ts
josephaxisa: In Python it is possible to have an integer as a dictionary key. Consider
```python
x = { 1: "one", "2": "two" }
```
These two values would be accessed with `x[1]` and `x["2"]` respectively in Python. However, in the Python code generator we are quoting all dictionary keys.

joeldodge79:
hrm, looks like we're missing CI for the paths touched in this PR. I'm guessing a separate CI file for now since the tssdk-ci only runs sdk tests? https://github.com/looker-open-source/sdk-codegen/blob/3c74927b0cd9409516ec027536dfeadfcfb65dfc/.github/workflows/tssdk-ci.yml#L93

packages/sdk-codegen/src/codeGen.ts
```diff
@@ -652,6 +659,7 @@ export abstract class CodeGen implements ICodeGen {
   arrayClose = ']'
   hashOpen = '{'
   hashClose = '}'
+  hashKeyQuote = ''
```
joeldodge79, line 662: is this mean to be an empty string? or did you mean `'"'` or conversely `"'"`?

packages/sdk-codegen/src/python.gen.ts
```diff
@@ -47,15 +47,17 @@ export class PythonGen extends CodeGen {
   indentStr = '    '
   argDelimiter = `,\n${this.indentStr.repeat(3)}`
   argSetSep = '='
+  hashSetSep = ': '
```
joeldodge79, line 50: same value as parent class

```suggestion
```

packages/sdk-codegen/src/python.gen.ts
```diff
@@ -47,15 +47,17 @@ export class PythonGen extends CodeGen {
...
   propDelimiter = '\n'
   dataStructureDelimiter = ', '
   enumDelimiter = '\n'
   codeQuote = '"'
   typeOpen = '('
   typeClose = ')'
-  hashOpen = 'dict('
-  hashClose = ')'
+  hashOpen = '{'
+  hashClose = '}'
```
joeldodge79, line 59: ditto
```suggestion
```

joeldodge79 (changes requested):
nice work @josephaxisa ! couple of small nits

packages/sdk-codegen/src/codeGen.ts
joeldodge79: oh, nm - javascript objects aren't quoted

jkaster: Should be empty string in the base class

jkaster (approved):
Looks good

joeldodge79 (approved):
👍

PR #473, opened 2021-02-11
Update examples, READMEs and CHANGELOGs to reflect new typescript @lo…

bryans99:
…oker/sdk-node package


jkaster (dismissed):
Thanks for the updates!

jkaster (approved):
Lerna updated package dependencies

PR #475, opened 2021-02-17
Minor improvements to some SDK readmes

jkaster:
Updated:
- Typescript Browser SDK readme
- Typescript Node SDK readme
- Kotlin SDK readme
- Swift SDK readme

### DIFF
--- a/packages/sdk-node/README.md
+++ b/packages/sdk-node/README.md
@@ -1,7 +1,310 @@
-# Looker SDK Library for Node
+# Looker SDK for Node
 
-You will need this package if you want to run the Looker SDK in a node environment. This package is NOT required if you want to the run the Looker SDK in a browser environment.
+The Looker SDK for Typescript/Javascript works with Node and browser run-times. The SDK provides a convenient way to communicate with a Looker server's APIs.
 
-The `@looker/sdk` package is updated with every Looker release. This package has a much longer update/release cycle.
+This package is specifically for using the Typescript SDK with Node.
 
-**DISCLAIMER**: This is a _beta_ version of the Looker SDK library for Node. Implementations are still subject to change.
+The SDK uses a plug-in architecture (also known as dependency injection) for initializing that supports run-time specific transports (like `NodeTransport` and `BrowserTransport`) and different approaches for managing API authentication (like `NodeSession`, `BrowserSession`, `ProxySession`, and `CorsSession`).
+
+**DISCLAIMER**: This is a _beta_ version of the Looker SDK. Implementations are still subject to change, but we expect most SDK method calls to work correctly. If you run into problems with the SDK, please feel free to [report an issue](https://github.com/looker-open-source/sdk-codegen/issues), and please indicate which language SDK you're using in the report.
+
+## Getting started
+
+The Looker SDK can be used in a node application in 3 steps:
+
+- install
+- configure
+- use
+
+### Install the Looker SDK into your node application
+
+Using `npm`:
+
+```bash
+npm install @looker/sdk @looker/sdk-node
+```
+
+Using `yarn`:
+
+```bash
+yarn add @looker/sdk @looker/sdk-node
+```
+
+**Note**: If you are only intending to use the SDK in a browser, omit `@looker/sdk-node`.
+
+### Configure the SDK for your Looker server
+
+**Note**: The `.ini` configuration for the Looker SDK is a sample implementation intended to speed up the initial development of Node applications using the Looker API. [Environment variables](#environment-variable-configuration) can also be used to configure the SDK.
+
+Create a `looker.ini` file with your server URL and API credentials assigned as shown in this example.
+
+```ini
+[Looker]
+# Base URL for API. Do not include /api/* in the url
+base_url=https://<your-looker-server>:19999
+# API 3 client id
+client_id=your_API3_client_id
+# API 3 client secret
+client_secret=your_API3_client_secret
+```
+
+**Note**: If the application using the Looker SDK is going to be committed to a version control system, be sure to
+**ignore** the `looker.ini` file to avoid publishing your API credentials.
+
+### Use the SDK in your code
+
+When the SDK is installed and the server location and API credentials are configured in your `looker.ini` file, it's ready to be used.
+
+Verify authentication works and that API calls will succeed with code similar to the following:
+
+```typescript
+import { LookerNodeSDK } from '@looker/sdk-node'
+(async () => {
+  // create a Node SDK object for API 3.1
+  const sdk = LookerNodeSDK.init31()
+  // retrieve your user account to verify correct credentials
+  const me = await sdk.ok(sdk.me(
+    "id, first_name, last_name, display_name, email, personal_space_id, home_space_id, group_ids, role_ids"))
+  console.log({me})
+  // make any other calls to the Looker SDK
+  const dashboards = await sdk.ok(
+    sdk.search_dashboards({title: 'My SDK dashboard'})
+  )
+  if (dashboards.length === 0) {
+    console.log('Dashboard not found')
+  }
+  const [ dashboard ] = dashboards
+  // do stuff with dashboard
+
+  ...
+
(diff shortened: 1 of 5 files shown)


packages/sdk-node/README.md
```diff
@@ -1,7 +1,256 @@
...
+
+const me40 = await sdk.ok(sdk.me())
+const me31 = await sdk.ok(sdk31.me()) // or sdk31.ok(sdk31.me())
+```
+
+## Using NodeSession for automatic authentication
+
+Almost all requests to Looker's API require an access token. Typically, this token is established when the `login` endpoint is called with correct API3 credentials for `client_id` and `client_secret`. When `login` is successful, the provided API3 credentials are used to look up the active **API user**.
+
+The `settings` provided to the `NodeSession` class include the base URL for the Looker instance, and the desired API version. When API requests are made, if the auth session is not yet established, `NodeSession` will automatically authenticate the **API User**. sed as that user.
```
bryans99, line 124: >>  **API User**. sed as that user.

sed?

packages/sdk/README.md
```diff
@@ -2,232 +2,65 @@
...
 
-**DISCLAIMER**: This is a _beta_ version of the Looker SDK, using a completely new code generator developed by Looker. Implementations are still subject to change, but we expect most SDK method calls to work correctly. If you run into problems with the SDK, please feel free to [report an issue](https://github.com/looker-open-source/sdk-codegen/issues), and please indicate which language SDK you're using in the report.
+**DISCLAIMER**: This is a _beta_ version of the Looker SDK. Implementations are still subject to change, but SDK method calls are expected to work correctly. Please [report any issues](https://github.com/looker-open-source/sdk-codegen/issues) encountered, and indicate the SDK language in the report.
 
 ## Getting started
 
-The Looker SDK can be used in a node application in 3 steps:
+These installation notes presume a `package.json` file is used to configure Typescript/Javascript package dependencies. The three steps required are:
```
bryans99, line 13: I don't find this sentence and three steps below clear. I thinks it's the reference to package.json. I understand the install but authenticate and use is not so clear 

Perhaps instead:

Three steps are required to configure Typescript/Javascript package dependencies:

packages/sdk/README.md
```diff
@@ -2,232 +2,65 @@
...
-  ...
+Authenticating for the browser takes a bit more setup than authenticating for use with Node.
 
-  await sdk.authSession.logout()
-  if (!sdk.authSession.isAuthenticated()) {
-    console.log('Logout successful')
-  }
-})()
-```
+The stand-alone version of the Looker API Explorer (included in this repository) uses OAuth and the `BrowserSDK` to get an authentication token for Looker API requests.
```
bryans99, line 41: Can we provide an outline of what needs to be done

1. register before starting
2. get the 2 urls
3. auth with redirect back
4. hopefully logged in

bryans99 (commented):
Found what I believe to be a typo.

There were a couple of other things I found confusing. Perhaps a little rewording?

packages/sdk-node/README.md
jkaster: clearly a bad edit. It was "used" but that whole fragment after `sed` was supposed to be deleted.

packages/sdk/README.md
jkaster: That's described in `cors.md` but probably that doc could use an update and this section could duplicate it.

packages/sdk-node/README.md
```diff
@@ -1,7 +1,256 @@
...
+yarn add @looker/sdk @looker/sdk-node
+```
+
+Using `npm`:
+
+```bash
+npm install @looker/sdk @looker/sdk-node
+```
+
+**Note**: If you are only intending to use the SDK in a browser, omit `@looker/sdk-node`.
```
bryans99, line 33: Is this going to cause confusion of the sdk and sdk-rtl are specified explicitly above? 

My feeling is that we should explicitly mention sdk and sdk-rtl. I know they are implicitly pulled in the linting rules for our repo indicate an error if an import references a package that is not listed in the dependencies.

packages/sdk-node/README.md
```diff
@@ -1,7 +1,256 @@
...
+
+### Configure the SDK for your Looker server
+
+**Note**: The `.ini` configuration for the Looker SDK is a sample implementation intended to speed up the initial development of Node applications using the Looker API. [Environment variables](#environment-variable-configuration) can also be used to configure the SDK.
```
bryans99, line 37: Do we want to explicitly call out that ini and environment variables should NEVER be used in browser applications. Perhaps we already do. Maybe a quick mention here

*** Note ***: the looker.ini and environment variables WILL NOT work with the Browser SDK!

packages/sdk-node/README.md
```diff
@@ -1,7 +1,256 @@
...
+```typescript
+// Ignore any SDK environment variables for the node runtime
+const settings = new NodeSettingsIniFile('')
+const sdk = LookerNodeSDK.init40(settings)
+const sdk31 = LookerNodeSDK.init31(settings)
+```
+
+### Developing with multiple API versions
+
+Starting with Looker release 7.2, the experimental version of API 4.0 is available. To support iterative migration to API 4.0 from API 3.1, the single Looker SDK package now supports multiple API versions for the generated SDK classes. Both API 3.1 and API 4.0 are supported for Node and browser-based use.
```
bryans99, line 99: > Node and browser-based

Case consistency?

packages/sdk/README.md
```diff
@@ -2,232 +2,65 @@
...
-```bash
-npm install @looker/sdk @looker/sdk-node
-```
+### Install the Looker SDK into your application
 
 Using `yarn`:
 
 ```bash
-yarn add @looker/sdk @looker/sdk-node
+yarn add @looker/sdk
```
bryans99, line 24: see previous comment about explicitly mentioning sdk-rtl. Mainly because of linting rules

packages/sdk/README.md
```diff
@@ -2,232 +2,65 @@
...
-const sdk = LookerNodeSDK.init40(settings)
-const sdk31 = LookerNodeSDK.init31(settings)
-```
+- [RunItSDK tests](/packages/run-it/src/utils/RunItSDK.spec.ts) support debugging the flow of `RunItSDK`.
 
-### Using the Browser SDK
+- The [OAuthScene](/packages/run-it/src/scenes/OAuthScene/OAuthScene.tsx) React component receives the OAuth response from the Looker server and logs the user in to retrieve the API authentication token.
 
-The default auth/auth mode for the Browser SDK uses CORS, so your Looker instance will need to be configured for CORS support, which is available in Looker 7.10 and above.
+Looker's OAuth support makes it possible to build a Looker SDK application that only requires the browser. If a browser application can use a proxy server instead, or already uses an existing backend server, it may be simpler to [use a proxy for authentication](#using-a-proxy-for-authentication)/
```
bryans99, line 49: Explicitly call out that looker.ini and environment variables will NEVER work in the browser

bryans99 (approved):
A few minor things. Up to you if you want to fix them

packages/sdk-node/README.md
jkaster: Updated

packages/sdk/README.md
jkaster: added to both readme installation instructions

packages/sdk/README.md
jkaster: ok

packages/sdk-node/README.md
jkaster: I mentioned that in the Browser SDK readme

packages/sdk-node/README.md
jkaster: I made the sdk-rtl imports explicit

packages/sdk-node/README.md
jkaster: And added another note here

bryans99 (approved):
LGTM

PR #485, opened 2021-02-27
Dependabot fix for immer

bryans99:
Deleted @spotlight/json and @spotlight/spectral packages, the latter of which was using an older version immer.
immer is no longer referenced

Deleted deleted code that referenced @spotlight/json package


jkaster (approved):
Nice! Thanks for this cleanup.

PR #488, opened 2021-03-01
refactor: remove sdk-codegen's dep on sdk

joeldodge79:
Trying to simplify the dependency graph just a tad. Currently sdk is not
needed by sdk-codegen but if we do need it in the future I'd suggest
we copy in by hand the required interfaces/values etc

### DIFF
--- a/packages/sdk-codegen/package.json
+++ b/packages/sdk-codegen/package.json
@@ -32,7 +32,6 @@
     "watch:cjs": "yarn lerna exec --scope @looker/sdk-codegen --stream 'BABEL_ENV=build_cjs babel src --root-mode upward --out-dir lib --source-maps --extensions .ts,.tsx --no-comments --watch'"
   },
   "dependencies": {
-    "@looker/sdk": "^21.0.7",
     "@looker/sdk-codegen-utils": "^21.0.7",
     "@looker/sdk-node": "^21.0.7",
     "@looker/sdk-rtl": "^21.0.7",
--- a/packages/sdk-codegen/src/python.gen.spec.ts
+++ b/packages/sdk-codegen/src/python.gen.spec.ts
@@ -24,17 +24,6 @@
 
  */
 
-import {
-  IAccessToken,
-  IDictionary,
-  IMergeQuerySourceQuery,
-  IRequestAllUsers,
-  IRequestCreateQueryTask,
-  IWriteLookWithQuery,
-  IWriteMergeQuery,
-  ISqlQueryCreate,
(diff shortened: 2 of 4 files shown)


jkaster (approved):
LGTM

joeldodge79:
@kaster I also have a question about [`specConverter.{getLookerSpecs,loadSpecs}`](https://github.com/looker-open-source/sdk-codegen/blob/b1e2b59ab52ea66fb8947bd606aa8572b471ce00/packages/sdk-codegen/src/specConverter.ts#L513) - do those exist solely for the sake of [testing](https://github.com/looker-open-source/sdk-codegen/blob/b1e2b59ab52ea66fb8947bd606aa8572b471ce00/packages/sdk-codegen/src/specConverter.spec.ts#L439) or are they used somewhere else I can't find? - I'm just wondering if there's a simple way of removing `sdk-codegen`'s dependency on `sdk-node` as well?

I'll merge this in now and open another PR for that if it makes sense at all.

jkaster:
API Explorer needs `upgradeSpec` and `upgradeSpecObject`. sdk-codegen should have no run-time dependencies on `sdk-node`. It may have dev dependencies because of tests? We can look into that.

joeldodge79:
yeah, it looks like there are two methods on `specConverter` ([`specConverter.{getLookerSpecs,loadSpecs}`](https://github.com/looker-open-source/sdk-codegen/blob/b1e2b59ab52ea66fb8947bd606aa8572b471ce00/packages/sdk-codegen/src/specConverter.ts#L513)) that use an `sdk` instance to reach out on the network and grab a spec but seem to only be used by tests: [describe('spec retrieval'...](https://github.com/looker-open-source/sdk-codegen/blob/b1e2b59ab52ea66fb8947bd606aa8572b471ce00/packages/sdk-codegen/src/specConverter.spec.ts#L439) The methods themselves don't rely on `sdk-node` - they take any `APIMethods` instance. The tests use a `LookerNodeSDK.init40` instance but those methods could be called with a `LookerBrowserSDK` instance as well.

So, if `getLookerSpecs` and `loadSpecs` are meant to be part of the `specConverter`'s public API then it makes sense to have `sdk-node` as a `devDependencies` entry (currently it's a `dependencies` entry). But, given the scope of `sdk-codegen`, does it even make sense to have those "get a spec from across the network" methods as part of the public API?

jkaster:
We can look at network activities for sdk-codegen to reduce the node dependency, but the network fetch will still need to exist.

PR #489, opened 2021-03-03
API Explorer: Move search to side nav and optimize

josephaxisa:
- Replaced `Combobox` and `ComboboxInput` with `InputSearch`. The performance difference from not rendering the results inside a drop down is significantly noticeable.
- Tweaked search UI to align with design spec

TODOs that will be addressed in other PRs:
- Fix highlighting
- Tests


jkaster (approved):
Moving right along ...

PR #496, opened 2021-03-08
API Explorer updates

jkaster:
- Styles updated for Google branding
- Extension requires Looker 21.3 and above
- Extension supports API spec fetching
- Search is in the sidebar
- Methods and Types tabs show total count
- Diff prototype is available via the delta symbol on the upper right


jkaster:
There are unfortunately bogus lint errors in here. This still needs review. Thanks!

bryans99 (approved):
Approved

PR #500, opened 2021-03-11
allow id_token as response_type

aissaFoxtel:
allow id_token as response_type as the below example:

https://developer.okta.com/docs/guides/implement-implicit/use-flow/

GCP API gateway works only with id_token


bryans99:
Closed in favor of

https://github.com/looker-open-source/sdk-codegen/pull/847

PR #501, opened 2021-03-11
chore: configure semantic PRs to only care about the PR title

joeldodge79:
Since we always squash/merge, it is really the PR title we are about more
than individual commit titles. On that note, we should consider
only allowing squash/merge on sdk-codegen

### DIFF
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -1,0 +1,5 @@
+# Config file for Semantic Pull Requests Probot
+# See https://github.com/probot/semantic-pull-requests
+
+# Always validate the PR title, and ignore the commits
+titleOnly: true


jkaster (approved):
Sounds good to me

.github/semantic.yml
```diff
@@ -0,0 +1,5 @@
+# Config file for Semantic Pull Requests Probot
+# See https://github.com/probot/semantic-pull-requests
+
+# Always validate the PR title, and ignore the commits
```
bryans99, line 4: ignore the commit message body?

joeldodge79: I think it will take a little playing with to figure out the best configuration: this is what components uses so i figured we start there.

Because we squash merge, the PR title is the most important part since the changelog entry is generated only from that one line pr/commit title.

The conventional commit spec allows for footers in the body but I think the only footer we'd care about is `BREAKING CHANGES` and for that we should use the `!` convention. E.g. `feat!: removed legacy public property on foo`.

PR #502, opened 2021-03-12
fix: firefox chatty serialization error

bryans99:
Firefox does not allow Error objects to be serialized over chatty message channel. This fix upgrades chatty to pick up a fix whereby chatty serializes error messages. It also fixes extension sdk handling of uncaught errors to only send the error message rather than the error object.

### DIFF
--- a/packages/extension-sdk/src/connect/extension_host_api.spec.ts
+++ b/packages/extension-sdk/src/connect/extension_host_api.spec.ts
@@ -379,7 +379,10 @@ describe('extension_host_api tests', () => {
     const hostApi = createHostApi()
     hostApi.error(new ErrorEvent('fake_error', errorDetail))
     expect(sendSpy).toHaveBeenCalledWith('EXTENSION_API_REQUEST', {
-      payload: errorDetail,
+      payload: {
+        ...errorDetail,
+        error: errorDetail.error.toString(),
+      },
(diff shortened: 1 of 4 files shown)


packages/extension-sdk/src/connect/extension_host_api.spec.ts
```diff
@@ -379,7 +379,10 @@ describe('extension_host_api tests', () => {
     const hostApi = createHostApi()
     hostApi.error(new ErrorEvent('fake_error', errorDetail))
     expect(sendSpy).toHaveBeenCalledWith('EXTENSION_API_REQUEST', {
-      payload: errorDetail,
+      payload: {
+        ...errorDetail,
+        error: errorDetail.error.toString(),
+      },
```
jkaster, line 385: Similar to what we need to do for the TS SDK -- provide more context around the error.

jkaster (approved):
LGTM

packages/extension-sdk/src/connect/extension_host_api.spec.ts
bryans99: Reports back to mama about any uncaught errors in the extension. Unfortunately FF does not like to send the actual Error object over the MessageChannel : ( . Chrome is fine with it

PR #514, opened 2021-03-22
feat: Added --versions option to the code generator

jkaster:
This change resulted in small but low-level changes that had a major ripple effect through the Typescript codebase

The `sdk-codegen-*` packages are now also marked as deprecated.


joeldodge79 (approved):

packages/extension-api-explorer/src/ExtensionApiExplorer.tsx
```diff
@@ -25,19 +25,21 @@
...
+import {
+  ApiModel,
+  getSpecsFromVersions,
+  SpecItem,
+  SpecList,
+  upgradeSpecObject,
+} from '@looker/sdk-codegen'
 import { Looker31SDK, Looker40SDK } from '@looker/sdk'
-import { getSpecsFromVersions } from '@looker/api-explorer/src/reducers'
+import ApiExplorer from '@looker/api-explorer/lib/ApiExplorer'
```
bryans99, line 42: NOTE TO SELF: Need to figure out why we have to reach in to get ApiExplorer. Unrelated to this change.

packages/sdk-codegen-utils/README.md
```diff
@@ -2,6 +2,11 @@
 
 This package contains utility routines for the SDK code generator projects
 
-The code in this package is really trivial and probably not pulling in as a dependency to anything.
+The code in this package is really trivial and should not be used directly in your own projects.
+
+This package has no node dependencies.
+
+**DEPRECATED AND NOT SUPPORTED**: Looker does not support direct use of this package, other than via direct use in the
```
bryans99, line 9: other than via direct use in the
should be?
other than via indirect use in the

packages/sdk-codegen/README.md
```diff
@@ -2,6 +2,11 @@
 
 This package contains the OpenAPI analysis files, and all source code necessary to generate method and type declarations for supported languages.
 
+This package can be used in a browser because it has no node dependencies.
+
+**DEPRECATED AND NOT SUPPORTED**: Looker does not support direct use of this package, other than via direct use in the
```
bryans99, line 7: via indirect use?

bryans99 (approved):
LGTM

Possibly a couple of typos in README.

packages/sdk-codegen/README.md
jkaster: Fixed in other PR. Thanks!

packages/sdk-codegen/src/specConverter.ts
```diff
@@ -26,7 +26,8 @@
 
 import { isEmpty } from 'lodash'
 import { APIMethods } from '@looker/sdk-rtl'
-import { ApiModel, ArgValues, IApiModel } from './sdkModels'
+import { IApiVersion, IApiVersionElement } from '@looker/sdk'
```
joeldodge79, line 29: seems a little counter-direction to be importing parts of our looker specific API spec here, no? Is there a generic/standard version/schema for this defined somewhere? If not, I'd suggest we define our own local copy here even if it's an exact replica of the Looker interfaces - that way this package isn't relying on `@looker/sdk`

packages/sdk-codegen/package.json
```diff
@@ -33,9 +33,10 @@
     "watch": "yarn lerna exec --scope @looker/sdk-codegen --stream 'BABEL_ENV=build babel src --root-mode upward --out-dir lib --source-maps --extensions .ts,.tsx --no-comments --watch'"
   },
   "dependencies": {
-    "@looker/sdk-codegen-utils": "^21.0.9",
-    "@looker/sdk-node": "^21.0.9",
-    "@looker/sdk-rtl": "^21.0.9",
+    "@looker/sdk": "^21.0.10",
```
joeldodge79, line 36: see my other comment about hopefully dropping this dependency

packages/sdk-codegen/package.json
```diff
@@ -33,9 +33,10 @@
...
+    "@looker/sdk-codegen-utils": "^21.0.10",
+    "@looker/sdk-node": "^21.0.10",
```
joeldodge79, line 38: we should move this to `devDependencies`

joeldodge79 (commented):
just some observations about small things we may want to clean up at some point

packages/sdk-codegen/package.json
jkaster: Yup. Let's put up a new PR for that. Drop `@looker/sdk` and remove it from sdk-codegen imports and see where we're at!

packages/sdk-codegen/src/specConverter.ts
jkaster: Yeah, I think I actually did that duplication in another source file. Reliance on `@looker/sdk-rtl` is fine, but it shouldn't rely at all on `@looker/sdk`. Then we can document that "versions" payload as something people could use to generate their own specs.

packages/sdk-codegen/package.json
jkaster: Remove `@looker/sdk` and `@looker/sdk-node`. We should only need `@looker/sdk-rtl` hopefully. The `sdk-codegen-utils` package is so incredibly trivial (one small source file) we could just duplicate it for `sdk-codegen` and `sdk-codegen-scripts`.

jkaster: I captured this in #533

PR #526, opened 2021-03-24
feat: adds scrolling to frame elements

noahamac:
https://looker.atlassian.net/browse/EMBED-1481

https://drive.google.com/file/d/1n7tL5mVMWUqqjSdkM8tt-pKJvHE24BeQ/view?usp=sharing&resourcekey=0-uKp3TXkcVWDaMBPILkHKqA

This change adds independent scrolling to the various windows within the extension. The problem with fixing side scrolling as assigned is that the rest of the content still hangs down the page. This change makes sure the windows do not exceed 100% page height and scroll as expected

### DIFF
--- a/packages/api-explorer/src/ApiExplorer.tsx
+++ b/packages/api-explorer/src/ApiExplorer.tsx
@@ -98,7 +98,7 @@ const ApiExplorer: FC<ApiExplorerProps> = ({
               specDispatch={specDispatch}
               toggleNavigation={toggleNavigation}
             />
-            <Layout hasAside>
+            <Layout hasAside height={'100%'}>
               {hasNavigation && (
                 <AsideBorder pt="large" width="20rem">
                   <SideNav api={spec.api} specKey={spec.key} />
--- a/packages/api-explorer/src/components/SideNav/SideNav.tsx
+++ b/packages/api-explorer/src/components/SideNav/SideNav.tsx
@@ -118,7 +118,7 @@ export const SideNav: FC<SideNavProps> = ({ api, specKey }) => {
         <Tab>Methods ({methodCount})</Tab>
         <Tab>Types ({typeCount})</Tab>
       </TabList>
-      <TabPanels {...tabs} pt="xsmall" height={'90%'} overflowY={'scroll'}>
+      <TabPanels {...tabs} pt="xsmall" height={'90%'} overflow={'auto'}>
         <TabPanel>
           <SideNavTags tags={tags} specKey={specKey} />
         </TabPanel>
--- a/packages/api-explorer/src/components/SideNav/SideNav.tsx
+++ b/packages/api-explorer/src/components/SideNav/SideNav.tsx
@@ -118,7 +118,7 @@ export const SideNav: FC<SideNavProps> = ({ api, specKey }) => {
         <Tab>Methods ({methodCount})</Tab>
         <Tab>Types ({typeCount})</Tab>
       </TabList>
-      <TabPanels {...tabs} pt="xsmall">
+      <TabPanels {...tabs} pt="xsmall" height={'90%'} overflowY={'scroll'}>
         <TabPanel>
           <SideNavTags tags={tags} specKey={specKey} />
         </TabPanel>
--- a/packages/api-explorer/src/scenes/MethodScene/MethodScene.tsx
+++ b/packages/api-explorer/src/scenes/MethodScene/MethodScene.tsx
@@ -65,10 +65,12 @@ export const MethodScene: FC<DocMethodProps> = ({ api }) => {
   const { methodName, specKey } = useParams<DocMethodParams>()
   const { value, toggle } = useToggle()
   const [method, setMethod] = useState(api.methods[methodName])
+  const [scrollPx, setScrollPx] = useState(0)
   const seeTypes = typeRefs(api, method.customTypes)
 
   useEffect(() => {
     setMethod(api.methods[methodName])
+    setScrollPx(0)
   }, [api, methodName])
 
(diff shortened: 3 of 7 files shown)


packages/api-explorer/src/scenes/MethodScene/MethodScene.tsx
```diff
@@ -84,9 +86,15 @@ export const MethodScene: FC<DocMethodProps> = ({ api }) => {
     </RunItButton>
   )
 
+  document.getElementById('top')?.scrollTo(scrollPx, 0)
```
noahamac, line 89: Open to recommendations on where to refactor this into

jkaster: make a util component for it. `<ScrollToTop>` possibly?

jkaster (changes requested):
Please create  a scroll to top component

packages/api-explorer/src/scenes/MethodScene/MethodScene.tsx
josephaxisa: We are using a package for this https://www.npmjs.com/package/react-router-hash-link and I'm hoping we can avoid having to do it manually. Is there anything in particular that makes it break on some methods?

noahamac:
@josephaxisa I was just looking into the `react-router-hash-link` package but couldn't get the "#top" to end of URL working. Was this functionality working at some point? https://github.com/looker-open-source/sdk-codegen/blob/6734edb21496bc44af68a3c26289cccf7c4d6095/packages/api-explorer/src/components/SideNav/SideNavMethods.tsx#L91

noahamac:
Ok @josephaxisa and @jkaster I actually feel good about this change now. :smile: 

Given we cannot use BrowerRouter, this change extends the Section `@looker/component` to use scrollToTop.

noahamac:
How to test this change? Do we have integration tests for this project?

jkaster:
We don't have integration tests. We currently only have React Testing Library tests. We haven't implemented browser-based "integration" tests for our OSS projects yet, but I think @lukelooker make have suggestions based on what the Components team has done? Perhaps a snapshot.

packages/api-explorer/src/ApiExplorer.tsx
```diff
@@ -98,7 +98,7 @@ const ApiExplorer: FC<ApiExplorerProps> = ({
               specDispatch={specDispatch}
               toggleNavigation={toggleNavigation}
             />
-            <Layout hasAside>
+            <Layout hasAside height={'100%'}>
```
josephaxisa, line 101: Very minor nit. Consider dropping the curly braces for literal values.

packages/api-explorer/src/components/SideNav/SideNav.tsx
```diff
@@ -118,7 +118,7 @@ export const SideNav: FC<SideNavProps> = ({ api, specKey }) => {
         <Tab>Methods ({methodCount})</Tab>
         <Tab>Types ({typeCount})</Tab>
       </TabList>
-      <TabPanels {...tabs} pt="xsmall">
+      <TabPanels {...tabs} pt="xsmall" height={'90%'} overflow={'auto'}>
```
josephaxisa, line 121: Same here

josephaxisa (changes requested):
Thanks for diving into this, Noah. It seems that this will solve it for the sandboxed iframe case and keep it working in the standalone version. I left some very minor suggestions. There are still two items that I think belong in this PR:

1. `top` needs to be redefined. We need to either make the APIX Header fixed such that it can never be scrolled away or make it the `top`, otherwise it will disappear whenever we scroll to top. 
2. Since this solution works for both standalone and extensions then the `react-router-hash-link` dependency can be eliminated by replacing `NavHashLink` with `NavLink` in `MethodScene` and `TypeScene`.

lukelooker:
> We don't have integration tests. We currently only have React Testing Library tests. We haven't implemented browser-based "integration" tests for our OSS projects yet, but I think @lukelooker make have suggestions based on what the Components team has done? Perhaps a snapshot.

You can definitely test the presence of the style rules in RTL (`toHaveStyle` & `toHaveStyleRule`) but that's poor proof that things are _actually working_. We use Storybook + Storyshots + Puppeteer + Jest to drive visual snapshot tests for cases like this but that also was a fairly extensive amount of infrastructure to configure and maintain within our monorepo. 

As an aside, we are working on some Layout specific issues that should negate the need for this change as well as provide good test coverage for use cases like this (I'll send links to internal bugs out-of-band)

noahamac:
@josephaxisa I'm not sure why I was thinking the Header should float up like that. 

This PR fixes that, refactors the solution, and removes `react-router-hash-link` and references to it.

packages/api-explorer/src/scenes/TypeScene/TypeScene.tsx
```diff
@@ -25,10 +25,10 @@
  */
 
 import React, { FC } from 'react'
-import { Section } from '@looker/components'
 import { typeRefs, methodRefs, ApiModel } from '@looker/sdk-codegen'
 import { useParams } from 'react-router-dom'
 import { DocReferences, DocSDKs, DocTitle, ExploreType } from '../../components'
+import { ApixSection } from '../../components/common'
```
josephaxisa, line 31: Nit: These two can be combined

josephaxisa (commented):
Great work! One very minor suggestion below

jkaster (approved):
LGTM. @josephaxisa covered all the concerns for this PR, I think

josephaxisa (approved):
👍

PR #531, opened 2021-03-24
ci: support a CI:HOLD label

joeldodge79:
by applying the CI:HOLD label you can avoid triggering
CI runs. This is mostly useful if you expect a lot of
thrashing on your PR before it is actually ready for tests


jkaster (approved):
LGTM

PR #534, opened 2021-03-25
refactor: rename sdk-rtl to rtl

joeldodge79:


joeldodge79:
Here's an example start with just renaming `@looker/sdk-rtl` to `@looker/rtl` and resetting the version to `1.0.0`. Manual steps:

1. edit `packages/sdk-rtl/package.json` to rename and set version
2. sed cmd to update other package.json files:
  a. `git grep '@looker/sdk-rtl' | grep package.json | cut -d: -f1 | sort | uniq | xargs sed -i 's|@looker/sdk-rtl": "^21.0.10"|@looker/rtl": "^1.0.0"|g'`
3. sed cmd to update all other source code references to `@looker/sdk-rtl`:
  a. `git grep '@looker/sdk-rtl' | grep -v package.json | cut -d: -f1 | sort | uniq | xargs sed -i 's|@looker/sdk-rtl|@looker/rtl|g'`
4. manually fix up `jest.config.js`
5. manually fix up `examples/access-token-server/package.json` and `examples/typescript/package.json`

jkaster:
Thanks for this work. I'm not sure we want to lose the `sdk-` prefix for the RTL but I'm open to persuasion.

PR #539, opened 2021-03-30
chore: release-please config fine tuning

joeldodge79:
trying out bump-minor-pre-major option to see if that coerces
a "feat" commit on a pre-major package to bump the patch instead
of the minor as minor bumps are considered breaking in pre-major
packages


jkaster (approved):
LGTM but I look forward to seeing if it works

PR #540, opened 2021-03-30
fix: legacy generator was broken

### LINKED ISSUE #442: Generate PHP SDK
Is there a way to generate PHP SDK?
As it is said in the readme:

_configure the desired language in codeGenerators.ts. Currently, only Go and C# have legacy language generation configured, and we now have a prototype Look# SDK that can be used instead of the legacy C# generator._

How exactly PHP language should be configured there? Commenting out PHP section of the code in the `codeGenerators.ts.` file doesn't work.

Could someone help?

jkaster:
- Fixed the issue with the legacy generator (it was missing an `await` after recent spec processing changes)
- Corrected the php legacy language declaration and left it active
- Corrected legacy language declarations that are still commented out

update and use the legacy generator with:
```sh
yarn && yarn build
yarn legacy php
```

Fixes #442 🦕

### DIFF
--- a/packages/sdk-codegen-scripts/src/fetchSpec.ts
+++ b/packages/sdk-codegen-scripts/src/fetchSpec.ts
@@ -285,6 +285,7 @@ export const getVersionInfo = async (
 /**
  * Fetch (if needed) and convert a Swagger API specification to OpenAPI
  * @param name base name of the target file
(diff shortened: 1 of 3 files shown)


joeldodge79 (approved):
🦵

PR #543, opened 2021-03-31
docs(python): Google Cloud Function python SDK example

izzymiller:
This is a simple example of a Google Cloud Function leveraging the python SDK. The specific README describes proper configuration of a GCF with the looker SDK, and main.py provides a practical user provisioning workflow.

### DIFF
--- a/examples/python/cloud-function-user-provision/main.py
+++ b/examples/python/cloud-function-user-provision/main.py
@@ -8,6 +8,8 @@ a looker user based on another value provided. It could be easily adapted to acc
 of email addresses, to catch a webhook from a third party service, and much more.
 
 It could also easily be adapted to accept a second parameter and disable users as well.
+
+Example body: {"email": "test@test.com"}
 """
 
 # We init the SDK in the "global scope" so that we don't have to do it on every execution.
@@ -55,6 +57,6 @@ def create_user(request):
             if res == 200:
                 return "User already existed; Password reset sent", 200
             else:
-                return "Failure in user creation", 400
+                return "Failure in user creation.", 400
     else:
         return "No email provided", 422
--- a/examples/python/cloud-function-user-provision/main.py
+++ b/examples/python/cloud-function-user-provision/main.py
@@ -1,0 +1,60 @@
+import requests
+import looker_sdk
+import os
+
+"""
+This Google Cloud Function takes in an email address and create
+a looker user based on another value provided. It could be easily adapted to accept an array
+of email addresses, to catch a webhook from a third party service, and much more.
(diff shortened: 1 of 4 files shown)


examples/python/cloud-function-user-provision/main.py
```diff
@@ -0,0 +1,62 @@
...
+
+Example body: {"email": "test@test.com"}
+"""
+
+# We init the SDK in the "global scope" so that we don't have to do it on every execution.
+# This saves time if the function is called multiple times within a short timeframe.
+
+sdk = looker_sdk.init40()
+
+def create_user(request):
```
josephaxisa, line 20: Context: Came back here after going through the function. I think the name can be made more accurate, perhaps "create_user_or_reset_existing"?

josephaxisa (changes requested):
Awesome example! I left some very minor suggestions below.

I also have an idea on how to potentially make it flow better but please don't feel like you need to implement it, this is fine as it is! Anyway, here's what I'm thinking:

You could break out L55 into its own function, called `find_existing(email: str) -> Optional[int]` which returns the id if the user exists otherwise `None`. The sequence would then be:

```py
def create_user_or_reset_existing(request):
  email = request['email']
   
  if not email:
   message = "No email provided"
   code = 422
  elif id := find_existing(email):
    # The := operator was introduced in 3.8
    try:
      res = sdk.send_user_credentials_email_password_reset(id)
      message = "User already existed; Password reset sent"
      code = 200
    except SDKError as ex:
      message = ex
      code = 400
  else:
    try:
      # create new user
      message = "New user created"
      code = 200
    except SDKError as ex:
      message = ex
      code = 400

  return message, code
```

you could even break out the create user and reset user into helper functions.

```py
# only showing the reset one for brevity
def send_credentials_reset_link(id):
   try:
      res = sdk.send_user_credentials_email_password_reset(id)
      message = "User already existed; Password reset sent"
      code = 200
   except SDKError as ex:
      message = ex
      code = 400
   return message, code

def create_user_or_reset_existing(request):
  email = request['email']
   
  if not email:
   message = "No email provided"
   code = 422
  elif id := find_existing(email):
    message, code = send_credentials_reset_link(id)
  else:
    message, code = create_user(email)

  return message, code
```

lanlooker:
Hi @izzymiller - hope you are doing well! Do you have time to polish the snippet per Jax's suggestion above? 

@josephaxisa If Izzy is busy, I think we can merge the commits here as they are. I'll work on refactoring, and resubmitting for your final check? Izzy, would this be ok with you?

PR #547, opened 2021-04-01
fix: Typescript SDK import generation logic

jkaster:
- Some assumptions about types that will **always** be imported was wrong, so now some sdk-rtl dependencies are specifically tracked for importing
- also moved the `IDictionary<T>` declaration to `@looker/sdk-rtl` where it belongs
- also logging a bit more of the generation process
- introduced `ICodeGen.reset()` to reset tracking variables before generating methods, streams, and models


packages/sdk-codegen/src/typescript.gen.ts
```diff
@@ -510,11 +529,13 @@ export interface IDictionary<T> {
             name: `${map.name}[]`,
           }
         case 'HashType':
+          this.rtlNeeds.add('IDictionary')
```
joeldodge79, line 532: thank goodness for `Set`s

joeldodge79 (approved):
LGTM

quick question though: every time I `yarn gen` I see `spec/Looker.3.1.json` et all change. Should we be including those changes in our PRs? I've been just discarding those changes...

also, do we need to track yarn.lock?

packages/sdk-codegen/src/typescript.gen.ts
jkaster: Yes!

jkaster: We track `yarn.lock` in other repositories so that's why I've been doing it here. The spec generation could be fine to ignore but I just prettify it and commit it again without doing a very deep compare. I wiped just to make sure the 21.4 reference release was used for the latest saved spec.

PR #549, opened 2021-04-01
feat: added -n | --nostreams to code generator options

jkaster:
- for SDKs that support streaming, this skips generating the `streams` file. It does *not* remove any existing streams files, or change any other generated files
- expanded the `-h` display
- invalid gen options are rejected with an error
- recognize 'all' as a valid language switch but it can still be overridden if other language specifiers are used, so it's undocumented

### DIFF
--- a/packages/sdk-codegen-scripts/src/utils.ts
+++ b/packages/sdk-codegen-scripts/src/utils.ts
@@ -65,10 +65,29 @@ export interface IGenProps {
   apis: string[]
   /** Last API version */
   lastApi: string
+  /** Skip generating streams files? */
+  noStreams: boolean
 }
 
 const generatorHelp = () => {
-  log(`sdkGen [languages...] [-v|--versions <versions file>] [-h|--help]`)
+  log(
+    `sdkGen [languages...] [-v|--versions <versions file>] [-n|--nostreams] [-h|--help]
+  languages...:   zero or more language specifiers separated by space or comma. Defaults to 'all'
+  -v|--versions:  location of a JSON versions file in ILookerVersions format to read for getting specs
+  -n|--nostreams: skip generation of a language SDK 'streams' files (if it supports streaming)
+  -h|--help:      display this output
+
+  examples:
+    # Generates all supported SDKs
+    yarn gen
+
+    # Generates Typescript and Python SDKs
+    yarn gen ts,py
+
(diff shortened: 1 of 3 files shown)


packages/sdk-codegen-scripts/src/utils.ts
```diff
@@ -97,16 +117,24 @@ export const doArgs = (args: string[]) => {
...
           {
             const values = arg.split(',').filter((v) => v.trim())
-            values.forEach((v) => {
-              const gen = findGenerator(v.trim())
-              if (gen) {
-                // Valid language match
-                langs.push(gen.language)
-              }
-            })
+            if (values[0] !== 'all') {
```
joeldodge79, line 127: I'm a little lost about how `'all'` comes into play (both here and L103)

joeldodge79 (approved):
one question but not blocking an LGTM

packages/sdk-codegen-scripts/src/utils.ts
jkaster: it's possible someone may do, for example:

```sh
yarn -n all
```
for example

jkaster: I could probably remove the premature optimization bit of 103, but it's such a corner case ...

packages/sdk-codegen-scripts/src/utils.ts
```diff
@@ -65,10 +65,29 @@ export interface IGenProps {
...
+  -v|--versions:  location of a JSON versions file in ILookerVersions format to read for getting specs
+  -n|--nostreams: skip generation of a language SDK 'streams' files (if it supports streaming)
+  -h|--help:      display this output
+
+  examples:
+    # Generates all supported SDKs
+    yarn gen
+
+    # Generates Typescript and Python SDKs
+    yarn gen ts,py
```
bryans99, line 85: so does it accept the file extension? I've been using `yarn gen Typescript`

bryans99 (approved):
Seems okay but I do have a question about the help text

packages/sdk-codegen-scripts/src/utils.ts
jkaster: yup. extension, label, or language name matches. Added that a little while ago.

jkaster: because I'm lazy

PR #550, opened 2021-04-02
fix: auth_session uses transport_options

### LINKED ISSUE #541: Python SDK does not forward transport_options in auth calls
any custom transport_options that a client sends through a generated method should be included in the auth_session.py _login

joeldodge79:
Now auth_session login/logout/login_user will pass along any original
transport_options from the sdk call.

Fixes #541 🦕


jkaster (approved):
LGTM. Distributing those options ...

PR #532, opened 2021-03-25
chore: release main

joeldodge79:
:robot: I have created a release \*beep\* \*boop\*
---
<details><summary>@looker/extension-api-explorer: 21.4.0</summary>

### Features

* Added --versions option to the code generator ([#514](https://www.github.com/looker-open-source/sdk-codegen/issues/514)) ([ee6f3e8](https://www.github.com/looker-open-source/sdk-codegen/commit/ee6f3e8f55e300df1a75c9be89b47f067bc08dee))

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @looker/api-explorer bumped from ^0.9.0 to ^0.9.1
    * @looker/extension-sdk bumped from ^21.0.10 to ^21.4.0
    * @looker/extension-sdk-react bumped from ^21.0.10 to ^21.4.0
    * @looker/run-it bumped from ^0.9.0 to ^0.9.1
    * @looker/sdk bumped from ^21.0.10 to ^21.4.0
    * @looker/sdk-codegen bumped from ^21.0.10 to ^21.0.11
</details>
<details><summary>@looker/extension-sdk-react: 21.4.0</summary>

### Features

* Added --versions option to the code generator ([#514](https://www.github.com/looker-open-source/sdk-codegen/issues/514)) ([ee6f3e8](https://www.github.com/looker-open-source/sdk-codegen/commit/ee6f3e8f55e300df1a75c9be89b47f067bc08dee))

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @looker/extension-sdk bumped from ^21.0.10 to ^21.4.0
    * @looker/sdk bumped from ^21.0.10 to ^21.4.0
    * @looker/sdk-rtl bumped from ^21.0.10 to ^21.0.11
</details>
<details><summary>@looker/extension-sdk: 21.4.0</summary>

### Features

* Added --versions option to the code generator ([#514](https://www.github.com/looker-open-source/sdk-codegen/issues/514)) ([ee6f3e8](https://www.github.com/looker-open-source/sdk-codegen/commit/ee6f3e8f55e300df1a75c9be89b47f067bc08dee))

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @looker/sdk bumped from ^21.0.10 to ^21.4.0
    * @looker/sdk-rtl bumped from ^21.0.10 to ^21.0.11
</details>
<details><summary>@looker/hackathon: 21.4.0</summary>

### Features

[...]

PR #551, opened 2021-04-02
chore(release-please): fix release-as for sdk-codegen and sdk-codegen-all

joeldodge79:
The python SDK had a breaking change which caused these semver packages to
major bump - not what we want right now.

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -10,7 +10,7 @@
     "packages/sdk-node": {},
     "packages/sdk": {},
     ".": {
-      "release-as": ""
+      "release-as": "1.1.0"
     },
     "packages/api-explorer": {
       "release-as": "",
@@ -27,7 +27,7 @@
       "release-as": ""
     },
     "packages/sdk-codegen": {
-      "release-as": ""
+      "release-as": "21.1.0"
     },
     "packages/sdk-rtl": {
       "release-as": ""


release-please-config.json
```diff
@@ -10,7 +10,7 @@
     "packages/sdk-node": {},
     "packages/sdk": {},
     ".": {
-      "release-as": ""
+      "release-as": "1.1.0"
```
jkaster, line 13: packages/sdk should be "21.4.0" or "21.4.1" shouldn't it?

jkaster (changes requested):
packages/sdk looks like it's the wrong version

release-please-config.json
joeldodge79: so, this config file is a little tricky. L4 is setting the default for every package. Then each package can override it with either a specific release-as version or `""` which means "override with whatever conventional commits would say"

That's what `"."` and `"packages/sdk-codegen"` were set to but then I goofed it up and leaked a "BREAKING CHANGE" into their history (really the breaking change is only for the python SDK but I included the sdk-codegen changes in that single commit which I really should not have done given these consequencyes). So temporarily we have to set them to just bump the minor version like we actually want them to.

After #532 eventually merges we'll set these back to to `""`

jkaster: All righty then! Glad you can wrap your head around this.

jkaster (approved):

release-please-config.json
joeldodge79: open to design feedback. Our use-case is complicated with a mix of semver and non-semver packages as well as the inter-connectedness of packages and how that relates to forming conventional-commits/PRs that then correspond to version bumps and changelog entries. I'm writing a doc to hopefully distribute that knowledge so I'm not the only key holder

PR #557, opened 2021-04-05
chore(release-please): post 21.4 cleanup

joeldodge79:
- next looker release is 21.6
- assume that sdk-* packages will patch bump again regardless
- no more need for one time "bootstrap-sha"

- if any looker versioned package needs a patch bump then a specific
  "release-as" key will need to be added here for that release

misc:
- no need to track Pipfile.lock files
- bump typescript example depenedencies to 21.4


release-please-config.json
```diff
@@ -10,7 +9,7 @@
     "packages/sdk-node": {},
     "packages/sdk": {},
     ".": {
-      "release-as": "1.1.0"
+      "release-as": ""
```
joeldodge79, line 12: this should be the normal value for the root package - it's just that last time the python "breaking change" was going to bump to 2.0.0 which didn't seem quite right.

release-please-config.json
```diff
@@ -21,16 +20,16 @@
...
-      "release-as": "21.0.11"
+      "release-as": "21.0.12"
     },
     "packages/sdk-codegen": {
-      "release-as": "21.0.11"
+      "release-as": "21.0.12"
     },
     "packages/sdk-rtl": {
-      "release-as": "21.0.11"
+      "release-as": "21.0.12"
```
joeldodge79, line 32: we may come up with another solution before the next release (e.g. renaming these packages) but for now this is their next version

jkaster (approved):
LGTM!

PR #558, opened 2021-04-06
fix: extends runit response height and fixes response models scrolling

noahamac:
https://looker.atlassian.net/browse/EMBED-1482
Addresses the run it response height

Also resolves a bug that came up where the "Response Models" `<Tree/>` overflow wasn't properly captured and was causing an unintended x-scroll to the Method container when pressing "Run It". Also fixes the Tree detail covering the label on smaller screen widths (such as when Run It is open).


packages/api-explorer/src/components/ExploreType/ExploreProperty.tsx
```diff
@@ -158,7 +158,7 @@ export const ExplorePropertyDescription: FC<ExplorePropertyProps> = ({
 export const ExplorePropertyDetail: FC<ExplorePropertyProps> = ({
   property,
 }) => (
-  <Space>
+  <Space style={{ fontSize: 'small', marginLeft: '10rem' }}>
```
noahamac, line 161: Prevents `ExplorePropertyDetail` from covering the label
Bug demo here:
https://drive.google.com/file/d/1c9eHOBJ2utbUKm_Oj8A1u8j2O7m66E7R/view?usp=sharing&resourcekey=0-ZX13bwjqnikSfFQ_WKc3rA

josephaxisa (commented):
I am seeing the below in both the standalone and extension versions for methods with a body:

image

joeldodge79:
should this be a `chore:...` or a `fix:...`? I don't actually know, just poking. Remember, the former **will not** show up in any changelogs nor cause version bumps (which for these non-published packages isn't really publicly meaningful anyway but still...). The latter **will** show up in the change log and promote the next release version to be eligible for at least a **patch** bump.

Also, keep in mind that this will end up as a single squashed commit and only the title will appear in the changelog. AND, as it currently stands, that exact same changelog entry will appear for both packages/api-explorer/CHANGELOG.md as well as packages/run-it/CHANGELOG.md

-- your friendly neighborhood conventional-commits/release-please education committee :-)

noahamac:
Glaring oversight @josephaxisa Thanks for the catch! 

Better solution now where ShowResponse passes in its own height param and other uses default to the AceEditor default height

jkaster (approved):
This looks much better!

PR #565, opened 2021-04-08
fix: RunIt request overflow 

noahamac:
While playing around with API Explorer, I found that the RunIt tab panels do not properly overflow. This PR adds styling to the extension html that restricts the page to 100%, and adds overflow to the RunIt tabs.

This is related to changes made in https://looker.atlassian.net/browse/EMBED-1482 or https://github.com/looker-open-source/sdk-codegen/pull/558

Gif repro:
runit_overflow

Gif fix:
runit_overflow_fix


packages/extension-api-explorer/src/ExtensionApiExplorer.tsx
```diff
@@ -40,6 +40,7 @@ import {
 } from '@looker/sdk-codegen'
 import { Looker31SDK, Looker40SDK } from '@looker/sdk'
 import ApiExplorer from '@looker/api-explorer/lib/ApiExplorer'
+import './styles.css'
```
noahamac, line 43: What do you think about this? Tried adding to ApiExplorer but the .css wasn't getting bundled.

noahamac: What is `limit`?

noahamac: No, the gap between `body` and `limit` isn't new, here's the current build:
image

The request body and sdk declaration AceEditor have the default height of 500px with dynamic content size whereas the response body height is 85% of page height which we can consistently expect

jkaster:
Interesting. I wonder when that changed. We can't release that.

noahamac:
Maybe we can decrease the default height for the request body by half? To 250px, as we can expect this code block to be shorter. This just means some request bodys will have overflow or could overflow when adding additional keys.

noahamac:
250px looks like this 
image

jkaster (changes requested):
I'm pretty sure we should convert the css to a styled component instead. Don't worry about the gap after `body` because we're going to replace editor shortly (within days).

packages/run-it/src/RunIt.tsx
```diff
@@ -192,7 +192,13 @@ export const RunIt: FC<RunItProps> = ({ api, inputs, method }) => {
         {isExtension ? <></> : <Tab key="performance">Performance</Tab>}
         <Tab key="makeTheCall">Code</Tab>
       </TabList>
-      <TabPanels px="xxlarge" {...tabs}>
+      <TabPanels
+        px="xxlarge"
+        {...tabs}
+        height="90%"
```
lukelooker, line 198: `90%` is just a guess value correct?

packages/extension-api-explorer/src/ExtensionApiExplorer.tsx
lukelooker: Couple of ways to solve this problem without needing an additional CSS file (and the webpack fun involved with loading that).

Specify `100vh` (100% of view-height) on the outer-most DOM element of the application layout.

Alternatively, you can use Styled Components' `createGlobalStyle` function:

```
export const BodyOverride = createGlobalStyle`
  html {
    height: 100%;
    overflow: hidden;
  }
`
```

packages/api-explorer/src/ApiExplorer.tsx
```diff
@@ -72,39 +74,42 @@ const ApiExplorer: FC<ApiExplorerProps> = ({
...
+                  specKey={spec.key}
+                  specs={specs}
+                  toggleNavigation={toggleNavigation}
+                />
+              </Layout>
+            </Page>
+          </SearchContext.Provider>
+        </LodeContext.Provider>
+      </ComponentsProvider>
+      <BodyOverride />
```
noahamac, line 111: This is the change. The other diffs in this block are from tabbing

jkaster (approved):
LGTM

PR #566, opened 2021-04-08
chore: fix release-metrics workflow yml

joeldodge79:


jkaster (approved):
LGTM. But then again, so did the last one.

joeldodge79:
> LGTM. But then again, so did the last one.

yeah, I need to find a good yaml validator. Found one online that's a little janky. I think the problem with the last one was the use of `|` and `>` characters in the shell command being part of a `plain scalar` string? anyway, hopefully single quoting the whole durn thing works this time

jkaster:
@joeldodge79 you could always use the codegen [yaml to json](https://github.com/looker-open-source/sdk-codegen/blob/main/packages/sdk-codegen-scripts/src/yamlToJson.ts) script to convert the YAML. Should blow up if invalid. Also, you can then use JSON validators.

PR #571, opened 2021-04-09
fix: remove search criteria selector and auto expand results

josephaxisa:

### DIFF
--- a/packages/api-explorer/src/components/SideNav/SideNav.spec.tsx
+++ b/packages/api-explorer/src/components/SideNav/SideNav.spec.tsx
@@ -24,14 +24,16 @@
 
  */
 import React from 'react'
-import { ApiModel } from '@looker/sdk-codegen'
+import { ApiModel, CriteriaToSet } from '@looker/sdk-codegen'
 import { pick } from 'lodash'
 import userEvent from '@testing-library/user-event'
-import { screen } from '@testing-library/react'
+import { act, screen, waitFor } from '@testing-library/react'
 
 import { api } from '../../test-data'
-import { renderWithRouter } from '../../test-utils'
+import { renderWithRouter, renderWithSearchAndRouter } from '../../test-utils'
+import { defaultSearchState } from '../../reducers'
 import { SideNav } from './SideNav'
+import { countMethods } from './searchUtils'
 
 describe('SideNav', () => {
   const testApi = ({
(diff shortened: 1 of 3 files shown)


joeldodge79 (changes requested):
can you remove the `(ApiExplorer)` scope from the PR title? That *should* cause the pre-populated squash merge commit title to also not have it but worth checking there too when you go to do it.

I'm about to post about it in the sdk dev channel.

jkaster (approved):
Looks pretty simple

jkaster:
Something else we'll have to add tests for when/before we have codecov :wink:

josephaxisa:
> Something else we'll have to add tests for when/before we have codecov 😉

@jkaster Unacceptable. I added search tests :)

packages/api-explorer/src/components/SideNav/SideNav.spec.tsx
```diff
@@ -74,3 +80,35 @@ describe('SideNav', () => {
...
 })
+
+describe('Search', () => {
+  test('it filters methods and types on input', async () => {
+    renderWithSearchAndRouter(<SideNav api={api} specKey="3.1" />)
+    const searchPattern = 'embedsso'
+    const input = screen.getByLabelText('Search')
+    jest.spyOn(api, 'search')
+    await act(async () => {
+      /** Pasting to avoid triggering search multiple times */
```
bryans99, line 91: I don't understand the comment. Perhaps a little more clarity on why.  Possibly because entering characters being entered one by results in search being triggered for each character?

Something like

`Using ... triggers search multiple times. Use of paste avoids this`

packages/api-explorer/src/components/SideNav/SideNavMethods.tsx
```diff
@@ -70,6 +68,13 @@ const SideNavMethodsLayout: FC<MethodsProps> = ({
     if (_isOpen) history.push(`/${specKey}/methods/${tag}`)
   }
 
+  useEffect(() => {
+    const status = match
```
bryans99, line 72: setIsOpen defaults to defaultOpen but you are modifying the value if the value of defaultOpen changes. Is defaultOpen a good name?

bryans99 (commented):
Not sure. See comments. Can discuss if you want or approve if you think its okay.

packages/api-explorer/src/components/SideNav/SideNavMethods.tsx
jkaster: We can find a better name for another PR

jkaster (approved):
LGTM. Minor internal tweaks can be done later.

PR #578, opened 2021-04-12
feat: add source declaration links to API Explorer

jkaster:
If a declarations index file produced with `yarn mine` is found, declaration source links get displayed for methods and types

The `apix-files` folder was added to provide an http server to serve the index files from the **current** source.

Run `yarn serve` from `apix-files` to serves the mined index files at `http://localhost:30000`.


joeldodge79:
Looking at the source code changes I think the commit message (aka changelog entry) is fine but just wanted to re-iterate that the exact same changelog entry will show up in each of `packages/{api-explorer,extension-api-explorer,sdk-codegen-scripts,sdk-codegen}/CHANGELOG.md` given that files in all of those packages are are touched in this PR. Again, I think that makes sense given the file changes but just wanted to call it out in case you hadn't realized and feel any different

apix-files/package.json
```diff
@@ -0,0 +1,38 @@
+{
```
joeldodge79, line 1: any reason not to house this under `packages/`? I realize there are no dependencies in either direction but still kind of seems to me like maybe we should keep all our typescript sdk-codegen packages under `packages/` (and who knows, maybe there will be a dependency some day?)

jkaster: Yes, it is not a monorepo package and doesn't need all the baggage associated with that. If it ever becomes something we customize, we'll move it into `packages/apix-files`. Currently, it's basically just a node script that needs to be isolated.

jkaster: Yeah, all those places were changed to support this so the CHANGELOG entry is appropriate.

apix-files/README.md
```diff
@@ -0,0 +1,40 @@
+# API Explorer file server
+
+API Explorer (aka APIX) can be uses some JSON-formatted "index" files to augment the information provided in a specification.
```
bryans99, line 3: > API Explorer (aka APIX) can be uses some 

?

> API Explorer (aka APIX) uses some 

instead?

apix-files/README.md
```diff
@@ -0,0 +1,40 @@
...
+```sh
+yarn && yarn build
+yarn mine
+cd apix-files
+yarn serve
+```
+
+This will start the API Explorer file server at `http://localhost:30000`
+
+## Mining the mother lode
```
bryans99, line 16: Any issues with mother lode as a politically correct term?

packages/api-explorer/src/components/DocSdkUsage/DocSdkUsage.tsx
```diff
@@ -28,7 +28,7 @@ import { TabList, Tab, TabPanels, TabPanel, useTabs } from '@looker/components'
 import { findExampleLanguages, IMethod } from '@looker/sdk-codegen'
 
 import { CollapserCard } from '../Collapser'
-import { LodeContext } from '../../context/examples'
+import { LodeContext } from '../../context/motherlode'
 import { DocExamples } from './DocExamples'
 
 interface DocSdkUsageProps {
```
bryans99, line 34: I read this as DocSausageProps :D

packages/extension-api-explorer/README.md
```diff
@@ -22,11 +22,12 @@ The API Explorer extension can be manually installed and run with a Looker insta
          use_form_submit: yes
          use_embeds: yes
          core_api_methods: ["versions", "api_spec"]
-         external_api_urls: ["https://raw.githubusercontent.com"]
+         external_api_urls: ["https://raw.githubusercontent.com", "http://localhost:30000"]
          oauth2_urls: []
       }
    }
    ```
+   the `http://localhost:3000` is for when you want to use the [API Explorer file server](/apix-files/README.md)
```
bryans99, line 30: Be interesting to see how this works in marketplace. Do we want to put a dummy entry in so that user can configure hosts to point at it? Or just go with localhost:3000. 3000 pretty popular. Consider using non popular port

bryans99 (commented):
Code looks okay. 

Some minor doc issues.

Some things to think about

apix-files/README.md
jkaster: Well, it's definitely wrong now. I'll fix.

apix-files/README.md
jkaster: I don't *think" that's a charged term. It's taken from mining but I guess we can change it to "examples index" and "examplesIndex.json" to be less obscure

packages/api-explorer/src/components/DocSdkUsage/DocSdkUsage.tsx
jkaster: Definitely leaving it as is then!

packages/extension-api-explorer/README.md
jkaster: whoops, documentation error. That's supposed to be `30000`

bryans99 (approved):
LGTM

PR #580, opened 2021-04-13
feat: TS SDK tree-shaking support

jkaster:
Implemented a prototype for TS SDK tree shaking (saving JS bundle size) by:

- generating a `funcs.ts` per API version
- also added `methodsInterface.ts` to define the SDK's contract
- silenced thousands of lint warnings by explicitly declaring the unnecessary return value for each function. Needed it for `methodsInterface.ts` anyway
- added a `functionalSdk` init helper function
- created `tinyDual` example. see `examples/typescript/README.md` to run it

This prototype does *not* use a special prefix for the functional names. However, people who want to use this could pretty easily
- start with a `Looker40SDK` object
- write their code
- switch the initialization to `functionalSdk`
- flip the `sdk.foo(` to `foo(sdk`

So maybe we can avoid the need to use a special prefix for clarity


examples/typescript/README.md
```diff
@@ -20,6 +20,7 @@ These examples assume that a `looker.ini` has been created in the root of the re
 2. Run `yarn` to install the dependencies.
 3. Run the examples - `yarn {command} [other command-line options]`
    - `yarn run-dual` - runs multiple apis example.
+   - `yarn run-tiny-dual` - runs the compact multiple apis example
    - `yarn run-config` - runs custom config reader example.
    - `yarn run-sudo` - runs sudo as user example.
    - `yarn run-tile` - runs download dashboard tile example.
```
jkaster, line 26: Could we omit the `run-` from these commands?

examples/typescript/tinyDual.ts
```diff
@@ -0,0 +1,48 @@
...
+import { me } from '@looker/sdk/lib/4.0/funcs'
+import { me as me31 } from '@looker/sdk/lib/3.1/funcs'
+import { sdkVersion } from '@looker/sdk'
+import { rootIni } from './utils'
+
+const localConfig = rootIni()
+const settings = new NodeSettingsIniFile('', localConfig, 'Looker')
+const session = new NodeSession(settings)
+const sdk = functionalSdk(session, '4.0', sdkVersion)
+const sdk31 = functionalSdk(session, '3.1', sdkVersion)
```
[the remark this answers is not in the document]
bryans99: I still like thus but I was looking at the implementation and wondering if we could simplify more. Can we add these methods to `@looker/sdk`?

`functionalSdk40(session)`
`functionalSdk31(session)`

That way '4.0' and sdkVersion do not need to be provided and imported.

packages/sdk-rtl/src/apiMethods.ts
```diff
@@ -26,17 +26,169 @@
...
 import { IAuthSession } from './authSession'
 
-export class APIMethods {
+/**
+ * Creates an "sdk" to be used with the Typescript SDK's funcs.ts file
+ * @param authSession authentication session
+ * @param apiVersion version of API to use (e.g. "3.1" or "4.0")
+ * @param sdkVersion Looker release version (e.g. "21.4")
+ */
+export const functionalSdk = (
```
bryans99, line 45: See my comment on the demo. Can we wrap this method in methods added to the SDK package. The SDK function can hard code the apiVersion and sdkVersion.

packages/sdk/src/constants.ts
```diff
@@ -24,5 +24,19 @@
...
+
 export const sdkVersion = '21.4'
 export const environmentPrefix = 'LOOKERSDK'
+
+/**
+ * Creates an "sdk" to be used with the Typescript SDK's funcs.ts file
+ * @param authSession authentication session
+ * @param apiVersion version of API to use (e.g. "3.1" or "4.0")
+ */
+export const functionalSdk = (
```
bryans99, line 37: Seems strange to have this in `constants.ts`

Again, add two methods here for `functionalSdk31` and `functionalSdk40`?

bryans99 (commented):
See comments

examples/typescript/tinyDual.ts
```diff
@@ -0,0 +1,46 @@
...
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+
+ */
+
+import { NodeSession, NodeSettingsIniFile } from '@looker/sdk-node'
+import { me, functionalSdk40 } from '@looker/sdk/lib/4.0/funcs'
```
bryans99, line 28: YES!!!!!

examplesIndex.json
```diff
@@ -4148,7 +4153,43 @@
           {
             "sourceFile": "packages/extension-api-explorer/src/ExtensionApiExplorer.tsx",
             "column": 36,
-            "line": 108
+            "line": 113
+          }
+        ]
+      }
+    },
+    "localStorageGetItem": {
```
bryans99, line 4161: Hm! Picked up an extensionSDK example :D . I don't think its an issue

bryans99 (approved):
LGTM

PR #581, opened 2021-04-13
feat: adds prism editor and implements for sdkdocs

noahamac:
Brought in the Prism editor by following the documentation very closely: https://www.npmjs.com/package/prism-react-renderer#usage and implementing for the SdkDocs component. 

Also added integration with the search functionality.

https://looker.atlassian.net/browse/EMBED-1258
https://looker.atlassian.net/browse/EMBED-1154

Looking for advice on comments
1) Where to move this? Looking to get those components out and function to a util file. Should be accessible for RunIt as well. Would like to be able to bring into Data Dict and Lookml Diagram at some point too. 
2) Search integration feedback. This is obviously unsophisticated, should we fuzzy match? Make case insensitive? Style a smaller token? Is it annoying that the main page zoomstotop? Should I add a prop to disable when searching? 
3) I'm having a hard time finding ApiExplorer's node_modules for where to import this. And TS doesn't seem to like the global addition... Any thoughts? 

Follow up PR could replace Ace usage for RunIt response.


packages/api-explorer/src/components/DocCode/PrismEditor.tsx
```diff
@@ -0,0 +1,109 @@
...
+  padding-right: 1em;
+  user-select: none;
+  opacity: 0.5;
+`
+
+const LineContent = styled.span`
+  display: table-cell;
+`
+
+export const PrismEditor: FC<PrismEditorProps> = ({
```
noahamac, line 65: 1. Where to move this? Looking to get those components out and function to a util file. Should be accessible for RunIt as well. Would like to be able to bring into Data Dict and Lookml Diagram at some point too.

packages/api-explorer/src/components/DocCode/PrismEditor.tsx
```diff
@@ -0,0 +1,109 @@
...
+      {({ className, style, tokens, getLineProps, getTokenProps }) => (
+        <Pre className={className} style={style}>
+          {tokens.map((line, i) => (
+            <Line key={i} {...getLineProps({ line, key: i })}>
+              <LineNo>{i + 1}</LineNo>
+              <LineContent>
+                {line.map((token, key) => {
+                  const tokenProps = getTokenProps({ token, key })
+                  const text = tokenProps.children
+                  if (pattern !== '' && text.includes(pattern)) {
```
noahamac, line 93: Search integration feedback. This is obviously unsophisticated, should we fuzzy match? Make case insensitive? Style a smaller token? Is it annoying that the main page zoomstotop? Should I add a prop to disable when searching?

packages/api-explorer/src/components/DocCode/PrismEditor.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+ */
+
+import React, { FC } from 'react'
+import styled from 'styled-components'
+import Highlight, { defaultProps, Language } from 'prism-react-renderer'
+// import { Prism } from "prism-react-renderer"
+import dracula from 'prism-react-renderer/themes/dracula'
+
+// (typeof global !== "undefined" ? global : window).Prism = Prism
```
noahamac, line 33: 3. I'm having a hard time finding ApiExplorer's node_modules for where to import this. And TS doesn't seem to like the global addition... Any thoughts?

noahamac: For some reason we need to do this for kotlin, c#, swift etc support https://www.npmjs.com/package/prism-react-renderer#faq

packages/api-explorer/src/components/DocCode/PrismEditor.tsx
```diff
@@ -0,0 +1,109 @@
...
+      {...defaultProps}
+      code={code}
+      language={getPrismLanguage(language)}
+      theme={dracula}
+    >
+      {({ className, style, tokens, getLineProps, getTokenProps }) => (
+        <Pre className={className} style={style}>
+          {tokens.map((line, i) => (
+            <Line key={i} {...getLineProps({ line, key: i })}>
+              <LineNo>{i + 1}</LineNo>
```
noahamac, line 88: 4. Do we like the line number? Can ditch this

noahamac:
new editor img: https://drive.google.com/file/d/1hy4GIEgvI8ZFJnKXXkhQcdN7ZleC_QNd/view?usp=sharing&resourcekey=0-kjhrx1HEFnuoTe8J-nNeQg

new editor search: https://drive.google.com/file/d/1bnjVbza-Fhn5TW4nokj_peZmtQ2r9TLH/view?usp=sharing&resourcekey=0-JGxc-1CAXRTUOIlJVp41lw

packages/api-explorer/src/components/DocCode/PrismEditor.tsx
```diff
@@ -0,0 +1,109 @@
...
+  language,
+  code,
+  pattern,
+}) => {
+  const getPrismLanguage = (language: string) => {
+    if (language === 'golang') {
+      return 'go'
+    }
+    return language as Language
+  }
```
josephaxisa, line 75: This is ok for now, but we can change the generator language/label to `Go` in `@looker/sdk-codegen`. It was changed to `Golang` to avoid exactly this for Ace :)

packages/api-explorer/src/components/DocCode/PrismEditor.tsx
josephaxisa: I don't have a preference

packages/api-explorer/src/components/DocCode/PrismEditor.tsx
josephaxisa: I don't think it's bad as is. There are times where it highlights the whole line and is a bit confusing but it's not a big deal. I do think we should open an issue to revisit highlighting across the whole app though because some of it got broken in the latest styling round

packages/api-explorer/src/components/DocSDKs/DocSDKs.tsx
```diff
@@ -77,7 +77,7 @@ export const DocSDKs: FC<LanguageSDKProps> = ({ api, method, type }) => {
                 : gen.declareType('', item as IType)
               return (
                 <TabPanel key={language}>
-                  <DocCode language={language} code={code} />
+                  <DocCode language={language} code={code} editor="prism" />
```
josephaxisa, line 80: Perhaps the prop should say `editor="ace"` for places where we are still using ace so that it's easier to find and replace later on. This is completely fine as is though, just throwing a suggestion out there.

packages/api-explorer/src/components/DocCode/PrismEditor.tsx
josephaxisa: I am not very familiar with how all things works under the hood. Maybe @bryans99 could chime in on this. Could we somehow lazy load the languages based on the user's sdk language preference?

josephaxisa (commented):
Nice progress! A few thoughts/comments below

packages/api-explorer/src/components/DocCode/PrismEditor.tsx
noahamac: And what if it needs to be used by both ApiExplorer and RunIt? (And DD and Lookml Diagram)? Are there are docs on creating a new packages item? Would this be a good candidate for that?

josephaxisa: If it needs to be used by both ApiExplorer and RunIt, like in this case, then it should live in RunIt. I don't believe we have any documentation on how to create a new package but I'd be happy to work on that with you

packages/api-explorer/src/components/DocCode/PrismEditor.tsx
jkaster: Line number can be useful for people discussing the code. Let's leave it in until we have people asking to remove it. Or we add a toggle/config setting when we start supporting APIX config settings more fully.

packages/api-explorer/src/components/DocCode/DocCode.tsx
```diff
@@ -56,13 +56,15 @@ interface DocCodeProps {
...
   width?: string
+  editor?: string
 }
 
 export const DocCode: FC<DocCodeProps> = ({
   code,
   language = 'json',
   fontSize = 16,
   width = 'auto',
+  editor = 'ace',
```
jkaster, line 67: Default to `prism`?

packages/api-explorer/src/components/DocCode/PrismEditor.tsx
```diff
@@ -0,0 +1,109 @@
...
+`
+
+export const PrismEditor: FC<PrismEditorProps> = ({
+  language,
+  code,
+  pattern,
+}) => {
+  const getPrismLanguage = (language: string) => {
+    if (language === 'golang') {
+      return 'go'
```
jkaster, line 72:
```suggestion
    // TODO revert back to `go` in generator language definitions instead of using this
    if (language === 'golang') {
      return 'go'
```

jkaster (commented):
Minor feedback

jkaster (approved):
This is a good step in the right direction

PR #588, opened 2021-04-14
feat: introduces prism-editor package; implements read-only usage

noahamac:
This PR expands on https://github.com/looker-open-source/sdk-codegen/pull/581 to implement CodeDisplay as a package. This centralizes the implementation for RunIt and ApiExplorer. This PR replaces AceEditor usage for all read-only code in ApiExplorer and RunIt.

I really like this package model! Please let me know if I missed something -- copy pasted the RunIt dir. Related question that may not be appropriate here: How to expose '@looker/code-display' for things not in sdk-codegen... like DD and Lookml Diagram? 

Follow up PRs, please advise on prio:
- Kotlin, C#, Swift highlighting 
- Remove all read only AceEditor from runit and apix


packages/prism-editor/webpack.dev.config.js
```diff
@@ -0,0 +1,44 @@
...
+    historyApiFallback: true,
+    publicPath: '/dist/',
+    headers: {
+      'Access-Control-Allow-Origin': '*',
+      'Access-Control-Allow-Methods': '*',
+      'Access-Control-Allow-Headers': '*',
+    },
+  },
+  devtool: 'inline-source-map',
+}
```
josephaxisa, line 44: This might not be needed for this package

packages/prism-editor/src/utils/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+  language = language.toLowerCase()
+  const unstyled = ['kotlin', 'csharp', 'swift']
+  // TODO revert back to `go` in generator language definitions instead of using this
+  if (language === 'golang') {
+    return 'go'
+  } else if (unstyled.includes(language)) {
+    return 'clike'
+  }
+  return language as Language
+}
```
josephaxisa, line 38: Hmm, this utility assumes that outside of a select few languages (Kotlin, csharp, swift and go), the Prism package will recognize all our language labels which might not always be the case. Let's add a test that ensures we are getting the desired language for all our codegenerator languages and fails if it encounters another Golang-like scenario

packages/prism-editor/src/PrismEditor.tsx
```diff
@@ -62,22 +63,16 @@ const LineContent = styled.span`
...
+ * for Looker apps. All @looker/sdk languages are supported.
+ * TODO: LookML syntax highlighting
+ */
 export const PrismEditor: FC<PrismEditorProps> = ({
-  language,
+  language = 'json',
   code,
-  pattern,
+  pattern = '',
 }) => {
```
josephaxisa, line 75: Can you add some tests for this component as well please? Let me know if you'd like to pair up :)

packages/prism-editor/src/PrismEditor.tsx
```diff
@@ -89,7 +84,7 @@ export const PrismEditor: FC<PrismEditorProps> = ({
         <Pre className={className} style={style}>
           {tokens.map((line, i) => (
             <Line key={i} {...getLineProps({ line, key: i })}>
-              <LineNo>{i + 1}</LineNo>
+              {language === 'json' || <LineNo>{i + 1}</LineNo>}
```
josephaxisa, line 87: I see that `LineNo` uses `span` under the hood. FWIW @looker/components has a `Span` component now, might want to use that instead. Same applies for L99

josephaxisa (changes requested):
Great progress Noah, this is shaping up nicely. A few recommended changes below.

It seems that Python syntax highlighting in run-it is not working but works fine for Typescript. Also, would it be possible to make text wrap inside Prism instead of having the horizontal scrollbar?

noahamac:
I thought the Python highlighting wasn't working in RunIt either but it looks like it is. Just that the dracula theme doesn't think anything in the default call should be highlighted... Here's what happens when I add a method body
image

packages/code-display/src/utils/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+
+ */
+import { Language } from 'prism-react-renderer'
+
+export const getPrismLanguage = (language: string): Language => {
+  language = language.toLowerCase()
+  const unstyled = ['kotlin', 'csharp', 'swift']
+  // TODO revert back to `go` in generator language definitions instead of using this
+  if (language === 'golang') {
+    return 'go'
```
josephaxisa, line 33: nit: consider having an enum/hash that maps our SDK languages to "prism languages" to avoid having to grow this if statement as our set of supported SDK languages grows

packages/run-it/src/components/ShowResponse/responseUtils.tsx
```diff
@@ -26,6 +26,7 @@
 import React, { ReactElement } from 'react'
 import { IRawResponse, ResponseMode, responseMode } from '@looker/sdk-rtl'
 import { Paragraph, CodeBlock } from '@looker/components'
+import { CodeDisplay } from '@looker/code-display'
 
 import { CodeStructure } from '../CodeStructure'
```
josephaxisa, line 31: It looks like Prism supports html syntax highlighting so you can replace `CodeStructure` on L117 as well and get rid of the import. In the end we should be left with `CodeStructure` only in `RequestForm`

josephaxisa (dismissed):
Woo! Great progress. A bit of feedback below

packages/code-display/package.json
```diff
@@ -0,0 +1,58 @@
...
+  ],
+  "author": "Looker",
+  "license": "MIT",
+  "private": true,
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/looker-open-source/sdk-codegen.git",
+    "directory": "packages/prism-editor"
+  },
+  "homepage": "https://github.com/looker-open-source/sdk-codegen/tree/main/packages/prism-editor",
```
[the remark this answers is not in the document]
josephaxisa: looks like I also missed L16 here

packages/code-display/src/utils/utils.ts
jkaster: Better to just change the language definition back to `go` from `golang` in sdk-codegen and this hack won't be needed

packages/code-display/package.json
```diff
@@ -0,0 +1,58 @@
...
+  "bugs": {
+    "url": "https://github.com/looker-open-source/sdk-codegen/issues?q=is%3Aopen+is%3Aissue+project%3Alooker-open-source%2Fsdk-codegen%2F7"
```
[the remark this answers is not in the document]
noahamac: :man_shrugging: lol...
This one is from runit. Next commit grabs the one from root

packages/code-display/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "react-test-renderer": "^17.0.1",
+    "style-loader": "^1.1.3",
+    "webpack-cli": "^3.3.11",
+    "webpack-dev-server": "^3.10.3"
+  },
+  "dependencies": {
+    "@looker/components": "^0.9.30",
+    "@looker/design-tokens": "0.9.27",
+    "@looker/sdk": "^21.4.0",
+    "@looker/sdk-codegen": "^21.0.11",
```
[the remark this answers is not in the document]
noahamac: [sdk-codegen is](https://github.com/looker-open-source/sdk-codegen/blob/110163e0c10126b240f767bf28c07c2d6ad09b8b/packages/code-display/src/utils/utils.spec.ts#L26)... sdk is not... will remove

packages/code-display/src/utils/utils.ts
noahamac: Do you have an example of an enum to enum mapping? 

I think most of the languages will be the same, as "go" was "golang" until it was changed for Ace. Once the other highlighers are brought in & we change golang back, this function becomes:
```
export const getPrismLanguage = (language: string): Language => {
  return language.toLowerCase() as Language
}
```
Which is an enum to enum mapper?

jkaster: codeGenerators.ts is where the languages are defined. If Ace is only left for the form editor now, that's json, so we don't need to worry about how Ace would highlight 'go'.  The cast should work fine depending on how Prism defines it.

packages/run-it/src/components/ShowResponse/responseUtils.tsx
```diff
@@ -117,7 +113,7 @@ const ShowImage = (response: IRawResponse) => {
 
 /** A handler for HTTP type responses */
 const ShowHTML = (response: IRawResponse) => (
-  <CodeStructure language={'html'} code={response.body.toString()} />
+  <CodeDisplay language={'html'} code={response.body.toString()} />
```
noahamac, line 116: For some reason I seem to have broken `ShowResponse.spec.tsx` test with this change... I spent a while looking into it but could not figure out why...

noahamac:
New theme... LMK what you think. 
image

You can try out the alternatives here by changing the theme import line 4:
https://codesandbox.io/s/prism-react-renderer-example-u6vhk?file=/src/Basic.js

jkaster:
Agreed that doesn't look like a monospaced font. The spacing indent is very subtle for the JSON too, but maybe it's ok. @miketorosian @jhardy @lukelooker what do you think?

miketorosian:
> Agreed that doesn't look like a monospaced font. The spacing indent is very subtle for the JSON too, but maybe it's ok. @miketorosian @jhardy @lukelooker what do you think?

Spoke to Noah about this this morning. We agreed that it should use Roboto Mono within the editor to be aligned with how we display code elsewhere.

packages/run-it/src/components/ShowResponse/responseUtils.tsx
jkaster: BTW, I like the name `CodeDisplay`

packages/run-it/src/RunIt.tsx
```diff
@@ -199,13 +199,7 @@ export const RunIt: FC<RunItProps> = ({
...
         <Tab key="makeTheCall">Code</Tab>
       </TabList>
-      <TabPanels
-        px="xxlarge"
-        {...tabs}
-        height="90%"
-        overflowX="hidden"
-        overflowY="auto"
-      >
+      <TabPanels px="xxlarge" {...tabs} overflow="auto" height="87%">
```
noahamac, line 202: Fixes runit button

noahamac:
image

noahamac:
CI:TEST

packages/code-display/src/CodeDisplay.tsx
```diff
@@ -26,80 +26,82 @@
...
-      theme={dracula}
+      language={highlighterLang}
+      theme={getOverridenTheme()}
     >
       {({ className, style, tokens, getLineProps, getTokenProps }) => (
         <Pre className={className} style={style}>
           {tokens.map((line, i) => (
             <Line key={i} {...getLineProps({ line, key: i })}>
-              <LineNo>{i + 1}</LineNo>
+              {language === 'json' || <LineNo>{i + 1}</LineNo>}
```
jkaster, line 98: Doesn't need to happen right now but this display decision could be made outside of the loop through each line

packages/code-display/src/components/index.ts
```diff
@@ -0,0 +1,25 @@
...
+
+ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+
+ */
```
jkaster, line 25: Is anything exported here?

packages/code-display/src/test-data/specs.ts
```diff
@@ -0,0 +1,48 @@
...
+    version: '3.1',
+  },
+  '4.0': {
+    key: '4.0',
+    isDefault: true,
+    status: 'experimental',
+    specURL: 'https://self-signed.looker.com:19999/api/4.0/swagger.json',
+    specContent: require('../../../../spec/Looker.4.0.oas.json'),
+    version: '4.0',
+  },
```
jkaster, line 44: FWIW the contents of these specs change, which could possibly affect tests that check specific values.

jkaster (commented):
Seems there's a file that's only a comment header?

packages/code-display/src/components/index.ts
noahamac: Removed

packages/code-display/src/test-data/specs.ts
noahamac: Removed. This was vestigial of an old test. Is resolved as I import { codeGenerators } from @looker/sdk-codegen now instead of hard coding the spec

packages/code-display/src/CodeDisplay.tsx
noahamac: Will address in follow up ticket. Will make ticket

jkaster (approved):
Good progress!

packages/code-display/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "@looker/code-display",
+  "version": "0.0.1",
```
joeldodge79, line 3: I think this version is fine. But just FYI this package will stay at `0.0.n` until you commit a breaking change (e.g. `feat!:`, `fix!:`, etc) as both `fix` AND `feat` only ever increment the "patch" for pre-major packages.

And, if/when you want to graduate to say `0.1.0`, then you'll have to manually update dependents as release-please won't make that "pre-major dependency's breaking change bump"

joeldodge79 (commented):
my comments are just FYIs

PR #594, opened 2021-04-15
feat: adds formatted sdk example table

noahamac:
This PR replaces the previous tab based SDK Example section for a formatted table implementation. Although the design spec called for a DataTable, I did not find that to be a suitable implementation mainly due to lack of control over column width. 

Design spec:
image

Formatted table:
image

Data table:
image


noahamac:
I think something may be off with the font weights or font sizes. Pedantic input appreciated

packages/api-explorer/src/components/DocSdkUsage/DocSdkUsage.tsx
```diff
@@ -40,31 +48,59 @@ interface DocSdkUsageProps {
...
-              <DocExamples
-                lode={examples}
-                language={language}
-                operationId={method.operationId}
-              />
-            </TabPanel>
-          ))}
-        </TabPanels>
-      </>
+      <Card height="auto" p="medium" mt="small">
```
noahamac, line 63: Does this whole Card deserve a separate file?

packages/api-explorer/src/components/DocSdkUsage/DocSdkUsage.tsx
```diff
@@ -40,31 +48,59 @@ interface DocSdkUsageProps {
...
+            <Text fontWeight="semiBold">Language</Text>
+          </FlexItem>
+          <FlexItem flexBasis="10%">
+            <Text fontWeight="semiBold" style={{ float: 'right' }}>
+              Line
+            </Text>
+          </FlexItem>
+        </Flex>
+        {tableExamples.map((exampleRow: ExampleLinkTableRow, i: number) => (
+          <div key={i}>
```
josephaxisa, line 78: I believe the use of the `div` here is due to the need for a keyed fragment. If that's the case, I suggest using 
`<React.Fragment key={key}></React.Fragment>`. @bryans99 in case he has other opinions or flat out disagrees with me 😄 

I also get a bit nervous about using the index as a key after being bitten by it a few times. Perhaps it's not a big deal on non input components but you could consider using something more unique, such as `${exampleRow.filename}#${exampleRow.line}`.

packages/api-explorer/src/components/DocSdkUsage/DocSdkUsage.tsx
josephaxisa: I think this is fine as is

packages/api-explorer/src/components/DocSdkUsage/utils.ts
```diff
@@ -0,0 +1,80 @@
...
+        return {
+          filename: getFileNameFromPath(exampleLink.description),
+          line: getLineNumberFromLink(exampleLink.permalink),
+          language: language,
+          permalink: exampleLink.permalink,
+        }
+      }),
+    ]
+  }, [])
+}
```
josephaxisa, line 80: Let's add some tests for these utilities please

packages/api-explorer/src/components/DocSdkUsage/DocSdkUsage.tsx
```diff
@@ -40,31 +48,59 @@ interface DocSdkUsageProps {
...
-                operationId={method.operationId}
-              />
-            </TabPanel>
-          ))}
-        </TabPanels>
-      </>
+      <Card height="auto" p="medium" mt="small">
+        <Flex alignItems="center" mx="medium" minHeight="2rem">
+          <FlexItem flexBasis="50%">
+            <Text fontWeight="semiBold">Filename</Text>
```
josephaxisa, line 66: Maybe try `ApixHeading` here instead?

josephaxisa (changes requested):

packages/api-explorer/src/components/DocSdkUsage/DocSdkUsage.tsx
bryans99: Use React.Fragment rather than div. Use unique data from example row permalink, filename + line + column? Is that unique enough.

packages/api-explorer/src/components/DocSdkUsage/DocSdkUsage.tsx
```diff
@@ -40,31 +48,59 @@ interface DocSdkUsageProps {
...
+        </Flex>
+        {tableExamples.map((exampleRow: ExampleLinkTableRow, i: number) => (
+          <div key={i}>
+            <Divider appearance="light" />
+            <Flex alignItems="center" mx="medium" minHeight="2rem">
+              <FlexItem flexBasis="2%">
+                <Icon
+                  style={{ float: 'right' }}
+                  mr="small"
+                  name="IdeFileGeneric" // TODO use InsertDriveFile when available
```
bryans99, line 85: Comment in the middle of JSX?

packages/api-explorer/src/components/DocSdkUsage/DocSdkUsage.tsx
```diff
@@ -40,31 +48,59 @@ interface DocSdkUsageProps {
...
+              Line
+            </Text>
+          </FlexItem>
+        </Flex>
+        {tableExamples.map((exampleRow: ExampleLinkTableRow, i: number) => (
+          <div key={i}>
+            <Divider appearance="light" />
+            <Flex alignItems="center" mx="medium" minHeight="2rem">
+              <FlexItem flexBasis="2%">
+                <Icon
```
bryans99, line 82: Going to have some merge issues with this and the change I am working on

packages/api-explorer/src/components/DocSdkUsage/DocSdkUsage.tsx
bryans99: It's a bit big so consider creating two child components.

ExampleHeader which contains the code below until the map.
ExampleItem(?) which contains a single item. That would solve the div/React Fragment key issue

packages/api-explorer/src/components/DocSdkUsage/DocSdkUsage.tsx
noahamac: I was surprised Prettier let me do it... Will resolve in the merge commit

packages/api-explorer/src/components/DocSdkUsage/DocSdkUsage.tsx
```diff
@@ -40,31 +48,59 @@ interface DocSdkUsageProps {
...
-  const tabs = useTabs()
   const { examples } = useContext(LodeContext)
   const languages = findExampleLanguages(examples, method.name)
   if (languages.length === 0) return <></>
 
+  const tableExamples = prepareExampleTableData(
+    languages,
+    examples,
+    method.operationId
+  )
```
jkaster, line 59: This should be done using DataTable because of all the things we get from DataTable "for free." Replicating DataTable's functionality for sorting, pagination, filtering, and searching is not reasonable. The layout portion of the examples table is the most trivial part.

jkaster (changes requested):
Please change to using DataTable as the base so we can enhance this examples display for the other things it needs in future revisions.

PR #595, opened 2021-04-15
feat: prettier loading scene while fetching specs

josephaxisa:
image


jkaster:
IMO this is a feature not a chore. Updated description.

jkaster (approved):
Looks much better

PR #597, opened 2021-04-15
fix: Typescript OAuth test

jkaster:
Typescript OAuth test got broken on main probably due to dependency update


bryans99 (approved):
LGTM but I'm not in a state to test it.

Me thinks the sdk-node refactor broke this?

PR #601, opened 2021-04-19
fix: error handling for TS SDK, GSon body serialization for Kotlin SDK

### LINKED ISSUE #544: Cannot create User / Group Attributes with Kotlin SDK if User Attribute is not hidden
With the Kotlin SDK, running against Looker 7.20.12, user attribute operations fail if the user attribute is not hidden.

For `create_user_attribute()` for example, this works...
>  WriteUserAttribute(name=git_password, label=Git Password, type=string, default_value=, value_is_hidden=**true**, user_can_view=true, user_can_edit=true, hidden_value_domain_whitelist=https://foobar.com/*)

… but this fails with a 422…
> WriteUserAttribute(name=git_username, label=Git Username, type=string, default_value=, value_is_hidden=**false**, user_can_view=true, user_can_edit=true, hidden_value_domain_whitelist=null)

Similiar for `set_user_attribute_group_values`, this works...   
> [UserAttributeGroupValue(can=null, id=null, group_id=4, user_attribute_id=38, value_is_hidden=**true**, rank=null, value=1.2.3.4)]

... but this fails with a 500...
> [UserAttributeGroupValue(can=null, id=null, group_id=4, user_attribute_id=28, value_is_hidden=**false**, rank=null, value=1.2.3.4)]

Testing with the API, I think the issue may be caused by the SDK always sending the `hidden_value_domain_whitelist`, even if the value is not hidden.

API Docs:
* [Create User Attribute](https://docs.looker.com/reference/api-and-integration/api-reference/v4.0/user-attribute#create_user_attribute)
* [Set User Attribute Group Values](https://docs.looker.com/reference/api-and-integration/api-reference/v4.0/user-attribute#set_user_attribute_group_values)

jkaster:
We have an outstanding feature request to improve analysis of Kotlin SDK errors, but this documents the encountered error in the Typescript SDK test, which has better error handling

API Explorer and the existing Looker API Docs UI can also be used to see what the actual 422 error is

Fixes #544 🦕

### DIFF
--- a/kotlin/src/main/com/looker/rtl/Transport.kt
+++ b/kotlin/src/main/com/looker/rtl/Transport.kt
@@ -24,12 +24,12 @@
 
 package com.looker.rtl
 
+import com.google.gson.Gson
 import io.ktor.client.HttpClient
 import io.ktor.client.call.receive
 import io.ktor.client.engine.okhttp.OkHttp
 import io.ktor.client.features.json.JacksonSerializer
 import io.ktor.client.features.json.JsonFeature
-import io.ktor.client.features.json.defaultSerializer
 import io.ktor.client.request.HttpRequestBuilder
 import io.ktor.client.request.forms.FormDataContent
 import io.ktor.client.request.request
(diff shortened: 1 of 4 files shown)


joeldodge79 (changes requested):
I'm kind of feeling like this should be two separate PRs for the sake of cleaner changelog entries? Granted right now we're not maintaining a kotlin changelog (hopefully will be soon). As written, the `packages/sdk/CHANGELOG.md` entry will be:

```markdown
### Fixes
* error handling for TS SDK, GSon body serialization for Kotlin SDK
```

But if you're ok with that then we can merge as-is.

jkaster:
The PR was based on analyzing and understanding what was happening with the endpoint. I could make separate PRs but since the changelog won't get processed for the Kotlin SDK right now I'd rather not spend the time.

joeldodge79 (approved):

PR #602, opened 2021-04-19
feat: adds 'SDK Examples' data table

noahamac:
Continuation of: https://github.com/looker-open-source/sdk-codegen/pull/594

Design spec:
image

Implementation:
image

With paginate (current "examples per page" set to 10):
image


packages/api-explorer/src/components/DocSdkUsage/DocSdkUsage.spec.tsx
```diff
@@ -0,0 +1,83 @@
...
+      getFileNameFromPath('/sdk-codegen/packages/foo.py line 200')
+    ).toEqual('foo.py')
+  })
+
+  test('can get line number from permalink', () => {
+    expect(
+      getLineNumberFromLink(
+        'https://github.com/looker-open-source/sdk-codegen/blob/ed8cf9e72fabff3266edb5235448a4a27ce0b1c7/packages/sdk-node/test/methods.spec.ts#L498'
+      )
+    ).toEqual('498')
```
jkaster, line 81: You shouldn't need to parse the line number from the link .. that's provided in the example index? Why is this needed?

packages/api-explorer/src/components/DocSdkUsage/utils.ts
```diff
@@ -0,0 +1,179 @@
...
+}
+
+/**
+ * Gets the line number from the end of a Github permalink
+ * @param link - github permalink
+ * @returns A stringified line number of the link
+ */
+export function getLineNumberFromLink(link: string): string {
+  return link.split('#L')[1]
+}
```
jkaster, line 85: Why do you need to re-extract the line number from the existing data?

jkaster (commented):
Just a question for now.

noahamac:
@jkaster The `line` is not directly returned from the `findExamples` call:
https://github.com/looker-open-source/sdk-codegen/blob/730de4ea6f80a2e90ec2e8f7222e6b3c8b54dbca/packages/sdk-codegen/src/exampleInfo.ts#L163

Should I add it? Would eliminate the need for the `getLineNumberFromLink` function to your point

jkaster:
> Should I add it? Would eliminate the need for the `getLineNumberFromLink` function to your point

IMO, better to add that return value than re-parse.

jkaster (approved):
Looks good

PR #604, opened 2021-04-20
chore: Change extension sdk react to use peer dependencies.

bryans99:


jkaster (approved):

packages/extension-sdk-react/package.json
```diff
@@ -44,8 +44,10 @@
     "@looker/extension-sdk": "^21.4.1",
     "@looker/sdk": "^21.4.1",
     "@looker/sdk-rtl": "^21.0.12",
+    "lodash": "^4.17.20"
+  },
+  "peerDependencies": {
     "history": "^4.9.0",
```
mdodgelooker, line 50: Does `history` need to be in `peerDependencies`? It doesn't seem like extension developers would usually need to install it alongside this package like they would the react packages.

The common pattern is to put the react packages in `devDependencies` as well, for jest, etc to work. But since this is in a monorepo where react is required in `dependencies` elsewhere, jest will still work without doing that. So I guess it would be a purely symbolic gesture. 🙂

packages/extension-sdk-react/package.json
```diff
@@ -44,8 +44,10 @@
...
-    "lodash": "^4.17.20",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
```
mdodgelooker, line 52: Would it be possible to require a slightly lower minimum version? In components we do:
```
    "react": "^16.11",
    "react-dom": "^16.11",
```

mdodgelooker (changes requested):
Thanks for doing this @bryans99!

mdodgelooker (approved):
Awesome!

packages/extension-sdk-react/package.json
jkaster: Should we downgrade all our other React packages in this repository to these versions?

mdodgelooker: No, I don't think so. In components we actually put a [higher version in `devDependencies`](https://github.com/looker-open-source/components/blob/main/packages/components/package.json#L58) just to stay as up-to-date as possible. But having a lower version in `peerDependencies` just means that as long as the developer installs that version or higher alongside this package, they're good to go. (And they won't get a pesky duplicate React warning.)

jkaster: Thanks for explaining that!

bryans99: So is it okay to have a different version in dev dependencies for this eaxmple?

mdodgelooker (dismissed):
👍

mdodgelooker (dismissed):
👍

mdodgelooker (approved):

PR #609, opened 2021-04-21
fix: looker component version merge issue

bryans99:


jkaster (approved):
Thanks for the fix

noahamac:
Thanks for doing that

PR #613, opened 2021-04-21
chore: reset release-please release-as config

joeldodge79:
Part of our "non-Looker-release" release cycle needs to be resetting
"release-as" keys so we can see what release-please would propose
normally based purely on conventional commits and THEN we can
make the necessary "release-as" adjustments


bryans99 (approved):
LGTM but what do I know :D . Rubber stamping as per request

PR #615, opened 2021-04-21
feat: adds editor functionality to CodeDisplay

noahamac:
Replaces `AceEditor` with `@looker/code-editor` for all code display within RunIt and ApiExplorer.

Search pattern marking:
image

Editor:
image

Response body:
image

Code tab:
image

Type declaration:
image


packages/run-it/src/components/ShowResponse/responseUtils.tsx
```diff
@@ -113,7 +116,7 @@ const ShowImage = (response: IRawResponse) => {
 
 /** A handler for HTTP type responses */
 const ShowHTML = (response: IRawResponse) => (
-  <CodeDisplay language={'html'} code={response.body.toString()} />
+  <CodeDisplay language={'html'} code={response.body.toString()} transparent />
```
josephaxisa, line 119:
```suggestion
  <CodeDisplay language="html" code={response.body.toString()} transparent />
```

yarn.lock
```diff
@@ -2193,6 +2193,15 @@
     debug "^2.2.0"
     es6-promise "^4.2.8"
 
+"@looker/components-providers@^0.9.29":
```
bryans99, line 2196: Hm, how'd this get pulled in?

yarn.lock
```diff
@@ -2205,11 +2214,38 @@
...
+  resolved "https://registry.yarnpkg.com/@looker/components-test-utils/-/components-test-utils-0.9.29.tgz#5e61437806c9a26c2d4a4759eb2ca406eee8e73d"
+  integrity sha512-m+JJ9rmvveWJRGF3vZRQo8cI35dTac1gW5pkEbs67pa5uIasyfTfI/kDWDiW7NbBClWX9aVlnpl3elsGnbt2Lw==
+
 "@looker/components-test-utils@^1.1.3":
   version "1.1.3"
   resolved "https://registry.yarnpkg.com/@looker/components-test-utils/-/components-test-utils-1.1.3.tgz#53ac11522cc7d485178c1a1d89f0751d8af88475"
   integrity sha512-z6fqkdTWw5YLcVd1FOSovl4y7W/Jsjc6GzA8sdeIMn9ghDMeH9GyGvoFIGV2tEQ/b6x1gZ+zDnKBSe8vUo4Fjg==
 
+"@looker/components@^0.9.30":
+  version "0.9.30"
```
bryans99, line 2228: Ditto

packages/code-display/src/CodeDisplay.tsx
```diff
@@ -112,4 +123,22 @@ export const CodeDisplay: FC<PrismEditorProps> = ({
       )}
     </Highlight>
   )
+  if (!onChange) {
+    return highlight(code)
```
josephaxisa, line 127: is this for when `CodeDisplay` is used as an editor? If so I think a comment would be handy

packages/code-display/src/CodeDisplay.tsx
```diff
@@ -36,20 +37,28 @@ import { getPrismLanguage, getOverridenTheme } from './utils'
...
   language?: string
   /** Code blob to be highlighted */
   code: string
   /** Pattern to be search (if applicable) */
   pattern?: string
+  /** onChange event handler, for Editor functionality */
+  onChange?: (text: string) => void
+  /** whether or not to render transparent background */
+  transparent?: boolean
+  height?: string
```
josephaxisa, line 51: Add a comment perhaps describing what happens if no height is specified? I don't see this prop being used anywhere

packages/code-display/src/CodeDisplay.tsx
josephaxisa, line 127: Also L117 can be modified to use the Span from @looker/components

packages/code-display/src/CodeDisplay.tsx
```diff
@@ -112,4 +123,22 @@ export const CodeDisplay: FC<PrismEditorProps> = ({
       )}
     </Highlight>
   )
+  if (!onChange) {
+    return highlight(code)
+  }
+  return (
+    <Editor
```
jkaster, line 130: I think we should have separate components for Display and Editing. I'm presuming `Editor` is heavier than `Highlight`

jkaster (changes requested):
Make CodeDisplay and CodeEditor distinct components

packages/code-display/src/CodeDisplay.tsx
noahamac: We can't use Span as it has default styles which conflict with Prism's styles applied onto a span

bryans99:
I removed a reference to an old looker components-test-utils from yarn.lock

jkaster:
Noah is working on completely removing references to Ace for this PR, so we'll hold off on approval for now

joeldodge79:
I wonder if we should just submit a separate PR to unbreak `main` first?

jkaster:
Is main still broken?

packages/code-editor/src/CodeDisplay.tsx
```diff
@@ -74,22 +79,21 @@ const LineContent = styled(Span)`
...
 }) => {
-  const highlighterLang = getPrismLanguage(language)
   return (
     <Highlight
       {...defaultProps}
       code={code}
-      language={highlighterLang}
-      theme={getOverridenTheme()}
+      language={getPrismLanguage(language)}
+      theme={getOverridenTheme(transparent)}
```
jkaster, line 96: I don't know what the thinking is on calling util functions like this in the markdown. Probably fine. I typically establish the values as consts (or use state if they can change from user interaction) but defer to @bryans99 or @josephaxisa on this topic

packages/code-editor/src/utils/utils.ts
```diff
@@ -57,7 +57,13 @@ export const getPrismLanguage = (language: string): Language => {
  * applies package overrides to the default vsCode prism theme
  * @returns modified prism theme object
  */
-export function getOverridenTheme() {
-  prismTheme.plain.backgroundColor = theme.colors.text
+export function getOverridenTheme(transparent: boolean) {
+  if (!transparent) {
```
jkaster, line 61: hugely minor point: typically you put the positive test first, then the negative goes in the else. Some linters will recommend that.

yarn.lock
```diff
@@ -3470,35 +3470,6 @@ accepts@~1.3.4, accepts@~1.3.5, accepts@~1.3.7:
     mime-types "~2.1.24"
     negotiator "0.6.2"
 
-ace-builds@^1.4.11, ace-builds@^1.4.6:
-  version "1.4.12"
-  resolved "https://registry.yarnpkg.com/ace-builds/-/ace-builds-1.4.12.tgz#888efa386e36f4345f40b5233fcc4fe4c588fae7"
-  integrity sha512-G+chJctFPiiLGvs3+/Mly3apXTcfgE45dT5yp12BcWZ1kUs+gm0qd3/fv4gsz6fVag4mM0moHVpjHDIgph6Psg==
```
jkaster, line 3476: Look at those glorious deletions!

jkaster (dismissed):
LGTM. Nice to bid Ace adieu!

packages/code-editor/src/CodeDisplay.tsx
noahamac: Yeah, curious to know the standard on this.

jkaster (approved):

noahamac:
What a ride. Thanks for the help, everyone :smile_cat:

PR #618, opened 2021-04-21
chore: updated mined examples

jkaster:
With the removal of the python hackathon app, we need to update the mined examples index and a test for the number of examples found


bryans99 (approved):
LGTM

PR #620, opened 2021-04-21
chore: publish workflows go automatically now

joeldodge79:
prior to this change the lerna publish workflow could only be triggered
by a manual run. Now it will go as soon as release-please creates the
release (which happens right after the Release PR is merged)

### DIFF
--- a/.github/workflows/lerna-publish.yml
+++ b/.github/workflows/lerna-publish.yml
@@ -1,6 +1,9 @@
 name: Publish NPM Packages
 
 on:
+  release:
+    types: published
+
   workflow_dispatch:
     inputs:
       ref:
@@ -8,7 +11,12 @@ on:
         required: true
 
 jobs:
-  release:
+  publish:
+    if: >
+      (github.event.inputs) ||
+      (startsWith(github.event.release.tag_name, 'sdk-codegen-all') &&
+      !github.event.release.draft &&
+      !github.event.release.prerelease)
     runs-on: ubuntu-latest
     env:
       LOOKERSDK_BASE_URL: https://localhost:20000
@@ -20,7 +28,7 @@ jobs:
         uses: actions/checkout@v2
         with:
           persist-credentials: false
-          ref: ${{ github.event.inputs.ref }}
+          ref: ${{  github.event.release.tag_name || github.event.inputs.ref }}
       - uses: actions/setup-node@v2
         with:
           node-version: 15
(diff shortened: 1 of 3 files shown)


.github/workflows/lerna-publish.yml
```diff
@@ -1,14 +1,22 @@
...
       ref:
         description: 'commit/tag/branch to release from'
         required: true
 
 jobs:
-  release:
+  publish:
+    if: >
+      (github.event.inputs) ||
+      (startsWith(github.event.release.tag_name, 'sdk-codegen-all') &&
```
joeldodge79, line 17: using `sdk-codegen-all` tag prefix because it will always be present when a Release PR is merged. It points to the same commit that all the other release tags point to but if we chose `sdk` or `sdk-rtl` etc, any one of those might not show up in a given Release PR if that package had no changes for that release

.github/workflows/lerna-publish.yml
```diff
@@ -1,14 +1,22 @@
...
+      !github.event.release.draft &&
+      !github.event.release.prerelease)
```
joeldodge79, line 19: this whole block is copied from the python publish: it just means "either use the form input from the manual workflow run OR use the tag as long as it's not draft or pre-release"

.github/workflows/python-publish.yml
```diff
@@ -1,18 +1,12 @@
 name: Python Publish to PYPI
 on:
-  # Automatic release based on either creating a published release or
-  # publishing a draft release (creating a draft release will not trigger the
-  # "release" event type according to the docs).
-  # The workflow file version used is that from the tag corresponding to the
-  # release. This is why the "workflow_dispatch" trigger is configured as well:
-  # in case we need to manually re-publish a tag using a modified workflow
   release:
-    types: [created, published]
```
joeldodge79, line 10: decided Python publish is safe enough now to just trigger automatically after a Release PR merge

.github/workflows/lerna-publish.yml
joeldodge79: it also means that now that `lerna publish` will run automatically after the Release PR is merged - I think we're ready for that

release-please-config.json
```diff
@@ -36,7 +36,6 @@
     },
     "python": {
       "release-type": "python",
-      "draft": true,
```
joeldodge79, line 39: actually this is what was preventing the python automatic publishing (I was manually editing the release to "publish" it). But again, ready for primetime

jkaster (approved):
Exciting times!

PR #622, opened 2021-04-22
fix: removes deprecated object attribute reference

noahamac:
Usage of the deprecated `label` attribute in these tests were causing the latest Api Explorer build to fail


jkaster (approved):
Good for fixing these tests, but of course we still need to fix the extension bundling script. Separate PR for that.

PR #623, opened 2021-04-22
fix: use "import type" in the Typescript SDK

jkaster:
Updated the code generator and the TS SDK to use `import type` where appropriate


joeldodge79:
Looks like this has been around since TS 3.8 - but I do have a question around what versions of typescript we're committed to supporting in this SDK?

Also, I'd consider switching to a `fix:` or `feat:` prefix - just so it shows up in the changelog so that maybe an old TS 3.7 project that breaks might more easily find a clue as to why. But if you don't want it showing up in the changelog, then seems like `refactor:` is a more human-reading-git-log meaningful prefix.

joeldodge79:
CI is wonky - it's reporting the "skipped" runs (from the cla-bot applying "cla: yes" tag) instead of the real runs: you can see both on the checks tab (CI did pass). Need to figure that out

jkaster:
Good question about supported TS versions. I'd say since we're just now approaching GA, TS 4.x is our minimum

jkaster:
We can consider it a "fix" although nothing was broken. It just wasn't optimal.

joeldodge79 (approved):

AprilArcus:
@joeldodge79 I agree; the minimum version we support in the GA should be the current version. That's TS 4.2 is current, as of today, with TS 4.3 scheduled for May 25.

At minimum, TS 4.1 has some extremely powerful new features that I would not want to be without in the long term.

jkaster:
We currently build codegen with TS 4.2.2, so I'm personally fine with saying 4.2 and above

PR #627, opened 2021-04-23
feat: add noComment toggle to code generators

### LINKED ISSUE #626: Add a toggle for including comments in codegen
Add a `noComment` property to `ICodeGen`, primarily for use in API Explorer. 
`true` omits comments from the generated output.
`false` is the default, and retains the current behavior

jkaster:
`noComment` defaults to `false`, retaining current generator behavior
`noComment = true` omits comments from code generation.
This option was primarily added for usage with API Explorer

Fixes #626 🦕


bryans99 (approved):
LGTM

PR #629, opened 2021-04-26
chore: release-please set versions

joeldodge79:
temporarily set changelog-sections on packages/sdk to work around
c7b5bf81 which should have been a 'fix:' commit

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -7,7 +7,11 @@
     "packages/extension-sdk": {},
     "packages/hackathon": {},
     "packages/sdk-node": {},
-    "packages/sdk": {},
+    "packages/sdk": {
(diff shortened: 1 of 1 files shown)


release-please-config.json
```diff
@@ -7,7 +7,11 @@
     "packages/extension-sdk": {},
     "packages/hackathon": {},
     "packages/sdk-node": {},
-    "packages/sdk": {},
+    "packages/sdk": {
+      "changelog-sections": [
+        {"type": "chore", "section": "Bug Fixes"}
```
joeldodge79, line 12: this "might" need to have a `"hidden": false` but I think that's the default so we'll see

jkaster (approved):

PR #632, opened 2021-04-27
fix: added a missing "import type" statement for the Typescript SDK

jkaster:


joeldodge79 (approved):
might be good to capture some of this `import type` stuff in typescript.gen.spec.ts if it doesn't already exist

also, keep an eye on https://github.com/looker-open-source/sdk-codegen/pull/632/checks to make sure the non-skipped checks pass

jkaster:
> might be good to capture some of this `import type` stuff in typescript.gen.spec.ts if it doesn't already exist

Won't be necessary when we add the eslint rule for the entire repository, which is what we should do.

PR #636, opened 2021-04-27
feat: APIX now uses google font appropriately

bryans99:
APIX will use google fonts when running as a non embedded extension OR, in stand alone mode, when running from a looker.com or google.com domain.

Also
1. adds extension-api-explorer package to apix tests script.
2. fixes looker design token dependencies. missing or wrong version from some packages. specifies an explicit version as an alpha version is the current version in npm.

### DIFF
--- a/packages/code-editor/package.json
+++ b/packages/code-editor/package.json
@@ -40,7 +40,7 @@
   },
   "dependencies": {
     "@looker/components": "^1.1.3",
-    "@looker/design-tokens": "^1.1.3",
+    "@looker/design-tokens": "1.1.3",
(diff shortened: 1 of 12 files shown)


packages/code-editor/package.json
```diff
@@ -40,7 +40,7 @@
   },
   "dependencies": {
     "@looker/components": "^1.1.3",
-    "@looker/design-tokens": "^1.1.3",
+    "@looker/design-tokens": "1.1.3",
```
jkaster, line 43: Good to note the pin

jkaster (approved):
LGTM

PR #652, opened 2021-04-30
chore: use looker-open-source-automation-bot token for release-please

joeldodge79:
better to stick with an action than switch to the app:
1. app we cannot see logs nor execution success/failure
2. configuration is more transparent through action than app


bryans99 (approved):
LGTM but what do I know :D

jkaster (approved):
Looks good as far as I can tell!

PR #653, opened 2021-05-01
chore: CI overhaul

joeldodge79:
- get rid of CI:TEST/HOLD label logic
-- this was causing problems: when other labels were applied
   (e.g. "cla: yes") then the workflow would report as skipped
   and it appeared that all checks passed.
-- replacement: go to the actions UI, and for each CI workflow trigger
   it on your branch

- add "noop" duplicate/dummy jobs to support enabling "required checks"
  on PRs so that we no longer accidentally merge broken changes.
  See code comments and
  https://github.community/t/github-actions-and-required-checks-in-a-monorepo-using-paths-to-limit-execution/16586/7?u=joeldodge79

This will allow us to enable the following required checks on PRs
* Python Test Results
* Typescript Test Results
* Codegen Test Results
* cla/google
* Semantic Pull Request

### DIFF
--- a/.github/workflows/codegen-ci.yml
+++ b/.github/workflows/codegen-ci.yml
@@ -1,13 +1,6 @@
 name: Codegen CI
 on:
   pull_request:
-    types:
-      - opened
-      - synchronize
-      - reopened
-      # first three are default, adding labeled for "CI:TEST" support
-      # (see jobs.unit.if below)
-      - labeled
     paths:
       - packages/sdk-codegen/**
       - packages/sdk-codegen-utils/**
@@ -21,6 +14,8 @@ on:
       - packages/sdk-codegen-utils/**
       - packages/sdk-codegen-scripts/**
 
+  workflow_dispatch:
+
 env:
   LOOKERSDK_BASE_URL: https://localhost:20000
   LOOKERSDK_VERIFY_SSL: false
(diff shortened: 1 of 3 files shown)


.github/workflows/codegen-ci.yml
```diff
@@ -1,13 +1,6 @@
 name: Codegen CI
 on:
   pull_request:
-    types:
-      - opened
-      - synchronize
-      - reopened
-      # first three are default, adding labeled for "CI:TEST" support
-      # (see jobs.unit.if below)
-      - labeled
```
joeldodge79, line 10: now that we're not checking `labeled`, we can just rely on the default types

.github/workflows/codegen-ci.yml
```diff
@@ -120,7 +98,7 @@ jobs:
           path: results/sdk-codegen
 
   publish-test-results:
-    name: Publish Test Results
+    name: Codegen Test Results
```
joeldodge79, line 101: Purposely named these `publish-test-results` jobs distinctly across workflow files so that that we can duplicate them as dummy-success jobs within the same file to satisfy the silly github shortfall regarding required checks (see code comment below and PR description for more detail)

.github/workflows/codegen-ci.yml
```diff
@@ -132,31 +110,33 @@ jobs:
...
       - name: Publish Unit Test Results
-        uses: EnricoMi/publish-unit-test-result-action@v1.6
+        uses: EnricoMi/publish-unit-test-result-action@v1.12
         with:
-          check_name: Test Results
+          check_name: Codegen Tests
           github_token: ${{ secrets.GITHUB_TOKEN }}
           report_individual_runs: true
+          hide_comments: orphaned commits
+          check_run_annotations_branch: '*'
```
joeldodge79, line 119: tweaking a couple new knobs (upgraded from 1.6 to 1.12) that will give us better signal on test results

jkaster (approved):
This looks cleaner than what we had before

PR #657, opened 2021-05-03
fix: improve TS SDK imports/exports

jkaster:


joeldodge79 (approved):
quick note: `refactor:` is not user facing. And because there are no other user facing commits since the last release for this package, release-please will not generate a Release PR for this yet. That's ok if you think there will be other user-facing commits on this package before you need to release. However, if this is the only change you want to release, then in needs to be user facing (`fix:` or `feat:`)

jkaster:
I'll make it `fix:` then because it does fix an import issue

PR #660, opened 2021-05-04
ci: noop CI for otherwise orphaned PRs

joeldodge79:


joeldodge79:
example of a candidate PR: https://github.com/looker-open-source/sdk-codegen/pull/596

That PR only touches a file under `docs/` and thus does not trigger any of our existing CI workflows and so some required checks are showing up pending... forever. My hope is that I can manually run this "noop CI" workflow on that branch and that will cause the required checks to show up green

josephaxisa (approved):

PR #667, opened 2021-05-05
chore: remove the Looker Release version from "do not edit" warning

jkaster:
This is only a change to a code comment header and does not modify the SDK behavior in any way


bryans99 (approved):
LGTM - this will make this kind of review easier in the future :D

PR #672, opened 2021-05-06
chore: release-please config reset and npm publish workflow back to node 15 for npmcli v7

joeldodge79:


.github/workflows/lerna-publish.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           ref: ${{  github.event.release.tag_name || github.event.inputs.ref }}
       - uses: actions/setup-node@v2
         with:
-          node-version: 14
+          node-version: 15
```
joeldodge79, line 34: this will get us back on npm cli v7 which works with @bryans99 build fix

bryans99 (approved):
LGTM but we have larger issues to solbe

PR #674, opened 2021-05-11
docs: Typescript and Python README updates for GA

joeldodge79:
removes "beta" language


jkaster:
I'd like to modify the TS generation to do correct docstring comment headers anyway so maybe we can bundle those two together for an updated deployment

joeldodge79:
> maybe we can bundle those two together for an updated deployment

sounds good: make sure that's a `fix:` commit so it triggers a version bump (since this commit is a `docs:` and thus will not, by itself, trigger a version bump)

jkaster (approved):

PR #678, opened 2021-05-12
fix: font download on route change

bryans99:
For both standalone and extensions. Favicon is still downloaded on route change for standalone.


jkaster (approved):
Looks like a better abstraction, too

PR #679, opened 2021-05-12
chore: release-please config fix

joeldodge79:
the `minor-pre-major` flags should be top level defaults
code-editor was missing them causing incorrect version
bumping for that package


bryans99 (approved):
LGTM but I'm not an expert as yet. Perhaps wait for John's review as well.

PR #680, opened 2021-05-12
ci: centralize noop checks

joeldodge79:
The benefit of having them duplicated in each workflow
file was the ability to delay their execution with the
`needs: [publish-test-results]` syntax. However, there
was buggy behavior where the python-ci workflow would
get attributed the actual jobs from the tssdk-ci workflow
and its own noop jobs would somehow fail with the
opaque `No such file or directory` error

The downside of centralizing them is losing the ability to
specify that `needs` syntax. I believe that means they
will "pass" immediately and there will be a window of
several minutes until the "real" jobs report where the PR
will satisfy all the required checks and could be
mergeable. If that is the case, I may consider hacking it
to do `sleep 300 && exit 0` which would introduce an
artifical delay that may or may not help avoid the issue

### DIFF
--- a/.github/workflows/noop-ci.yml
+++ b/.github/workflows/noop-ci.yml
@@ -1,45 +1,43 @@
+# The jobs in this workflow are necessary because they are configured as
+# required checks on PRs. Each job actually has a corresponding "real" job
+# definition in a different workflow file:
+#
+# codegen-ci.yml: Codegen Tests
+# tssdk-ci.yml: Typescript Tests
+# python-ci.yml: Python Tests
+#
+# By hard-coding passing jobs with the same name here every PR will satisfy the
+# repo's "required checks" EXCEPT those that trigger the "real" job AND that
+# "real" job fails.
+#
+# It will look a little confusing: any PR that triggers a corresponding "real"
+# job will have duplicate check entries, one that passes (this noop job) and the
+# "real" job that either passes or fails.
 name: Noop CI
 on:
-  # This workflow is intended to be manually run on a branch for which the
-  # PR triggers no other CI workflow (and thus cannot trigger any of the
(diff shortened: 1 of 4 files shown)


.github/workflows/noop-ci.yml
```diff
@@ -1,45 +1,43 @@
...
+# It will look a little confusing: any PR that triggers a corresponding "real"
+# job will have duplicate check entries, one that passes (this noop job) and the
+# "real" job that either passes or fails.
 name: Noop CI
 on:
-  # This workflow is intended to be manually run on a branch for which the
-  # PR triggers no other CI workflow (and thus cannot trigger any of the
-  # required jobs). Eg. docs, examples, new packages that don't have a CI
-  # workflow setup yet.
-  workflow_dispatch:
```
joeldodge79, line 7: this was a failed experiment: according to the docs I should have been able to create a "check-run" on a commit that is linked to an open PR and have that check register on the PR itself. I even verified that the checks I created had associated `pull_request` objects but the GitHub UI never showed the checks :-( bogus advertising

.github/workflows/noop-ci.yml
```diff
@@ -1,45 +1,43 @@
...
     steps:
-      - uses: octokit/request-action@v2.x
-        name: Noop Codegen Tests
-        with:
-          route: POST /repos/{owner}/{repo}/check-runs
-          owner: looker-open-source
-          repo: sdk-codegen
-          name: Codegen Tests
-          head_sha: ${{ github.sha }}
-          conclusion: success
```
joeldodge79, line 21: cool action to keep in mind: let's you execute REST calls directly

jkaster (approved):

PR #681, opened 2021-05-13
feat: upgrade react-markdown

bryans99:
upgrade required for the following reasons
1. dependabot reported a security issue with a package (trim) the older version of react-markdown was using. Note that the trim packages is still referenced by eslint tooling. This will be dealt with in a separate PR.
2. dependency on path-browserify and process-browser can be removed.

The replacement was not straight forward as the mechanism for providing alternative rendering of components significantly changed. The Markdown component was refactored into its own component as react-markdown was being used in two places. It turns out that this was not necessary as the 2nd usage needs to use the raw react-markdown component at least for. The refactor was kept as it is possible the markdown component may be useful elsewhere. The original implementation had a dependency on specKey which has now been removed.

In addition the following changes have been done:
1. Links to looker documentation in an extension now supported.
2. List components render correctly.
3. Links that have markup from search now work.


packages/api-explorer/src/components/Markdown/Markdown.tsx
```diff
@@ -0,0 +1,144 @@
...
+    if (ele.parentElement) {
+      return findAnchor(ele.parentElement)
+    }
+    return undefined
+  }
+
+  const handleClick = (e: BaseSyntheticEvent) => {
+    e.preventDefault()
+    if (linkClickHandler) {
+      // Could be a <mark> tag wrapped by an anchor
```
noahamac, line 68: I like this new file a lot. 

Similar to how anchors have a special case,
code blobs have a currently uncaught special case. It looks like your change makes this easier to address:
image

josephaxisa (changes requested):
It seems that highlighting is not working as expected everywhere:

image

bryans99:
@josephaxisa - that's not an issue introduced by my change. Also happens in the current marketplace version. Suspect it's going to be tricky to solve as I think code is similar to pre in that it ignores child elements. Create a separate issue is my recommendation.

Marketplace APIX version below

image

josephaxisa (dismissed):
Caught up about this in person. Issues mentioned above are better suited for another PR. This lgtm. Thanks Bryn!

packages/api-explorer/src/components/DocMarkdown/utils.ts
```diff
@@ -70,8 +52,9 @@ export const remapHashURL = (specKey: string, url: string) =>
 /**
  * Clean urls from mark tags and remap tag/method hashbang urls to match the MethodScene route
  * @param specKey A string to identify the spec in the URL
- * @param url
+ * @param url - url
```
jkaster, line 55: What does this comment mean?

packages/api-explorer/src/components/Markdown/Markdown.spec.tsx
```diff
@@ -0,0 +1,100 @@
...
+ */
+
+import React from 'react'
+import { screen } from '@testing-library/react'
+import { renderWithTheme } from '@looker/components-test-utils'
+
+import { renderWithSearch } from '../../test-utils'
+import { Markdown } from './Markdown'
+
+describe('Markdown', () => {
```
jkaster, line 34: ooh. Tests!

packages/api-explorer/src/components/Markdown/Markdown.tsx
jkaster: Agreed. I think this file belongs in code-editor rather than here, so we can reuse it outside of APIX. I also hope to convert triple tick code blocks to use code display

jkaster (approved):
This is a great improvement. Very minor nit on a code comment, but LGTM.

PR #683, opened 2021-05-13
chore: Upgrade Looker eslint and prettier config

bryans99:
Knock on effects:
1. eslint needed to be upgraded
2. prettier needed to be upgraded
3. new rules required quite a lot of code changes :(

In addition removed prettier and eslint configs from access-token-server package.json as not needed there.


packages/api-explorer/src/components/DocMethodSummary/utils.tsx
```diff
@@ -27,7 +27,7 @@ import React from 'react'
 import styled from 'styled-components'
 import { Icon, Status } from '@looker/components'
 import { Beaker } from '@looker/icons'
-import { VisibilityOff } from '@styled-icons/material'
+import { VisibilityOff } from '@styled-icons/material/VisibilityOff'
```
jkaster, line 30: Interesting the deeper import is required. Did you fix this with `--fix`?

jkaster (approved):
LGTM. Other than the deeper import link (which I don't really care about), I like these prettier reformatting choices better.

packages/api-explorer/src/components/DocMethodSummary/utils.tsx
bryans99: Unfortunately not :( . That one was a bit painful

jkaster: Ouch indeed

josephaxisa (approved):
lgtm

PR #685, opened 2021-05-14
chore: better TS SDK function header documentation

jkaster:
Applied more TSDoc tags to functions, methods, streams, and method interface declarations

Fixing the undesired stripping of a trailing blank line from comment headers had a bit of a ripple effect into some files unrelated to TypeScript


joeldodge79 (dismissed):
couple of thoughts for consideration about how this ends up in various changelogs:

merged as-is, the PR title will show up in both TS and Python changelogs (same would be true for kotlin/go/cs/swift if they were setup with release-please)

Other options:
1. s/`feat`/`chore`/ - while it is a "codegen" feature, I wouldn't consider it a feature of the SDKs
2. break out separate PRs
- `feat: better function header formatting` for codegen changes
- `chore: re-formatted doc strings` for generated code changes

But I'll leave it up to you if you want to wrangle git. The actual code changes LGTM

jkaster:
Happy to make it a chore to simplify things. Just realized a minor perf tweak I want to have anyway for `no comment` so I'll do that first. I was also interested in the phrasing I chose to auto-generate. @josephaxisa @bryans99 do you have any opinions on the phrasing?

joeldodge79 (approved):
👍

PR #688, opened 2021-05-26
chore: update to the latest Looker components

jkaster:
Also updated related packages


packages/api-explorer/src/components/Markdown/common.tsx
```diff
@@ -43,14 +43,11 @@ MDCode.defaultProps = {
   pr: 'xxsmall',
 }
 
-// Temp fix for release build failure
-export const MDCodeBlock = styled(CodeBlock)`
+export const MDCodeBlock = styled(CodeBlock).attrs(
+  ({ fontSize = 'small', mb = 'large' }) => ({ fontSize, mb })
```
noahamac, line 47: Is this a new recommended pattern?

lukelooker: Definitely within the @looker/components code-base it is :)

See: https://dev.to/layershifter/how-to-kill-tree-shaking-in-webpack-with-static-properties-p72

noahamac: Thanks for the link!

noahamac (approved):

packages/api-explorer/src/components/Markdown/common.tsx
jkaster: I guess, because @lukelooker recommended it!

lukelooker: I'm pretty sure that answer is almost as good as "I saw it on the internet so it _must_ be true" 😄

bryans99 (approved):
LGTM as well

PR #702, opened 2021-06-16
chore: release-please config tweaks

joeldodge79:
- alphabetize for easier comparison with .release-please-manifest.json
- inline package definitions for easer GitHub PR diff viewing
- hardcode some versions for next release


jkaster (approved):
LGTM. I like the more compact display, too

PR #703, opened 2021-06-16
fix: python remove runtime asserts

### LINKED ISSUE #695: `query_task_results` method not permitting bytes type response
**Expected**: Permit bytes type response for `query_task_results` method. We were trying to use async run to generate `.xlsx` files and ran into the assertion error for the method's response.

**Full Review**:
There are 100 other occurrences of this assertion pattern (`assert isinstance(response, str)`) in the SDK. I did not go through them as our fledgling implementation of the python SDK is pretty low-coverage and does not utilize most of the methods in these files.

**Proposed Fix:**
I am under time constraints to get work done so I have done what I see is the "minimum" required in this [draft PR](https://github.com/looker-open-source/sdk-codegen/pull/694). If significant additional work is needed to merge this then I may be delayed in responding due to current work demands. If the draft PR is sufficient or needs limited changes I will do those and whatever is required to formally request its review.

joeldodge79:
let's keep the type checking out of the runtime. We may end up returning
values that are not true to the annotated type but we'll call that a
feature not a bug for a language like Python that isn't truly statically
typed.

fixes #695


joeldodge79:
oops, this needs work as the `assert` was narrowing the return type.

jkaster (approved):
LGTM

PR #705, opened 2021-06-16
feat: Looker 21.8 bindings

joeldodge79:


python/looker_sdk/sdk/api31/methods.py
```diff
@@ -83,7 +83,6 @@ def login(
             query_params={"client_id": client_id, "client_secret": client_secret},
             transport_options=transport_options,
         )
-        assert isinstance(response, models.AccessToken)
```
joeldodge79, line 86: oops, left over from another branch I'm working on: forgot to `yarn clean && yarn build`

packages/sdk/src/3.1/models.ts
```diff
@@ -2477,14 +2477,6 @@ export enum Format {
...
 }
 
-/**
- * Scheme that is running on application server (for PRs, file browsing, etc.) Valid values are: "http", "https".
- */
-export enum GitApplicationServerHttpScheme {
-  http = 'http',
-  https = 'https',
-}
-
```
joeldodge79, line 2487: @jkaster this feels like borderline breaking for 3.1 - what do you think?

jkaster: This enum was removed because it was not accurate. Sometimes `null` or an empty string would be provided instead, so it broke the Kotlin SDK for example. It was wrong for both 3.1 and 4.0, so I'd consider it a bug fix?

jkaster (approved):
LGTM

PR #711, opened 2021-06-17
chore: lerna publish workflow to use yarn install/build

joeldodge79:


joeldodge79:
oops, I think this is the branch we used to fix the publish problems last time and I just totally dropped the ball and forgot to get it merged in.

PR #718, opened 2021-06-28
feat: syntax highlight application/sql responses

jkaster:
- also changed to `ProgressCircular` "processing" component

- fixed text color on RunIt response progress indicator


noahamac (approved):
Nice!

jkaster:
> Nice!

@noahamac `CodeDisplay` makes it so darned easy!

PR #719, opened 2021-06-29
feat: adds embed layout variant

noahamac:
Removes the header and moves its configuration options to sidenav when hostType is 'embed'.

When embed in dev portal:
image


jkaster:
This is a great end result but could it be abstracted/componentized more?

packages/api-explorer/src/ApiExplorer.tsx
```diff
@@ -113,16 +113,24 @@ const ApiExplorer: FC<ApiExplorerProps> = ({
               value={{ searchSettings, setSearchSettings }}
             >
               <Page style={{ overflow: 'hidden' }}>
-                <Header
-                  specs={specs}
-                  spec={spec}
-                  specDispatch={specDispatch}
-                  toggleNavigation={toggleNavigation}
-                />
+                {envAdaptor.runAsEmbed() || (
```
bryans99, line 116: Is this a similar look to the standalone version? If standalone should use the same technique. Rather than using `runAsEmbed` change to headess. This should be a property as you only need to check it once and the property (a boolean can be passed in).

packages/api-explorer/src/ApiExplorer.tsx
```diff
@@ -113,16 +113,24 @@ const ApiExplorer: FC<ApiExplorerProps> = ({
...
+                    specDispatch={specDispatch}
+                    toggleNavigation={toggleNavigation}
+                  />
+                )}
                 <Layout hasAside height="100%">
                   {hasNavigation && (
                     <AsideBorder pt="large" width="20rem">
-                      <SideNav api={spec.api} specKey={spec.key} />
+                      <SideNav
+                        api={spec.api}
```
bryans99, line 128: You've added logic to get the context to SideNav. Simplify things and pass  the flag in as a property as you already have it.

Again, avoid the term runAsEmbed. Suggest headless but other terms maybe appropriate.

packages/api-explorer/src/components/SideNav/SideNav.tsx
```diff
@@ -58,14 +73,21 @@ interface SideNavParams {
...
+  specDispatch,
+}) => {
   const tabNames = ['methods', 'types']
   const match = useRouteMatch<SideNavParams>(`/:specKey/:sideNavTab?`)
   let defaultIndex = tabNames.indexOf('methods')
   if (match && match.params.sideNavTab) {
     defaultIndex = tabNames.indexOf(match.params.sideNavTab)
   }
   const tabs = useTabs({ defaultIndex })
+  const { envAdaptor } = useContext(EnvAdaptorContext)
```
bryans99, line 90: Just pass in a property

packages/extension-api-explorer/src/utils.ts
```diff
@@ -58,6 +58,13 @@ export class ExtensionEnvAdaptor implements IApixEnvAdaptor {
     return this._themeOverrides
   }
 
+  runAsEmbed(): boolean {
```
bryans99, line 61: Change to property and set in constructor

PR #724, opened 2021-06-29
feat(APIError): New Python error class giving access to API response

### LINKED ISSUE #723: Python: Make Looker endpoint error response more accessible
I would like to be able to inspect the Looker API response data to better understand the type of error I have got when performing an action. Knowing that information I can then go on to perform a different set of actions after the error.

For example the `create_user` can have a variety of [responses](https://docs.looker.com/reference/api-and-integration/api-reference/v3.1/user#response_messages_3) like 'Bad Request', 'Validation Error' etc. but in order to get that information my exception handling has to do the following:

```
try:
    """
    do something in Looker SDK
    """
except looker_error.SDKError as exc:
    error_data = json.loads(exc.args[0])
    """
    parse error_data to take different actions depending on response
    """
```

My suggestion would be to have a new error class that inherits from the existing `SDKError`, which takes the response as the first argument making it accessible as a property.

```
class APIError(SDKError):
    """API error class
    """
    def __init__(self, sdk_response, *args, **kwargs):
        super(APIError).__init__(*args, **kwargs)
        self._sdk_response = json.loads(sdk_response)
    
    @property
    def sdk_response(self):
        return self._sdk_response
```

Then updating the `APIMethods._return` [call](https://github.com/looker-open-source/sdk-codegen/blob/main/python/looker_sdk/rtl/api_methods.py#L87) raise this error instead.

```
        encoding = response.encoding
        if not response.ok:
            raise error.APIError(response.value.decode(encoding=encoding))
        ret: TReturn
        if structure is None:
```

This implementation would mean that all existing handling of SDKErrors would continue to work, and those who want to take a richer set of handling would be able to do so! 🎆

ob1-dev:
This change introduces a new `APIError` class (inheriting from SDKError) which takes one argument of the `api_response` which should be the JSON payload from the Looker endpoint.

This gets loaded by the class and made accessible as the `APIError.api_response` property.

The idea is to give simpler access to the type of error that was received from the endpoint in order to take different actions.

- [X] Make sure to open an issue as a [bug/issue](https://github.com/looker-open-source/sdk-codegen/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea

Fixes #723 🦕

### DIFF
--- a/python/looker_sdk/error.py
+++ b/python/looker_sdk/error.py
@@ -22,8 +22,20 @@
 
 """API error class
 """
+import json
 
 
 class SDKError(Exception):
+    """SDK error class
+    """
+
+class APIError(SDKError):
     """API error class
     """
+    def __init__(self, sdk_response, *args, **kwargs):
+        super(APIError).__init__(*args, **kwargs)
+        self._sdk_response = json.loads(sdk_response)
+    
+    @property
+    def sdk_response(self) -> dict:
+        return self._sdk_response
--- a/python/looker_sdk/error.py
+++ b/python/looker_sdk/error.py
@@ -32,10 +32,10 @@ class SDKError(Exception):
 class APIError(SDKError):
     """API error class
     """
-    def __init__(self, sdk_response: str, *args, **kwargs):
+    def __init__(self, api_response: str, *args, **kwargs):
         super(APIError).__init__(*args, **kwargs)
-        self._sdk_response = json.loads(sdk_response)
+        self._api_response = json.loads(api_response)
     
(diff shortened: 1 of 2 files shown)


ob1-dev:
@googlebot I signed it!

jkaster:
Thanks for your submission. We've implemented the functionality in a similar PR. #1170

jkaster:
There wasn't anything wrong with your PR, but this was something we already had in process. Thanks again for your contribution. You definitely saw the need!

PR #728, opened 2021-07-01
fix: paging parsing issues

annguy3n:
This PR fixes 2 issues:

1. raw response from [parse](https://github.com/looker-open-source/sdk-codegen/blob/main/packages/sdk-rtl/src/paging.ts#L366) is a relative path, when used with `BaselessSettings` and `credentials: same-origin` directly from the TS sdk.
2. total count header appears lowercase `x-total-count`


packages/sdk-rtl/src/paging.ts
```diff
@@ -34,7 +34,7 @@ import { IAPIMethods } from './apiMethods'
 import { BaseTransport } from './baseTransport'
 
 export const LinkHeader = 'link'
-export const TotalCountHeader = 'X-Total-Count'
```
joeldodge79, line 37: I feel like I see the capitalized form of http headers more often: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

also, server/client should consider them case-insensitive: https://stackoverflow.com/a/5259004/13802304

packages/sdk-rtl/src/paging.ts
```diff
@@ -363,7 +363,7 @@ export class Paging<TSuccess extends ILength, TError>
   }
 
   parse(raw: IRawResponse): IPager<TSuccess, TError> {
-    const req = new URL(raw.url)
+    const req = new URL(raw.url, 'https://localhost')
```
joeldodge79, line 366: hm, do we have access to the settings `base_url` here?

packages/sdk-rtl/src/paging.ts
jkaster: Yeah, it's a problem with the headers handling in JS. I responded via chat that I think we need to support both the Title case and lowercase at least.

packages/sdk-rtl/src/paging.ts
jkaster: Exactly my thought, as well. Sorry, I should have responded to the PR rather than via chat.

packages/sdk-rtl/src/paging.ts
```diff
@@ -362,21 +362,28 @@ export class Paging<TSuccess extends ILength, TError>
     return result
   }
 
+  static findHeader(raw: IRawResponse, name: string) {
+    return (
+      raw.headers[name] ||
+      raw.headers[name.toLowerCase()] ||
+      raw.headers[name.toUpperCase()]
```
jkaster, line 369: Recognizes all three reasonable casing options for the header

packages/sdk-rtl/src/paging.ts
```diff
@@ -362,21 +362,28 @@ export class Paging<TSuccess extends ILength, TError>
...
+    )
+  }
+
   parse(raw: IRawResponse): IPager<TSuccess, TError> {
-    const req = new URL(raw.url)
-    const params = req.searchParams
+    const params = new URLSearchParams(raw.url)
```
jkaster, line 374: Handles full or relative urls

jkaster (approved):
LGTM

PR #734, opened 2021-07-02
fix: python datetime [de]serialization bug

### LINKED ISSUE #720: Python: Synchronize Looker objects and json
**Issue**: The SDK Python will return looks, schedule plans etc, ... as JSON-dict alike objects. While we could update the objects using Python's syntax for dict() without issues, we can't use the object in the parameter for the update call. Instead, we have to use an additional "Write" method from `sdk.models`. 

**Ideal workflow**: Synchronize Looker objects and dict()/json() so that we could update() without having to take the additional steps. 

Here is an article on the same issue with [copying look](https://community.looker.com/looker-api-77/using-looker-api-to-copy-looks-23069). I'll demonstrate an example with update a schedule plan using the SDK: 

```
# Get info about schedule_plans for this look: 
look = sdk.scheduled_plans_for_look(321, all_users=True)

# Update 'enabled' = False -- for simplicity, I'll update the first schedule plan of this look (schedule_plan_id = 662)
data = look[0]
data['enabled'] = False 
```

Then I call update_schedule_plan(), and have this issue `TypeError TypeError: Object of type datetime is not JSON serializable` 

```
# Update Schedule
sdk.update_scheduled_plan(662,body=look)
```

Taking a look at type(look): It is a ScheduledPlan object and contains datetime for created_at (return body is abbreviated) 

```
<class 'looker_sdk.sdk.api40.models.ScheduledPlan'>
ScheduledPlan(name='Mail', created_at=datetime.datetime(2021, 6, 26, 15, 12, 38, 458000, tzinfo=datetime.timezone.utc), updated_at=datetime.datetime(2021, 6, 26, 16, 39, 8, 662000, tzinfo=datetime.timezone.utc))
```

Tagging @joeldodge79 -- thank you for taking a look!

joeldodge79:
we'd forgotten to register a `cattr.register_unstructure_hook` for `datetime.datetime` objects.
Clearly no one had been trying to send in any data that included a datetime.datetime till now

closes #720

### DIFF
--- a/python/looker_sdk/rtl/serialize.py
+++ b/python/looker_sdk/rtl/serialize.py
@@ -144,6 +144,7 @@ def unstructure_hook(api_model):
     return data
 
 
+DATETIME_FMT = "%Y-%m-%dT%H:%M:%S.%f%z"
 if sys.version_info < (3, 7):
     from dateutil import parser
 
@@ -158,9 +159,10 @@ else:
     def datetime_structure_hook(
         d: str, t: Type[datetime.datetime]
     ) -> datetime.datetime:
-        return datetime.datetime.strptime(d, "%Y-%m-%dT%H:%M:%S.%f%z")
+        return datetime.datetime.strptime(d, DATETIME_FMT)
 
 
 converter31.register_structure_hook(datetime.datetime, datetime_structure_hook)
 converter40.register_structure_hook(datetime.datetime, datetime_structure_hook)
 cattr.register_unstructure_hook(model.Model, unstructure_hook)  # type: ignore
+cattr.register_unstructure_hook(datetime.datetime, lambda dt: dt.strftime(DATETIME_FMT))
--- a/python/tests/rtl/test_serialize.py
+++ b/python/tests/rtl/test_serialize.py
@@ -21,6 +21,7 @@
 # THE SOFTWARE.
 
 import copy
+import datetime
 import enum
 import functools
 import json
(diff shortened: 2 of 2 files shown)


python/looker_sdk/rtl/serialize.py
```diff
@@ -158,9 +159,10 @@ def datetime_structure_hook(
...
         d: str, t: Type[datetime.datetime]
     ) -> datetime.datetime:
-        return datetime.datetime.strptime(d, "%Y-%m-%dT%H:%M:%S.%f%z")
+        return datetime.datetime.strptime(d, DATETIME_FMT)
 
 
 converter31.register_structure_hook(datetime.datetime, datetime_structure_hook)
 converter40.register_structure_hook(datetime.datetime, datetime_structure_hook)
 cattr.register_unstructure_hook(model.Model, unstructure_hook)  # type: ignore
+cattr.register_unstructure_hook(datetime.datetime, lambda dt: dt.strftime(DATETIME_FMT))
```
joeldodge79, line 168: this is the whole fix

python/tests/rtl/test_serialize.py
```diff
@@ -99,6 +99,7 @@ class Model(ml.Model):
     # standard types
     id: Optional[int] = None
     name: Optional[str] = None
+    datetime_field: Optional[datetime.datetime] = None
```
joeldodge79, line 102: adding this `datetime_field` everywhere in the tests identically reproduced the issue in #720

python/looker_sdk/rtl/serialize.py
```diff
@@ -144,6 +144,7 @@ def unstructure_hook(api_model):
     return data
 
 
+DATETIME_FMT = "%Y-%m-%dT%H:%M:%S.%f%z"
```
jkaster, line 147: Love the infinite syntax variation for datetime string formatting

jkaster (approved):
Good fix

PR #740, opened 2021-07-07
fix: extension-api-explorer build

bryans99:


jkaster (approved):
Thanks. Dunno how I missed this. I thought I tested apix as well

bryans99:
> Thanks. Dunno how I missed this. I thought I tested apix as well

It was the bundle that failed. We should probably add that to CI even if its throwaway. I'll look into it as part of the standalone build for Noah

PR #741, opened 2021-07-07
fix: add a test to additional query param required bug

annguy3n:
Address query params parsing bug for pagination. See tests for bug info.


jkaster (approved):
Thanks!

PR #745, opened 2021-07-09
fix: browserTransport.ts retains ok: false now

jkaster:
Also improved error handling for both browser and node transports


bryans99 (approved):
LGTM

PR #751, opened 2021-07-12
feat: apix sidebar toggle for dev portal

bryans99:
# Description

This change allows an entity other than APIX explorer to hide/show the APIX sidebar.

# Changes

The dev portal navigation bar is independent of the APIX nav bar. A previous change, allowed the APIX nav bar to be hidden when APIX is running in the dev portal. The issue with that change  was that the toggle to open/close the sidebar was no longer visible and the toggle in the dev portal had no way to communicate with the APIX explorer so that it could hide/show the side bar. This change adds a window message listener to APIXExplorer when it operates in headless mode. When the window message is received it toggles the sidebar visibility.

# Screen shots

image

image

### DIFF
--- a/packages/api-explorer/src/ApiExplorer.tsx
+++ b/packages/api-explorer/src/ApiExplorer.tsx
@@ -24,7 +24,14 @@
 
  */
 
-import React, { FC, useReducer, useState, useEffect } from 'react'
+import React, {
+  FC,
+  useReducer,
+  useState,
+  useEffect,
+  useRef,
+  useCallback,
+} from 'react'
 import { useLocation } from 'react-router'
 import styled, { createGlobalStyle } from 'styled-components'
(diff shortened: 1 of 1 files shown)


packages/api-explorer/src/ApiExplorer.tsx
```diff
@@ -85,6 +92,31 @@ const ApiExplorer: FC<ApiExplorerProps> = ({
   const toggleNavigation = (target?: boolean) =>
     setHasNavigation(target || !hasNavigation)
 
+  const hasNavigationRef = useRef<boolean>(hasNavigation)
```
jkaster, line 95: aha. a `useRef` ref!

jkaster (approved):
Glad you did this bit and not me

PR #752, opened 2021-07-13
feat: simplify apix dev portal navigation toggle

bryans99:
While walking the dog I realized that it was not necessary to use useRef to hold the current value of the toggle. The alternative is to pass a function to setState that accepts the current value. This change does that.


josephaxisa (approved):
lgtm

jkaster (approved):
Nice cleanup

PR #754, opened 2021-07-15
fix: bullet proof APIX response handling

bryans99:
# Description

APIX now renders raw data if it cannot parse the response returned as expected.

 # Changes

1. Added try/catches in appropriate places. Note that the try/catch in RunIt is probably NOT necessary but the code being wrapped indicates that an error could be thrown (logically it should never happen as the user should always be authenticated).
2. Renamed `parseJson` method to `json2Csv` to better reflect what it does (I was initially confused by the original method name).
3. Created a ShowRaw response handler.
4. Created a fallbackResponseHandler to use ShowRaw.

# Screen Shots

image

### DIFF
--- a/packages/run-it/src/components/ShowResponse/ShowResponse.spec.tsx
+++ b/packages/run-it/src/components/ShowResponse/ShowResponse.spec.tsx
@@ -35,6 +35,7 @@ import {
   testJsonResponse,
   testTextResponse,
   testUnknownResponse,
+  testBogusJsonResponse,
 } from '../../test-data'
 import { ShowResponse } from './ShowResponse'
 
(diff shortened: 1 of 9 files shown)


packages/run-it/src/components/ShowResponse/ShowResponse.spec.tsx
```diff
@@ -97,4 +98,15 @@ describe('ShowResponse', () => {
...
+
+  test('it renders bogus json responses', () => {
+    renderWithTheme(<ShowResponse response={testBogusJsonResponse} />)
+    expect(screen.getByText('200: application/json')).toBeInTheDocument()
+    expect(
+      screen.getByText(
+        'The response body could not be parsed. Displaying raw data.'
+      )
+    ).toBeInTheDocument()
+    expect(screen.getByText('I AM A LYING JSON RESPONSE')).toBeInTheDocument()
```
jkaster, line 110: YES YOU ARE!

packages/run-it/src/components/ShowResponse/responseUtils.tsx
```diff
@@ -141,6 +137,21 @@ const ShowPDF = (response: IRawResponse) => {
   return ShowUnknown(response)
 }
 
+/** A handler for responses that cannot be parsed */
+const ShowRaw = (response: IRawResponse) => (
```
jkaster, line 141: _The door has been opened!_

jkaster (approved):
This looks great! Very clean implementation. Thank you.

PR #761, opened 2021-07-20
fix: uses consistent heading text

noahamac:
As per Chris' suggestion

https://github.com/llooker/developer-portal/pull/43#issuecomment-883550714


noahamac:
image

bryans99 (approved):
LGTM

PR #762, opened 2021-07-21
chore: Add nix present check to .envrc

bryans99:
I was making a similar change in the `embed-sdk` repo. It was suggested I test for nix before using nix to be nice to non nix users. Propagating that fix here.

Note, you will need to run `direnv allow` again.


jkaster (approved):
Nice fix

PR #766, opened 2021-07-26
fix: improved writeable type support

jkaster:
- A warning in the description of the type was also added for types with no writeable properties that should have writeable properties

- regenerated against Looker 21.10


packages/sdk-codegen/src/sdkModels.spec.ts
```diff
@@ -366,6 +366,48 @@ describe('sdkModels', () => {
       ])
     })
 
+    it('Setting', () => {
```
josephaxisa, line 369: may I suggest changing the test names to be more descriptive? This test file is quite big and it would make it easier in case anyone needs to venture here again. Feel free to ignore these suggestions, the tests been fine as is but in my experience it's sometimes difficult to find what applies if you don't know what you're looking for

```suggestion
    it('generates writeable type for nested types with some readonly properties', () => {
```

packages/sdk-codegen/src/sdkModels.spec.ts
```diff
@@ -366,6 +366,48 @@ describe('sdkModels', () => {
...
+      // Code concession for linting that doesn't understand "toBeDefined()" test
+      if (type) {
+        expect(type.name).toEqual('WriteSetting')
+        const writeWhite = type.properties.whitelabel
+        expect(writeWhite).toBeDefined()
+        expect(writeWhite.type.name).toEqual('WriteWhitelabelConfiguration')
+      }
+    })
+
+    it('WriteMergeQuery', () => {
```
josephaxisa, line 387:
```suggestion
    it('generates and annotates a writeable type collection ', () => {
```

packages/sdk-codegen/src/sdkModels.spec.ts
```diff
@@ -366,6 +366,48 @@ describe('sdkModels', () => {
...
+      const source = apiTestModel.types.MergeQuery
+      const actual = apiTestModel.mayGetWriteableType(source)
+      const props = actual?.properties
+      const sq = props?.source_queries
+      expect(sq).toBeDefined()
+      expect(sq?.type.name).toEqual('MergeQuerySourceQuery[]')
+    })
+
+    it('CreateDashboardRenderTask', () => {
```
josephaxisa, line 396:
```suggestion
    it('does not generate writeable type if type contains no readonly properties', () => {
```

packages/sdk-codegen/src/sdkModels.spec.ts
```diff
@@ -366,6 +366,48 @@ describe('sdkModels', () => {
...
+      const source = apiTestModel.types.CreateDashboardRenderTask
+      const actual = apiTestModel.mayGetWriteableType(source)
+      expect(actual).toBeUndefined()
+    })
+
+    it('ContentMetaGroupUser', () => {
```
josephaxisa, line 402:
```suggestion
    it('does not generate writeable type if type contains only readonly properties', () => {
```
or
```suggestion
    it('does not generate writeable type if type contains only writeable properties', () => {
```

josephaxisa (commented):
Thanks for looking into this so quickly, John. It looks good to me but I did leave some suggestions/questions below

joeldodge79 (commented):
This all seems fine to me (pending @josephaxisa's comments). My inline comment is just musings for a possible future way to handle input vs output

Also, I think the [particular genesis for this change is flawed](https://github.com/looker/helltool/pull/63754#discussion_r677509472) but the change itself is probably necessary because API authors will inevitably write input objects with nested mutability.

packages/sdk-codegen/src/sdkModels.ts
```diff
@@ -2218,18 +2240,41 @@ export class ApiModel implements ISymbolTable, IApiModel {
...
-    if (writes.length === 0 || writes.length === props.length) return undefined
+    if (writes.length === 0) {
+      // No writeable properties is an error
+      const immutable =
+        'WARNING: no writeable properties found for POST, PUT, or PATCH'
+      if (type.description.indexOf(immutable) < 0) {
+        type.description += type.description.length > 0 ? '\n' : '' + immutable
+      }
+      return undefined
+    }
```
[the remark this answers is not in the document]
jkaster: As discussed elsewhere, (also literally years ago now) we could implement a `strict` mode for the SDKs that would log an error (during run-time) for any read-only type violations when requests are sent to the endpoint. However, we still need to decide if that's a complete replacement for the writeable type behavior, or if we support both, or what.

packages/sdk-codegen/src/sdkModels.spec.ts
```diff
@@ -366,6 +366,48 @@ describe('sdkModels', () => {
...
+      const whitelabel = apiTestModel.types.WhitelabelConfiguration
+      const prop = setting.properties.whitelabel
+      expect(prop.type).toEqual(whitelabel)
+      expect(prop.readOnly).toEqual(false)
+      expect(whitelabel.readOnly).toEqual(false)
+      const type = apiTestModel.mayGetWriteableType(setting)
+      expect(type).toBeDefined()
+      // Code concession for linting that doesn't understand "toBeDefined()" test
+      if (type) {
+        expect(type.name).toEqual('WriteSetting')
```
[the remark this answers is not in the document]
jkaster: Yes, I just didn't want to do that for every single `type` reference that followed

packages/sdk-codegen/src/sdkModels.ts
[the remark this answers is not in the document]
jkaster: FWIW, I don't think these are edge cased in Writeable Types. I think it's a bug fix for type determination, and it points out a bug in some of our API types

joeldodge79 (approved):
👍

Issue #765, opened 2021-07-26
Golang - Unable to use SDK v4 due to unused import of `net/url`


Jberlinsky:
When using `github.com/looker-open-source/sdk-codegen/go@v0.0.2-0.20210707183859-6b3895d35e13`, importing `sdk/v4` yields an error due an an unused import of `net/url`:

```
❯ make
go build -o my-looker-client
# github.com/looker-open-source/sdk-codegen/go/sdk/v4
../../../../go/pkg/mod/github.com/looker-open-source/sdk-codegen/go@v0.0.2-0.20210707183859-6b3895d35e13/sdk/v4/models.go:39:3: imported and not used: "net/url"
make: *** [build] Error 2
make  19.85s user 5.40s system 399% cpu 6.313 total
```

At a quick glance, it does not appear that `net/url` (defined in https://github.com/looker-open-source/sdk-codegen/blob/main/go/sdk/v4/models.go#L39) is used. There are two potential remediations here:

1. If `net/url` does not need to be imported, remove the import line.
2. If there is a side effect from importing `net/url` that I'm not seeing, in accordance with the Golang [spec](https://golang.org/ref/spec#Import_declarations), the import should be declared as `_ "net/url"`):

> An import declaration declares a dependency relation between the importing and imported package. It is illegal for a package to import itself, directly or indirectly, or to directly import a package without referring to any of its exported identifiers. To import a package solely for its side-effects (initialization), use the blank identifier as explicit package name: `import _ "lib/math"`

joeldodge79:
looks like this became unused with the 21.8 release: https://github.com/looker-open-source/sdk-codegen/pull/705/files#diff-220fd72c15715632ea2366ad7271a8c4e86c578204627ad50ebdc82fbe31c645L87

PR #767, opened 2021-07-27
fix: remove unused net/url import from go

joeldodge79:
url.URL type properties were removed in v4 as of 21.8

fixes: #765


go/sdk/v4/models.go
```diff
@@ -36,7 +36,6 @@ package v4
 
 import (
   "github.com/looker-open-source/sdk-codegen/go/rtl"
-  "net/url"
```
joeldodge79, line 39: this is the direct cause of #765

package.json
```diff
@@ -43,8 +43,8 @@
     "lint-quiet:es": "eslint 'packages/**/*.ts{,x}' --cache --quiet",
     "lint:ts": "tsc",
     "lint-staged": "yarn exec lint-staged",
-    "lint-changed": "eslint $(git diff --name-only --diff-filter=ACMRTUXB main | grep  -E \"(.js$|.ts$|.tsx$)\") --quiet",
-    "lint-changed-fix": "eslint $(git diff --name-only --diff-filter=ACMRTUXB main | grep  -E \"(.js$|.ts$|.tsx$)\") --quiet --fix",
+    "lint-changed": "eslint $(git diff --name-only --diff-filter=ACMRTUXB origin/main | grep  -E \"(.js$|.ts$|.tsx$)\") --quiet",
+    "lint-changed-fix": "eslint $(git diff --name-only --diff-filter=ACMRTUXB origin/main | grep  -E \"(.js$|.ts$|.tsx$)\") --quiet --fix",
```
joeldodge79, line 47: unrelated but commit hooks failed and I had to `lint-changed-fix` and I never keep a local `main` branch around so here we are

packages/sdk-codegen/src/go.gen.spec.ts
```diff
@@ -34,6 +34,41 @@ const gen = new GoGen(apiTestModel)
 const indent = ''
 
 describe('Go generator', () => {
+  describe('bookends', () => {
+    it('has a v3 models prologue', () => {
+      const origPkgName = gen.packageName
+      gen.packageName = gen.apiVersion = 'v3'
+      const prologue = gen.modelsPrologue('')
+      gen.packageName = gen.apiVersion = origPkgName
```
joeldodge79, line 42: didn't really dig too deep into how `apiVersion` and `packageName` are set but this works

go/sdk/v4/models.go
jkaster: Ah. It doesn't load on demand like some of the other SDKs do. Good find/fix.

package.json
jkaster: I think this is a better statement for the script anyway

jkaster (approved):
It's a reasonable stopgap fix. When we get back to working on this generator we should fix its import logic.

PR #770, opened 2021-07-28
chore: release-please config

joeldodge79:
removing special handling for sdk-codegen* packages

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -12,11 +12,11 @@
     "packages/hackathon": { },
     "packages/run-it": { "release-as": "" },
     "packages/sdk": { },
-    "packages/sdk-codegen": { "release-as": "21.0.22" },
-    "packages/sdk-codegen-scripts": { "release-as": "21.0.23" },
-    "packages/sdk-codegen-utils": { "release-as": "21.0.12"},
+    "packages/sdk-codegen": { "release-as": "" },
+    "packages/sdk-codegen-scripts": { "release-as": "" },
+    "packages/sdk-codegen-utils": { "release-as": ""},
     "packages/sdk-node": { },
-    "packages/sdk-rtl": { "release-as": "21.0.18" },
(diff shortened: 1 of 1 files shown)


jkaster (approved):

release-please-config.json
```diff
@@ -12,11 +12,11 @@
...
     "packages/sdk": { },
-    "packages/sdk-codegen": { "release-as": "21.0.22" },
-    "packages/sdk-codegen-scripts": { "release-as": "21.0.23" },
-    "packages/sdk-codegen-utils": { "release-as": "21.0.12"},
+    "packages/sdk-codegen": { "release-as": "" },
+    "packages/sdk-codegen-scripts": { "release-as": "" },
+    "packages/sdk-codegen-utils": { "release-as": ""},
     "packages/sdk-node": { },
-    "packages/sdk-rtl": { "release-as": "21.0.18" },
+    "packages/sdk-rtl": { "release-as": "" },
```
joeldodge79, line 19: we've been keeping these at `21.0.n` by hand editing the release-please-config.json file between each release. While I do have an out-standing PR/feature request to make that less onerous, it's gonna be a while before or if it ever lands. Unfortunately, leaving an old `release-as` version here (one that was already released) wreaks havoc.

What I'm proposing now is that we just let these versions "fly away" and increment as they would normally via semver from here on out. This will avoid any nasty "old release-as" problems. But it also means that at some point we could have a `@looker/sdk-rtl` version 21.45.9 ... 

Thoughts?

jkaster: interesting. We've already documented that sdk-rtl doesn't follow the versioning of the language SDK packages, so I guess we just let semver do its thing?

jkaster (commented):
Trying to reduce release friction

release-please-config.json
joeldodge79: that would be my preference. especially if we think the package re-naming might happen soon-ish

PR #779, opened 2021-08-04
fix: fixes a golang leak on every REST request

panoskoug:
per Do documentation: "If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close."


joeldodge79 (approved):
great, thanks for the fix!

PR #780, opened 2021-08-04
fix: remove custom Exceptions and `assert` to make scripts executable at the simplest possible setup. 

lanlooker:
To address the question raised by our community on this error: `ModuleNotFoundError: No module named 'sdk_exceptions' `

Fixing: 
looker-open-source#756
https://community.looker.com/looker-api-77/looker-api-connection-test-26122


examples/python/README.md
```diff
@@ -4,6 +4,11 @@ You can find Python language examples in this folder.
 
 The full details of all Looker API endpoints are listed in Looker Docs: [Version 3.1](https://docs.looker.com/reference/api-and-integration/api-reference/v3.1), [Version 4.0](https://docs.looker.com/reference/api-and-integration/api-reference/v4.0)
 
+## Notes
+- Many example scripts include the following statement `import sdk_exceptions`, which refers to this file [sdk_exceptions.py](sdk_exceptions.py). You could either uncomment the statement or download `sdk_exceptions.py` and put it in the same folder with your current script. 
+
+
+
```
joeldodge79, line 11: interesting - it make sense that people are probably just copy/pasting an example or snippets from an example into their own pre-configured environment (where they maybe installed the other imports from pypi) rather than within the context of cloning this repo and executing the script with Pipfile.

@josephaxisa I'm inclined to say that no standalone example within this folder should import any other module in this folder in able to better support what is probably this common use-case. If you agree, @lanlooker instead of making this change could you delete `sdk_examples.py` and just copy/paste those error classes into each individual script? Or honestly, they could all be replaced with `raise Exception` (or maybe a more suitable stdlib one from https://docs.python.org/3/library/exceptions.html but don't spend more than a few minutes trying to find a replacement, just go with `Exception`). I don't think we need to be in the business of enforcing best patterns around specific exception handling with these examples.

The exception to the "don't import other /examples/python modules" might be if an example is a package (e.g. a folder like the current `/examples/python/{hackathon_app,lookersdk-flask} examples. Then folks should realize they need that whole folder/package structure

joeldodge79 (changes requested):

examples/python/README.md
josephaxisa: Agreed. I think making examples self contained makes perfect sense

lanlooker: I see, I'll review and make changes to all of the scripts currently using the sdk_exceptions (I think most of them were written by @josephaxisa, I'll tag you so you could review them before merging)

lanlooker:
Hi @joeldodge79  & @josephaxisa

Currently 5 of our code snippets are using `import sdk_exceptions`: 
- Download dashboard PDF
- Download look
- Run query Task
- Run Look with Applied Filters
- Test connection

What I did: 
- Replace all custom exceptions with `raise Exception("messages")`
- Deleted all `assert` per Google's Python style guideline: `Do not use assert statements for validating argument values of a public API.` (Some of the scripts did not work when there was `assert` -- perhaps related to datatype required between API 3.0 and API 4.0. I found a new bug that the SDK 4.0 requires `id` of a dashboard to be a string, so I removed `assert` and the script works) 
- Tested these scripts to make sure they work correctly: are using API 3.0, so I initialized 4.0, and test them on an internal instance
- Added docstrings at the beginning of the scripts: I moved and rewrote docstrings of the function main() to the beginning of the scripts. Many scripts place main() at the end, so I thought it may be difficult for developers to scroll through the entire scripts to read the docstrings

To-Dos (if I have the time): 
- Write docstrings for each function, and perhaps a short summary at the top of each scripts describing the API workflow (since all scripts here use multiple API endpoints)
- Refactor the scripts: Some scripts have inconsistent datatype for parameters (i.e.: `python3 scripts.py "Title" png` -- would be nice to clarify datatypes and make them all the same)
- Review the remaining snippets 

I am using Python's style guideline to review the code: https://google.github.io/styleguide/pyguide.html

Thank you! Please let me know if you would have further comments and suggestions for me!

examples/python/download_dashboard_pdf.py
```diff
@@ -59,10 +61,10 @@ def download_dashboard(
     width: int = 545,
     height: int = 842,
     filters: Optional[Dict[str, str]] = None,
-):
+    ):
+
     """Download specified dashboard as PDF"""
-    assert dashboard.id
-    id = int(dashboard.id)
+    id = dashboard.id
```
lanlooker, line 67: New issue here ... API 4.0 (SDK) requires sdk.create_dashboard_render_task(dashboard_id: str) (it was int according to this script)

examples/python/download_dashboard_pdf.py
```diff
@@ -103,4 +105,4 @@ def download_dashboard(
     print(f'Dashboard pdf saved to "{filename}"')
 
 
-main()
+main()
```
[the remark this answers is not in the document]
lanlooker: Thank you, I fixed the EOF character!

examples/python/download_dashboard_pdf.py
```diff
@@ -8,28 +21,18 @@
 import looker_sdk
 from looker_sdk import models
 
-import sdk_exceptions
-
-sdk = looker_sdk.init31("../../looker.ini")
+sdk = looker_sdk.init40(config_file='../../../looker.ini', section='Looker')
```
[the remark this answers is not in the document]
lanlooker: Thank you, I changed all scrips back to using this one --> `sdk = looker_sdk.init40("../looker.ini")`

(I was using one ini file for other projects so I had multiple sections in my ini file. I made a separate ini file in this example folders and gitignore it, so we should be good)

examples/python/download_dashboard_pdf.py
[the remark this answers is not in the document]
lanlooker: Fixed, and merged to the current PR

examples/python/download_dashboard_pdf.py
```diff
@@ -8,28 +21,18 @@
 import looker_sdk
 from looker_sdk import models
 
-import sdk_exceptions
-
-sdk = looker_sdk.init31("../../looker.ini")
+sdk = looker_sdk.init40("../looker.ini")
```
joeldodge79, line 24: oops, down one level too many
```suggestion
sdk = looker_sdk.init40("../../looker.ini")
```

examples/python/download_look.py
```diff
@@ -1,28 +1,32 @@
...
 import time
 
 import looker_sdk
 from looker_sdk import models
 
-import sdk_exceptions
-
-sdk = looker_sdk.init31("../../looker.ini")
-
+sdk = looker_sdk.init40("../looker.ini")
```
joeldodge79, line 20:
```suggestion
sdk = looker_sdk.init40("../../looker.ini")
```

examples/python/query_task.py
```diff
@@ -6,9 +16,7 @@
 import looker_sdk
 from looker_sdk import models40 as models
 
-import sdk_exceptions
-
-sdk = looker_sdk.init40("../../looker.ini")
+sdk = looker_sdk.init40("../looker.ini")
```
joeldodge79, line 19:
```suggestion
sdk = looker_sdk.init40("../../looker.ini")
```

examples/python/run_look_with_filters.py
```diff
@@ -1,31 +1,33 @@
...
 import sys
 from typing import cast, Dict, List, Union
 
 import looker_sdk
 from looker_sdk import models, error
 
-import sdk_exceptions
-
-sdk = looker_sdk.init31("../../looker.ini")
+sdk = looker_sdk.init40("../looker.ini")
```
joeldodge79, line 21:
```suggestion
sdk = looker_sdk.init40("../../looker.ini")
```

examples/python/test_connection.py
```diff
@@ -1,26 +1,31 @@
...
 from typing import cast, MutableSequence, Sequence
 
 import looker_sdk
 from looker_sdk import models
 
-import sdk_exceptions
-
-sdk = looker_sdk.init31("../../looker.ini")
-
+sdk = looker_sdk.init40("../looker.ini")
```
joeldodge79, line 20:
```suggestion
sdk = looker_sdk.init40("../../looker.ini")
```

joeldodge79 (changes requested):
I'll commit the suggestions and merge

joeldodge79 (approved):
awesome, thanks @lanlooker !

PR #785, opened 2021-08-05
fix: apix fix

bryans99:
APIX Extension was failing in latest master with invalid json. This was with the published extension AND the development version. Not sure why it suddenly started failing.


jkaster (approved):
Sorry. I broke that.

josephaxisa (approved):

PR #787, opened 2021-08-06
fix: remove python setup.py author_email

joeldodge79:
we no longer have support@looker.com and this field only accepts
a valid email address. sadly twine check does not catch this error
yet, hopefully it will some day.

Factored the twine check out of the unit matrix: it only needs to run once


jkaster (approved):
LGTM

PR #790, opened 2021-08-10
chore: end-to-end tests for the standalone api explorer

josephaxisa:
End to end tests for API Explorer. Use `yarn test:apix-e2e` from the root of the mono repo to run.


packages/api-explorer/e2e/e2e.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+    beforeEach(async () => {
+      await page.goto(BASE_URL)
+    })
+    it('renders a method page', async () => {
+      await expect(page).toClick('h5', { text: 'Dashboard' })
+      await expect(page).toClick('a', { text: 'Get All Dashboards' })
+      const specKey = await getSpecKey()
+      await expect(page.url()).toEqual(
+        `${BASE_URL}/${specKey}/methods/Dashboard/all_dashboards`
+      )
```
jkaster, line 52: With everything being async here is there a significant performance difference between puppeteer and cypress?

jkaster (commented):
Again, syntax is pretty easy to follow. You didn't need to create many helper functions, either. I just did an inline comment about performance differences between puppeteer and cypress so that would be good to know. And cypress is the mullti-browser option while puppeteer is Chrome/webkit?

packages/api-explorer/e2e/e2e.spec.ts
AprilArcus: if anything i'd be concerned about cypress using blocking I/O

jkaster (approved):
This is a great addition!

PR #791, opened 2021-08-10
fix: relax python dict key access

### LINKED ISSUE #778: Python: Unable to loop if key.value() is Null
I am working on a use case to remove the "All Users" group from a folder, and I am using `all_content_metadata_accesses`. 

In Looker, if a folder doesn't have any groups as its editor or viewer, then group_id is null (json), so I expect to loop and write some conditions with group_id, similar to this: 

```
testing_loop = [{'a':1, 'group_id': None }, {'a':1, 'group_id':1}]  
for i in testing_loop:
  if i['group_id'] is not None:
    print('group_id is not Null')
  else: 
    print('group_id is Null')
```

However, I am unable to loop or do anything with the key-value pair 'group_id' if the value is None because we have an exception handling a key with a None value: 

```
accesses = sdk.all_content_metadata_accesses(content_metadata_id=767)
test_dict = accesses[0] # Purposely choose accesses[0] because I know group_id here is Null 
print(type(test_dict))
print(test_dict['id']) # OK
print(test_dict['group_id']) # KeyError: 'group_id'
```

```
__getitem__(self, key)
     63             ret = ret.value
     64         if ret is None:
---> 65             raise KeyError(key)
     66         return ret
     67 

KeyError: 'group_id'
```

I'll post a reproduction case in our internal tool.

joeldodge79:
dictionary key access should reflect model property access
as closely as possible. In this case, do NOT raise a KeyError
for a model property that exists. https://git.io/JRrKm

fixes #778


jkaster (approved):
LGTM

PR #799, opened 2021-08-12
feat: first cut of the ResponseExplorer

jkaster:
Adds:
- header exploration
- collapsible body section

It's not quite final but it's ready for some feedback


packages/sdk-rtl/src/browserTransport.ts
```diff
@@ -201,6 +201,7 @@ export class BrowserTransport extends BaseTransport {
     }
     const headers = {}
     res.headers.forEach((value, key) => (headers[key] = value))
+    console.log({ headers })
```
josephaxisa, line 204: leftover debugging cruft?

packages/run-it/src/components/ResponseExplorer/ResponseExplorer.tsx
```diff
@@ -0,0 +1,145 @@
...
+interface ShowHeadersProps {
+  response: ResponseContent
+}
+
+export const ResponseHeaders: FC<ShowHeadersProps> = ({ response }) => {
+  const data = getHeaders(response)
+  if (data.length === 0) return <></>
+  const rows = data.slice(1)
+  return (
+    <>
```
josephaxisa, line 77: this fragment looks extra as well

packages/run-it/src/components/ResponseExplorer/ResponseExplorer.spec.tsx
```diff
@@ -0,0 +1,81 @@
...
+describe('ResponseExplorer', () => {
+  test('ShowHeaders', async () => {
+    renderWithTheme(<ResponseHeaders response={response} />)
+    const heading = 'Headers (3)'
+    const header = screen.getByText(heading)
+    expect(header).toBeInTheDocument()
+    await userEvent.click(header)
+    await waitFor(() => {
+      // TODO figure out a more specific selector?
+      const table = screen.getByRole('table')
```
josephaxisa, line 56: I assume you're saying this because there are (or there could be) multiple tables in the page. In that case you could use [`within`](https://testing-library.com/docs/dom-testing-library/api-within/).

Although it doesn't apply to this particular component, it's worth mentioning that there are places where it would also help to use the `data-testid` prop ([reference](https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change#whats-with-the-data-testid-query)). This would also make end to end tests easier.

josephaxisa (commented):
great progress! I couldn't get any response to render the headers section on my side though, probably not using the right endpoint

packages/run-it/src/components/ResponseExplorer/Collapser.tsx
```diff
@@ -0,0 +1,74 @@
...
+export const Collapser: FC<CollapserProps> = ({
+  id,
+  heading,
+  children,
+  defaultOpen = true,
+}) => {
+  const { value, toggle } = useToggle(defaultOpen)
+
+  return (
+    <>
```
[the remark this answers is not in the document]
jkaster: Removed that component entirely!

PR #805, opened 2021-08-31
ci: 21.14 image

joeldodge79:


.github/workflows/python-ci.yml
```diff
@@ -94,10 +94,10 @@ jobs:
         os:
           - ubuntu
         looker:
-          - '21_6'
           - '21_8'
```
jkaster, line 97: Isn't 21.6 still the extended support release?

.github/workflows/tssdk-ci.yml
```diff
@@ -98,9 +98,10 @@ jobs:
         os:
           - ubuntu
         looker:
-          - '21_6'
           - '21_8'
```
jkaster, line 101: 21_6 or 21_8?

jkaster (approved):
Looks good as long as 21.8 is the correct version instead of 21.6

.github/workflows/python-ci.yml
joeldodge79: oh interesting, I didn't realize the ESR pattern. According to https://docs.looker.com/relnotes/supported-releases 21.6 ESR support ends today (as does 21.8) so the list should actually only have 21_10, 21_12, 21_14

jkaster (approved):
GTG

PR #807, opened 2021-08-31
feat: Looker 21.14 release

joeldodge79:


python/looker_sdk/rtl/api_methods.py
```diff
@@ -44,7 +44,7 @@
         Sequence[model.Model],
     ]
 ]
-TStructure = Optional[Union[Type[str], serialize.TStructure]]
+TStructure = Optional[Union[Any, Type[str], serialize.TStructure]]
```
joeldodge79, line 47: the new `api_spec` method returns `Any` so needed to add that to the union type here

jkaster: Yeah, it's now (correctly) a JSON document

release-please-config.json
```diff
@@ -16,7 +17,7 @@
     "packages/sdk-codegen-scripts": { "release-as": "" },
     "packages/sdk-codegen-utils": { "release-as": ""},
     "packages/sdk-node": { },
-    "packages/sdk-rtl": { "release-as": "" },
+    "packages/sdk-rtl": { "release-as": "21.0.20" },
```
jkaster, line 20: I'm sure this was a manual fix. Maybe we should rethink this behavior for the rtl package.

jkaster (approved):
LGTM

release-please-config.json
joeldodge79: sorry, forgot to annotate this. tldr I'm back to desiring to keep these as `21.0.x` till we re-name and then just let it semver? see https://github.com/looker-open-source/sdk-codegen/pull/802#discussion_r699397245

PR #809, opened 2021-08-31
fix: placate TypeScript 4.4

AprilArcus:
TypeScript 4.4 assigns errors the type `unknown` by default in catch clauses. This PR supplies the necessary assertions to safely extract the `message` field in `browserTransport.ts`, and rethrows the error if it fails to meet these expectations.

### DIFF
--- a/packages/sdk-rtl/package.json
+++ b/packages/sdk-rtl/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@looker/sdk-rtl",
-  "version": "21.0.20",
+  "version": "21.0.19",
   "description": "Looker SDK Runtime Library",
   "main": "lib/index.js",
   "module": "lib/esm/index.js",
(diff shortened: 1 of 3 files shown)


jkaster:
@joeldodge79 adding you because of the version bump in the rtl package.json .. that may be the domain of release-please?

packages/sdk-rtl/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@looker/sdk-rtl",
-  "version": "21.0.19",
+  "version": "21.0.20",
```
joeldodge79, line 3: yep, release-please will handle this for us (based on the commit prefix of `fix:`, although doing this won't mess up release-please since it looks at `.release-please-manifest.json` as the source of truth and just overwrites this version number accordingly)

```suggestion
  "version": "21.0.19",
```

joeldodge79 (changes requested):

jkaster (approved):
LGTM

joeldodge79 (approved):

PR #817, opened 2021-09-14
chore: update lint config

jkaster:
`yarn fix` is currently borked


lukelooker:
Pulling myself off of the review since I made changes here.

.prettierrc.json
```diff
@@ -0,0 +1,5 @@
+{
```
lukelooker, line 1: @looker/eslint-config moving to _not_ including a default prettier and instead leveraging what's in the project specification (or prettier defaults if no per-repo config available)

bryans99 (approved):
LGTM

PR #818, opened 2021-09-14
docs: Simple example of creating a dashboard filter and applying it to dashboard elements

### LINKED ISSUE #816: Looker SDK (Python): Create example of how to create dashboard filter
This is a somewhat common ask for DCL. Creating the filter object is straightforward, but applying it to tiles is not. Having an example of what this looks like would be helpful for folks who are programmatically generating dashboards with filters.

dunnkevin:
Wanted to submit an example of creating a dashboard filter and listeners for dashboard tiles that update based on the filter.  This is based on a number of support requests that ask how to do this using Looker's API. We have some community posts about what endpoints to use, but no real working examples.

- 👇 Edit "Developer Checklist" to reflect items relevant to this PR (and try to make sure to check everything off before asking for review)

Fixes #816  🦕


examples/python/create_dashboard_filter.py
```diff
@@ -0,0 +1,88 @@
...
+    """Updates a dashboard element's result maker to include a listener on the new dashboard filter.
+
+
+    Args:
+        element (DashboardElement): Dashboard element to update with the new filter
+        filter (DashboardFilter): Dashboard filter the element will listen to
+    """
+    # Keep track of the current result_maker and add to it, otherwise listeners for other filters would be removed
+    current_filterables = element.result_maker.filterables
+    for filterable in current_filterables:
```
fabio-looker, line 65: This is surely largely to do with my inexperience with Python, but I'm confused why we are iterating through current filterables twice, once here and then once on line 75. I am sure your code is doing the right and necessary thing, but as far as readability, I've tried to figure it out a few times and I'm not 100% clear on this point. Either some additional comment, or combining these loops somehow might help? Not sure!

examples/python/create_dashboard_filter.py
```diff
@@ -0,0 +1,88 @@
...
+        new_listens = []
+        for each in filterable.listen:
+            new_listens.append(each)
+        # Add listener for new filter
```
fabio-looker, line 69: Is this where a customer might want to validate that the model/explore matches between the filter's field and the current query/filterable, before trying to add the listener? It seems like such a common thing to want to validate that maybe we should include that "if" statement as well, as a default piece of sample logic?

fabio-looker (commented):
I personally was a bit unclear on the purpose of some of the looping where I commented, which if that could be clarified in any way, would be a big win in my opinion. Other than that, they overall organization looks good to me.

examples/python/create_dashboard_filter.py
dunnkevin: Great catch. The code was doing the right thing, just not as efficiently as it could have. Originally when I drew up this process, I used one step to get the current filterables, and a second step to then update them. But, this isn't nearly as efficient as doing that all at once. Consolidated these two loops into one per your suggestion.

examples/python/create_dashboard_filter.py
dunnkevin: Great idea. Originally I thought this may be more complicated due to the possibility of a merged result tile, but since we're already looping through the filterables in order to take care of merge results, it's an easy addition!

dunnkevin:
@fabio-looker - Thanks for the original review, great suggestions. Added a couple small changes described in the above conversations. When you've got a few, would you mind having a look at the new changes?

examples/python/create_dashboard_filter.py
fabio-looker: The two loops being consolidated in to one makes this so much clearer, I no longer feel lost :)

On lines 66-68, are you essentially just making a clone of the array? If there is a python method to do that, it might be clearer to use it, otherwise this looks good as is!

fabio-looker (dismissed):

examples/python/create_dashboard_filter.py
dunnkevin: Oh great catch. I've consolidated that down into just the one line in line 66. That loop's definitely unnecessary.

fabio-looker (dismissed):
I think this is now clear enough what's going on, and it will be a valuable example!

PR #819, opened 2021-09-15
fix: intrinsic type params are not Partial<T>

jkaster:
codegen now generates affected `body` parameters as `string` instead of `Partial<string>`.

Typically `body` parameters are structures rather than simple (aka `intrinsic`) types like a string. While TypeScript correctly handles `Partial<string>` this is confusing to the code reader, so this tweaks the generator to remove `Partial<*>` wrapping of the type name for the body parameter on simple types.

This PR uses Looker 21.14 specifications for the updated SDK code that got generated.


bryans99 (approved):
LGTM

PR #820, opened 2021-09-16
chore: update search highlighting style

jkaster:
No longer using the default yellow highlighting for search result matches in markdown text and code results

# Screenshot

image


bryans99 (approved):
LGTM

PR #827, opened 2021-09-21
feat: type tags for APIX SideNav

jkaster:
The Types SideNav now uses the same categories for types as the methods do

## Issues

- clicking a type link doesn't update the sidenav although the content pane is updated
- fix six broken e2e tests because type urls now have tags


jkaster:
This PR is back to draft because it will need #831 when that's merged to main

josephaxisa (approved):
nice job! That had quite the ripple effect

PR #830, opened 2021-09-21
fix: Extension Api Explorer markdown link clicks

bryans99:
Clicking a link in markdown now works for extensions.

Fix is to rely on getAttribute method to get pathname and href values rather
than get pathname and href directly. Also, now tests if url begins with
the spec key as well as the pathname.

Possibly broken by markdown upgrade although why standalone was not
broken is puzzling


jkaster (approved):
Thanks for the fix!

PR #831, opened 2021-09-21
fix: Sidenav selection

bryans99:
Tabs in sidenav were not being synced with history. Now if tab selected,
history is updated. If page is reloaded, tabs are synced to current
history. If link is clicked that updates sidenav panel, panel is updated
accordingly.


jkaster (approved):
LGTM. Nice fixes.

PR #835, opened 2021-09-24
chore: refreshing hackathon app

jkaster:
also

 - updated TypeScript
 - updated Prettier
 - bumped ts-jest
 - updated ts-node


.github/workflows/apix-ci.yml
```diff
@@ -56,7 +56,7 @@ jobs:
           yarn build
 
       - name: Run unit tests
-        run: yarn jest "packages/(code-editor|run-it|api-explorer|extension-api-explorer)/src" --reporters=default --reporters=jest-junit
```
jkaster, line 59: Our "source of truth" is `yarn test:apix` so that's what we're using here

examples/access-token-server/src/routes/status.ts
```diff
@@ -35,15 +35,16 @@ const router = Router()
...
+router.get('/status', async (_req, res) => {
   let statusCode = 200
   let status: any = {
     errors: [],
   }
   try {
     const statusString = readFileSync(`${cwd()}/status.json`, 'utf8')
     status = { ...JSON.parse(statusString), ...status }
-  } catch (err) {
+  } catch (err: any) {
```
jkaster, line 46: TS 4.4 complains about "Unknown" with leaving this `catch (err)`. We have an `IsErrorLike` test in sdk-rtl we can use for skipping annotating with `: any` but I'm not convinced it's worth it. If it is, it's easy enough to catch this pattern in a follow-up PR and modify it then.

package.json
```diff
@@ -36,6 +36,7 @@
     "build:es": "lerna exec --stream 'BABEL_ENV=build babel src --root-mode upward --out-dir lib/esm --source-maps --extensions .ts,.tsx --no-comments'",
     "build:ts": "lerna exec --stream --sort 'tsc -b tsconfig.build.json'",
     "dev:apix": "yarn workspace @looker/api-explorer develop",
+    "dev:hack": "yarn workspace @looker/hackathon develop",
```
jkaster, line 39: added a shortcut for development work on the hack app

examples/access-token-server/src/routes/status.ts
josephaxisa: I remember @AprilArcus  chimed in on this somewhere but I can't find the conversation

josephaxisa (approved):
lgtm. Glad we're on TS 4.4 amongst other things

jkaster:
@AprilArcus is the one who added `isErrorLike` to sdk-rtl

```
/**
 * Type predicate. Asserts that a given object is error-like.
 * @param error a value of unknown type
 * @return boolean true if the error has a `message` key of type string.
 */
export function isErrorLike<T extends unknown>(
  error: T
): error is T & { message: string } {
  if (typeof error !== 'object') return false
  if (!error) return false
  if (!Object.prototype.hasOwnProperty.call(error, 'message')) return false
  if (typeof (error as { message: unknown }).message !== 'string') return false
  return true
}
```

examples/access-token-server/src/routes/status.ts
AprilArcus: It was [here](https://chat.google.com/room/AAAAMqYxkbc/i7Vyiy20YoQ)

packages/run-it/src/components/ConfigForm/ConfigForm.tsx
```diff
@@ -153,7 +153,7 @@ export const ConfigForm: FC<ConfigFormProps> = ({
       [FETCH_INTENT]:
         base_url !== '' && looker_url !== '' ? POSITIVE : CRITICAL,
     })
-  }, [configurator])
+  }, [configurator, getConfig])
```
AprilArcus: good catch

PR #840, opened 2021-09-30
fix: logic for 2D JSON data detection

jkaster:
Only truly 2D JSON data should have a grid. This fixes erroneous DataGrid rendering

### DIFF
--- a/packages/run-it/src/components/ShowResponse/responseUtils.tsx
+++ b/packages/run-it/src/components/ShowResponse/responseUtils.tsx
@@ -73,19 +73,73 @@ export const isColumnar = (data: any[]) => {
   return !complex
 }
 
+type ItemType = 'a' | 'o' | 'v' | 'u'
+
+/**
+ * Is this an array, an object, a value, or undefined
+ * @param value to check
+ */
+const itemType = (value: any): ItemType => {
+  if (!value) return 'u'
+  if (Array.isArray(value)) return 'a'
+  if (value instanceof Object) return 'o'
+  return 'v'
+}
+
+/**
+ * Get the 2D type mapping for the object
+ * @param json to analyze
+ */
+const getTypes = (json: any) => {
+  const types = [new Set<ItemType>(), new Set<ItemType>()]
+  if (!json) {
+    types[0].add('u')
+    return types
+  }
+  for (const key of Object.keys(json)) {
+    const value = json[key]
+    const type = itemType(value)
+    types[0].add(type)
+    switch (type) {
+      case 'a':
+      case 'o':
+        Object.keys(value).forEach((k) => {
+          const v = value[k]
+          types[1].add(itemType(v))
+        })
+        break
+    }
+  }
+  return types
+}
+
+/**
+ * Is this a uniform object that can be converted into a table?
+ * @param json to analyze
+ */
+export const canTabulate = (json: any) => {
+  const types = getTypes(json)
+  return (
+    types[0].size === 1 &&
+    (types[0].has('a') || types[0].has('o')) &&
+    types[1].size <= 1
+  )
+}
+
 /**
  * Show JSON responses
  *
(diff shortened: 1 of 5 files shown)


packages/run-it/src/components/ShowResponse/responseUtils.tsx
```diff
@@ -73,19 +73,73 @@ export const isColumnar = (data: any[]) => {
   return !complex
 }
 
+type ItemType = 'a' | 'o' | 'v' | 'u'
```
bryans99, line 76: Be nice to have a comment describing what these are
a = array
o = object
v = value
u = undefined

Consider p for primitive instead of v for value? All of them (except maybe u) could be considered values. I guess date is not primitive.

Or define an enum

packages/run-it/src/components/ShowResponse/responseUtils.tsx
```diff
@@ -73,19 +73,73 @@ export const isColumnar = (data: any[]) => {
...
+  if (value instanceof Object) return 'o'
+  return 'v'
+}
+
+/**
+ * Get the 2D type mapping for the object
+ * @param json to analyze
+ */
+const getTypes = (json: any) => {
+  const types = [new Set<ItemType>(), new Set<ItemType>()]
```
bryans99, line 94: can json be a string, number, boolean or date?

packages/run-it/src/components/ShowResponse/responseUtils.tsx
```diff
@@ -73,19 +73,73 @@ export const isColumnar = (data: any[]) => {
...
- * If JSON cannot be parsed it will be show as is
+ * If JSON cannot be parsed it will be shown as is
  * @param response
  */
 const ShowJSON = (response: IRawResponse) => {
   const content = response.body.toString()
-  const data = json2Csv(content)
-  const showGrid = isColumnar(data.data)
-  const json = JSON.stringify(JSON.parse(response.body), null, 2)
+  const parsed = JSON.parse(response.body)
```
bryans99, line 139: What if body is not JSON? parse will fail

bryans99 (commented):
Got a bunch of questions to answer before approving

packages/run-it/src/components/ShowResponse/responseUtils.tsx
josephaxisa: the `pickResponseHandler` decides what renderer to use. If we're in this function it has to be a json:

https://github.com/looker-open-source/sdk-codegen/blob/main/packages/run-it/src/components/ShowResponse/responseUtils.tsx#L249-L259

josephaxisa: the pickResponseHandler decides what renderer to use. If we're in this function it has to be a json:

https://github.com/looker-open-source/sdk-codegen/blob/main/packages/run-it/src/components/ShowResponse/responseUtils.tsx#L249-L259

packages/run-it/src/components/ShowResponse/responseUtils.tsx
jkaster: Ok. It was internal only but I suppose it's better for someone else to understand it. I'll look at enum

jkaster: Rather than having someone obsess about "primitive" vs. "scalar" vs. "value" etc I think I'll call it a "simple" type vs. a complex/compound type (object or array).

packages/run-it/src/components/ShowResponse/responseUtils.tsx
jkaster: the value passed in that we _expect_ to be JSON could be any type

josephaxisa (approved):
lgtm, 🚢  it!

PR #841, opened 2021-10-01
fix: remove body from REST call if no body elements are defined

jkaster:
Fixes the sporadic scenario where a `body` value of `{}` is sent to an endpoint. This update removes any "empty" body from the RunIt request submission


josephaxisa (approved):
lgtm

PR #843, opened 2021-10-05
feat: @looker/redux package

josephaxisa:
This is a copy of the `@looker/redux` package used internally


package.json
```diff
@@ -189,6 +189,14 @@
           "testing-library/render-result-naming-convention": "off"
         }
       },
+      {
+        "files":  ["packages/redux/**/*.ts?(x)", "packages/redux/**/*.spec.ts?(x)"],
```
jkaster, line 193: can you also add a `yarn test:redux` entry?

jkaster (changes requested):
Quick change request

package.json
josephaxisa: Good idea. Unfortunately `packages/redux/src/createSliceHooks/index-integration.spec.tsx` seems to be failing right now

jkaster: Bummer. Puppeteer config issue?

jkaster (approved):
LGTM for the intro

package.json
josephaxisa: it's not a e2e test. The name is misleading but it's just an RTL test. I cannot rename otherwise updating the package will become an issue over time.

PR #845, opened 2021-10-05
fix: stack overflow for type tagging during codegen

jkaster:
Change was made for APIX type tagging, and a scenario was missed that caused infinite recursion


josephaxisa (approved):
:shipit:

PR #846, opened 2021-10-05
fix: validateBody for API Explorer

jkaster:
Body values that were already parsed were failing to parse (obviously) so the validation was failing.

I completely missed that scenario

This needs to be merged before  API Explorer is deployed because RunIt will not work for any endpoint with a body

Definitely need to add e2e tests for RunIt!


josephaxisa (approved):
lgtm

PR #855, opened 2021-10-13
chore: make environment adaptor globally available

josephaxisa:
This eliminates the need for the envAdaptor context and paves the way for envAdaptor usage in sagas


jkaster:
Is Bryn's feedback going to be address in a different PR, or will this PR be updated for it?

josephaxisa:
I addressed and resolved all comments already except for 1 which is part of a follow up PR.

PR #860, opened 2021-10-15
feat: headless nav toggle

josephaxisa:
Added a toggle for when APIX is running in headless mode. Screenshots don't really help, I suggest trying this locally.

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -3,7 +3,7 @@
   "bump-minor-pre-major": true,
   "bump-patch-for-minor-pre-major": true,
   "packages": {
-    ".": { "release-as": "" },
+    ".": { "release-as": "21.16.0" },
     "packages/api-explorer": { "release-as": "" },
     "packages/code-editor": { "release-as": "" },
     "packages/extension-api-explorer": { },


josephaxisa:
I just tested it locally and can confirm the same result but this PR does not touch that logic. I am seeing the same issue on `main`. Hardcoding `headless={true}` results in everything working as expected. 

https://github.com/looker-open-source/sdk-codegen/blob/31fe97c53c0d7c564e9a7b630b61725730a3533c/packages/api-explorer/src/StandaloneApiExplorer.tsx#L105-L110

Looking into this briefly I can see that on initial load we are reading versions from the statically served `versions.json` that's bundled with API Explorer. After a successful OAuth flow, the `/versions` endpoint on the API server is called and it is expected to contain the `headless` property which doesn't sound right, please correct me if I'm wrong. 

Perhaps the indicator of whether an instance of API Explorer is headless or not should be determined at the usage site by passing in the prop `headless={true}`. I don't think there's a need to persist it in local storage either. Right now the prop value is getting overridden because of the absence of the `headless` prop from the /versions payload.

PR #862, opened 2021-10-18
chore: update release-please config for 21.16

drstrangelooker:


jkaster:
It would be nice to have semantic checks for the PR, but first the plugin we're currently using has to show it's reliable again. Then we can work on adding the check for PRs too. Or I guess we write our own, but I'd rather the existing one become reliable again.

PR #881, opened 2021-10-29
fix: optimize scripts for parallel startup of Looker docker image

drstrangelooker:
I move the setup for getting and starting the Docker image to the start of the process. Then I let the other setup proceed.
Finally, just before the actual tests, I run the wait_for_looker script to check that Looker has started successfully.


.github/workflows/codegen-ci.yml
```diff
@@ -32,13 +32,32 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Repo Checkout
+        uses: actions/checkout@v2
+
```
jkaster, line 37: does moving this up affect the responsiveness of canceling previous runs? Just wondering why it was moved

jkaster (approved):
LGTM. One question about moving one of the steps

.github/workflows/codegen-ci.yml
drstrangelooker: The repo checkout can overwrite some of the previous work that was done. I ran into a weird situation with the python-ci.yml where not having the Repo Checkout first caused the gcloud auth step to fail. So I decided to do it first for all of them just to make sure there wasn't a race condition.

jkaster: Thanks. Sounds good

PR #883, opened 2021-10-29
feat: enforce required body parameters for RunIt

jkaster:
Body parameters that are required must be provided before RunIt will Run

## Screenshot

image


josephaxisa (dismissed):
cool addition, lgtm

josephaxisa (approved):
still lgtm

PR #884, opened 2021-10-29
fix: Hackathon styling and resource update

jeremytchang:
- Updated Loading component to default to always display if not
passed in loading prop. Making it as dumb as possible
- Updated Loading rendering logic and styling in scene components
so layouts does not shift/change after loading is finished.
- Updated side nav to be thinner to accomodate smaller viewports
- Updated resources with new component and dataset tag
- Added header to each page for easier navigability
- Added hacker faq link to home/agenda page
- Added additional text and field styling to resource page for
better UX

### DIFF
--- a/packages/hackathon/src/scenes/HomeScene/HomeScene.tsx
+++ b/packages/hackathon/src/scenes/HomeScene/HomeScene.tsx
@@ -27,8 +27,9 @@
 import type { FC } from 'react'
 import React from 'react'
 
-import { Heading, SpaceVertical } from '@looker/components'
+import { Heading, SpaceVertical, Paragraph, Span } from '@looker/components'
 import type { IHackerProps } from '../../models'
+import { ExtMarkdown } from '../../components'
 import { Agenda } from './components'
 import { localAgenda } from './agenda'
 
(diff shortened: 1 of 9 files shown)


packages/hackathon/src/scenes/HomeScene/HomeScene.tsx
```diff
@@ -42,9 +43,18 @@ export const HomeScene: FC<HomeSceneProps> = ({ hacker }) => {
   return (
     <>
       <SpaceVertical gap="u5">
-        <Heading as="h2" fontSize="xxxlarge" fontWeight="medium">
-          Agenda
-        </Heading>
+        <Span>
```
jeremytchang, line 46: Nested in span so the paragraph is not too far from Heading in SpaceVertical.

jkaster: For v.next with artifact storage we'll schedule some design review too

jkaster (approved):
Lots of nice little improvements. Good sweep.

packages/hackathon/src/scenes/HomeScene/HomeScene.tsx
jeremytchang: Totally. I'm pretty sure I'm not following design best practices here, only aiming for what looks good. Will appreciate their input.

PR #885, opened 2021-11-01
fix: api explorer initialization 

josephaxisa:
The call of a saga from within another saga was resulting in issues for the extension version. This PR changes the way envAdapter methods are called from within sagas.


jkaster (approved):
LGTM, fix confirmed

PR #891, opened 2021-11-03
feat: judging view for Hackathon app

jkaster:
# Added

- Judging view dialog
- Project view dialog to project list as "View Project"

# Removed

- Project View scene 

# Screenshots

## Judging View

image

image

## Project View

image

image


packages/hackathon/src/scenes/JudgingEditorScene/components/JudgingViewDialog.tsx
```diff
@@ -23,4 +23,31 @@
...
+import type { IJudgingProps } from '../../../models'
+import { JudgingView } from './JudgingView'
+
+interface JudgingViewDialogProps {
+  judging?: IJudgingProps
+  closer: () => void
+}
+
+export const JudgingViewDialog: FC<JudgingViewDialogProps> = ({
+  judging,
```
jeremytchang, line 38: `JudgingViewDialog` props aren't clear about their behavior. The existence of judging determines if the dialog is open or closed. Is the props to follow the old props of `ProjectViewDialog`? 

Minor, but i'd recommend naming `closer` to `onClose` since it's an event listener and passing it through.

Both are minor comments, not blocking.

packages/hackathon/src/scenes/ProjectsScene/components/utils.ts
```diff
@@ -24,4 +24,28 @@
 
  */
 
-export * from './ProjectView'
+import type { ITechnologyProps } from '../../../models'
```
jeremytchang, line 27: This file should be moved up a folder or two since its shared between projects and judging scene.

jeremytchang (approved):
LGTM. Thanks! Looks great! :D

jeremytchang (dismissed):
LGTM.
Typically you dont need to comment the `onClose` handler. Prefixing `on`, pretty much means its an event handler. But thank you for heeding the feedback!

jeremytchang (approved):

PR #893, opened 2021-11-05
fix: Updated hackathon readme landing page path

jeremytchang:
Updated to match what is running in production

### DIFF
--- a/packages/hackathon/README.md
+++ b/packages/hackathon/README.md
@@ -82,7 +82,7 @@ Remember to add a model to the project that has a valid connection.
 
 ## Specific steps to `yarn`
 1. run `yarn install` in sdk-codegen
-2. run `yarn build` in sdk-codgen
+2. run `yarn build` in sdk-codegen
 3. run `yarn start` in examples/access_token_server
 4. run `yarn dev:hack` in sdk-codegen to start the development server and connect to the extension
 
--- a/packages/hackathon/README.md
+++ b/packages/hackathon/README.md
@@ -21,8 +21,8 @@ To configure your access token server, see the instructions in the [access token
 Until the Hackathon extension is available via the Looker Marketplace, a Hackathon manifest needs to be used. This manifest should have:
 
 ```lookml
-project_name: "hackathon_app"
-application: hackathon_app {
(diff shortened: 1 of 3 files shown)


packages/hackathon/README.md
```diff
@@ -60,7 +81,7 @@ application: hackathon_app {
 Remember to add a model to the project that has a valid connection.
 
 ## Specific steps to `yarn`
-1. run `yarn install` in sdk-codegen 
+1. run `yarn install` in sdk-codegen
 2. run `yarn build` in sdk-codgen
```
jkaster, line 85:
```suggestion
2. run `yarn build` in sdk-codegen
```

packages/hackathon/src/data/add_user/sagas.ts
```diff
@@ -50,7 +50,7 @@ const HACKATHON_USER_ATTR_NAME = 'hackathon' // case sensitive
 const HACKATHON_USER_ATTR_LABEL = 'Hackathon'
 const HACKATHON_GROUP_PREFIX = 'Looker_Hack: '
 const LANDING_PAGE_ATTR_NAME = 'landing_page'
-const LANDING_PAGE_PATH = '/extensions/hackathon::hackathon_app'
+const LANDING_PAGE_PATH = '/extensions/hackathon::hackathon'
```
jkaster, line 53: Is the user registration script setting this path for the default landing page? Are all existing registrants going to have their homepage updated to this new URL?

jkaster (commented):
A suggestion and a  question

packages/hackathon/src/data/add_user/sagas.ts
jeremytchang: The user registration group creates a group for hackathon 2021. The script assigns a role  to the group that defines the landing page. When users are added, they are added to the group and thus get the landing page attribute overwritten by the group role. 
So technically, this change shouldn't be needed. And i can jsut change our manifest.kml to rename to `hackathon_app`. But i want to be safe.

jkaster (dismissed):
LGTM

jkaster (approved):
LGTM

PR #894, opened 2021-11-08
feat: disable "Add Project" when judging starts

jkaster:
- Judging time for the hackathon is always shown at the bottom of the project list now
- If judging has started:
  - the judging start time is displayed to the left of the `Add Project` button
  - if the user has the ability to lock projects, `Add Project` remains enabled
  - if the user is a regular hackathon attendee, `Add Project` is disabled

- Made intro agenda on the home page display English or Japanese depending on the user's locale

# Screenshots

## Regular hackathon attendee

image

## User who can lock projects

image

### DIFF
--- a/packages/hackathon/src/scenes/HomeScene/HomeScene.tsx
+++ b/packages/hackathon/src/scenes/HomeScene/HomeScene.tsx
@@ -53,7 +53,7 @@ export const HomeScene: FC<HomeSceneProps> = ({ hacker }) => {
             <ExtMarkdown
               source={`### Our [Hackathon FAQ](https://community.looker.com/hackathome-2021-1026/hackathome-2021-attendee-faq-28429) contains all event details!
 *Change your [account](https://hack.looker.com/account) timezone to display times in your timezone.*${MARKDOWN_LINEBREAK}
-*Change your [account](https://hack.looker.com/account) locale to ${'`ja_JP`'} to display agenda in japanese.*`}
+*Change your [account](https://hack.looker.com/account) locale to ${'`ja_JP`'} to display the agenda in Japanese.*`}
             />
           </Paragraph>
         </Span>
--- a/packages/hackathon/src/scenes/HomeScene/HomeScene.tsx
+++ b/packages/hackathon/src/scenes/HomeScene/HomeScene.tsx
@@ -41,6 +41,16 @@ const MARKDOWN_LINEBREAK = '  '
 
 export const HomeScene: FC<HomeSceneProps> = ({ hacker }) => {
   const schedule = localAgenda(hacker.locale)
+  const host = 'https://hack.looker.com'
+  const intro =
+    hacker.locale === 'ja_JP'
+      ? `### ハッカソン詳細については、[よくある質問記事](https://community.looker.com/hackathome-2021-1026/hackathome-2021-%E3%82%88%E3%81%8F%E3%81%82%E3%82%8B%E8%B3%AA%E5%95%8F-28518)をご確認いただけます。
+現地時間が表示されるため、[アカウント](${host}/account)の「タイムゾーン」を設定できます。${MARKDOWN_LINEBREAK}
+アジェンダが日本語で表示されるため、[アカウント](${host}/account)の「Locale」は「ja_JP」に指定できます。
+`
+      : `### Our [Hackathon FAQ](https://community.looker.com/hackathome-2021-1026/hackathome-2021-attendee-faq-28429) contains all event details!
+*Change your [account](${host}/account) timezone to display times in your timezone*${MARKDOWN_LINEBREAK}
+*Change your [account](${host}/account) locale to \`ja_JP\` to display the agenda in Japanese.*`
 
   return (
     <>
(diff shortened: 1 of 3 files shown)


jeremytchang (approved):
LGTM. We still have to lock projects when judging time comes. Ideally this should all be consolidated into one automatic flow, where when judging time starts, projects are auto locked and projects additions is stopped for hackers. These flows would not apply for staff and admin.

packages/hackathon/src/scenes/ProjectsScene/ProjectsScene.tsx
```diff
@@ -63,6 +67,23 @@ export const ProjectsScene: FC<ProjectSceneProps> = () => {
     if (hackathon) dispatch(lockProjects(false, hackathon._id))
   }
 
+  useEffect(() => {
```
jeremytchang, line 70: Not sure if useEffect is needed here. Unless we are triggering project scene re-render periodically, this is not needed.

packages/hackathon/src/scenes/ProjectsScene/ProjectsScene.tsx
```diff
@@ -63,6 +67,23 @@ export const ProjectsScene: FC<ProjectSceneProps> = () => {
...
+    if (hackathon && hacker) {
+      const started = hackathon.judging_starts?.getTime() < new Date().getTime()
+      setJudgingStarts(
+        `Judging start${started ? 'ed' : 's'} ${zonedLocaleDate(
```
jeremytchang, line 74: `started ? 'ed' : 's'` is dead code given conditional rendering at line 97. 
With that said, I say we remove the conditional rendering at line 97 and always show when project judging will begin.

Also need to add `at` so it reads `Judging start(s)(ed) at date).

jeremytchang (changes requested):
LGTM with changes.

packages/hackathon/src/scenes/ProjectsScene/ProjectsScene.tsx
jkaster: I am undecided where judging start time should show in the final implementation, so I'm only showing it for now when it has an impact on the project board. If I added "at" to be grammatically correct, I'd also need to add "on".

So I'm going with a colon instead.

packages/hackathon/src/scenes/ProjectsScene/ProjectsScene.tsx
jkaster: I may not understand `useSelector()` well enough yet, but it seems `hackathon` and `hacker` may not be established when this page is first opened. Is that correct?

jeremytchang: Redux state change will automatically kick off a render. 
`When an action is dispatched, useSelector() will do a reference comparison of the previous selector result value and the current result value. If they are different, the component will be forced to re-render. If they are the same, the component will not re-render.`
https://react-redux.js.org/api/hooks
There's more nuances when dealing with state comparision, but that shouldn't matter here, you can test it to double check.

jeremytchang: Also afaict, project scene is not rendered or can be navigated to until hacker/hackathon is all loaded in the redux state in the first place. Hence the long lag between app opening and project tab appearing.
(Something to fix later)

packages/hackathon/src/scenes/ProjectsScene/ProjectsScene.tsx
```diff
@@ -73,7 +94,16 @@ export const ProjectsScene: FC<ProjectSceneProps> = () => {
...
+          </Span>
+        )}
+        <Button
+          iconBefore={<Add />}
+          onClick={handleAdd}
+          disabled={(isLoading || judgingStarted) && !canLockProject(hacker)}
+        >
           Add Project
         </Button>
         <>
```
jeremytchang, line 109: The line below this one already demonstrates dependence on hackathon and hacker props to render. Also do not be confused by useSelector. It is almost the same as mapState, which maps state to props on the component. If a prop changes, the component re-renders.

jeremytchang (changes requested):
I wrote up an example of what the code could be. 

Also moved the text to after the button. From previous frontend work, typically statuses appear after the button, so the button is not shifted when status changes (though that doesn't apply in this case, since the button will never shift). It also reads better, from left to right. You see the Add button disabled, then you read the status or reason for it.

https://github.com/looker-open-source/sdk-codegen/blob/jc/add_project_example/packages/hackathon/src/scenes/ProjectsScene/ProjectsScene.tsx

packages/hackathon/src/scenes/ProjectsScene/ProjectsScene.tsx
jkaster: Thanks for the explanation about `useSelector()`. I think there may be other places with `useEffect()` that don't need it, but we can worry about that when we switch to the new persistence layer. It is possible someone could be on the project page when they reload, and not move from it until after it successfully loads, but we have a "not found" scenario that already handles that. Less code is more.

jeremytchang (approved):
LGTM

packages/hackathon/src/scenes/HomeScene/HomeScene.tsx
```diff
@@ -41,6 +42,16 @@ const MARKDOWN_LINEBREAK = '  '
 
 export const HomeScene: FC<HomeSceneProps> = ({ hacker }) => {
   const schedule = localAgenda(hacker.locale)
+  const host = getExtensionSDK().lookerHostData?.hostUrl
+  const intro =
+    hacker.locale === 'ja_JP'
+      ? `### ハッカソン詳細については、[よくある質問記事](https://community.looker.com/hackathome-2021-1026/hackathome-2021-%E3%82%88%E3%81%8F%E3%81%82%E3%82%8B%E8%B3%AA%E5%95%8F-28518)をご確認いただけます。
```
jeremytchang, line 48: Where did these translation come from?

jeremytchang (approved):
LGTM

packages/hackathon/src/scenes/HomeScene/HomeScene.tsx
jkaster: From a PR @lanlooker did last night. 🪄

packages/hackathon/src/scenes/ProjectsScene/ProjectsScene.tsx
jkaster: Good stuff. Thanks for the cleaner version. I put it into this PR.

PR #899, opened 2021-11-09
feat: Looker 21.20 bindings

drstrangelooker:


packages/sdk-codegen/src/python.gen.ts
```diff
@@ -86,6 +86,7 @@ export class PythonGen extends CodeGen {
     'in',
     'is',
     'lambda',
+    'models',
```
joeldodge79, line 89: Honestly I think we might be better off renaming the import. The plus side is we stay truer to the API spec. The downside is we break anyone who is doing something like `from looker_sdk.sdk.api40.methods import models` but they should be doing `from looker_sdk.sdk.api40 import models`

joeldodge79 (changes requested):
what do you think about just changing the import inside methods.py to something like

`from . import models as mdls`

joeldodge79 (dismissed):
not pretty but I think that's the best we'll do for python. I think you *might* need to update some tests but we'll find out

joeldodge79:
@drstrangelooker looks like the look-id int -> string change needs to be updated in the python tests

joeldodge79:
@drstrangelooker I feel like we have a gap in CI. I would have expected [this test](https://github.com/looker-open-source/sdk-codegen/blob/d3386ae034dd5bacd067de791fe5ba5f3ed06ac4/packages/sdk-codegen/src/python.gen.spec.ts#L193) to fail. Maybe the [Codegen CI workflow](https://github.com/looker-open-source/sdk-codegen/blob/d3386ae034dd5bacd067de791fe5ba5f3ed06ac4/.github/workflows/codegen-ci.yml#L87) is too narrow?

packages/sdk-codegen/src/python.gen.ts
drstrangelooker: I went with this, BTW.

joeldodge79 (changes requested):
I think we should fix the workflows so that the `.gen.spec.ts` files get run, especially because I *think* at least python.gen.spec.ts is broken now? https://github.com/looker-open-source/sdk-codegen/pull/899#issuecomment-964602558

drstrangelooker:
@joeldodge79 can you give some more detail about what you mean by running the `.gen.spec.ts`? Don't they get run during the `yarn gen` process?

jkaster:
Joel correctly called out that python.gen.spec.ts (the *test* for python.gen.ts) needs to be updated and should also be part of the CI call. Interesting that `yarn test:sdk` isn't being invoked (evidently, because it should fail).

joeldodge79:
sorry, yeah what I mean is the specific language generator tests under `packages/sdk-codegen/src/` seem to not be getting run as part of the build

drstrangelooker:
I've been using `bin/smoke` and `bin/unit_test`. I'm cleaning up the results of `yarn test` right now.

packages/sdk-codegen/src/sdkModels.spec.ts
```diff
@@ -73,7 +73,7 @@ describe('sdkModels', () => {
     const api4 = readFileSync(rootFile('spec/Looker.4.0.oas.json'), 'utf-8')
     const api = ApiModel.fromString(api4)
     const actual = Object.keys(api.typeTags)
-    expect(actual).toHaveLength(29)
+    expect(actual).toHaveLength(30)
```
jkaster, line 76: I keep postponing the `TODO` of making this use a stable spec snapshot

packages/sdk-codegen/src/specDiff.spec.ts
```diff
@@ -236,7 +236,7 @@ describe('spec differ', () => {
       expect(actual).toHaveLength(1)
       expect(actual[0].diffCount).toEqual({
         added: 0,
-        changed: 2,
+        changed: 3,
```
jkaster, line 239: yeah, "stable spec" `TODO` again

jkaster (dismissed):
LGTM. I think this should pass CI.

.github/workflows/python-ci.yml
```diff
@@ -98,7 +98,6 @@ jobs:
         os:
           - ubuntu
         looker:
-          - '21_12'
```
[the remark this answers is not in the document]
drstrangelooker: reverted this already

.github/workflows/tssdk-ci.yml
```diff
@@ -101,7 +101,6 @@ jobs:
         os:
           - ubuntu
         looker:
-          - '21_12'
```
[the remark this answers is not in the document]
drstrangelooker: reverted this already

drstrangelooker:
Found our issue. The tests are run by
```
yarn jest "packages/sdk-codegen(-utils|-scripts)/src" --reporters=default --reporters=jest-junit
```
Any tests in `packages/sdk-codegen/src` are skipped. Instead we need this expression:
```
yarn jest "packages/sdk-codegen(|-utils|-scripts)/src" --reporters=default --reporters=jest-junit
```

drstrangelooker:
> I think we should fix the workflows so that the `.gen.spec.ts` files get run, especially because I _think_ at least python.gen.spec.ts is broken now? [#899 (comment)](https://github.com/looker-open-source/sdk-codegen/pull/899#issuecomment-964602558)

Should be resolved now.

.github/workflows/codegen-ci.yml
```diff
@@ -84,7 +84,7 @@ jobs:
           ${{ github.workspace }}/.github/scripts/wait_for_looker.sh
 
       - name: Run unit tests
-        run: yarn jest "packages/sdk-codegen(-utils|-scripts)/src" --reporters=default --reporters=jest-junit
+        run: yarn jest "packages/sdk-codegen(|-utils|-scripts)/src" --reporters=default --reporters=jest-junit
```
jkaster, line 87: Great find. That's been there a long time. We've been lucky!

jkaster (approved):
LGTM

.github/workflows/codegen-ci.yml
drstrangelooker: Thanks!

PR #905, opened 2021-11-10
fix: Remove hackathon invite only option

jeremytchang:
Turns out invite only has not been implemented yet.
Removing UI option till it is implemented.
TODO work tracked in ticket.


jkaster (approved):
Looks better to me

PR #906, opened 2021-11-11
feat: show full request url in RunIt response

jkaster:
The Response tab now shows the full request URL, not including the request body (which can't be included in this scenario)

# Screenshots

image

image

image

### DIFF
--- a/packages/run-it/src/RunIt.tsx
+++ b/packages/run-it/src/RunIt.tsx
@@ -35,6 +35,7 @@ import {
   useTabs,
 } from '@looker/components'
 import type { ApiModel, IMethod } from '@looker/sdk-codegen'
+import type { BaseTransport } from '@looker/sdk-rtl'
 import type { ResponseContent } from './components'
 import {
   RequestForm,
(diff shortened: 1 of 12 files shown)


packages/run-it/src/RunIt.tsx
```diff
@@ -126,6 +127,7 @@ export const RunIt: FC<RunItProps> = ({
     initRequestContent(configurator, inputs)
   )
   const [activePathParams, setActivePathParams] = useState({})
+  const [activeQueryParams, setActiveQueryParams] = useState({})
```
josephaxisa, line 130: suggestion: given how tightly coupled path and query params are and that they are always updated together, consider merging these into one state object, `activeRequestParams`. Resultant state updates can then happen in one call.

josephaxisa (commented):
LGTM but I think it's worth adding a few test cases for `ResponseExplorer`

jkaster:
This just got a bit ugly when I was trying to achieve full URL derivation when used as an extension, so this is going to need a little bit of redesign, possibly as part of a RunIt adaptor interface.

PR #908, opened 2021-11-16
feat: Python SDK example: write query result to BigQuery using Cloud Functions 

lanlooker:
This script provides an API/SDK workaround for the most requested feature suggestion for the System Activity team:
"Edit LookML in the System Activity model" (https://feedback.us.pendo.io/app/#/case/17625)
"Export system activity explores directly to BQ" (https://portal.feedback.us.pendo.io/app/#/case/61595)

For context: b/204835776

Log: 
image

Screencast: 

<p align="center">
  Setting environmental variables in Cloud Function UI
</p>


examples/python/cloud-function-write-to-bigquery/main.py
```diff
@@ -0,0 +1,55 @@
...
+  ### Transform headers and data ("User Name" to become "User_Name")
+  cnt = 0
+  for i in data: 
+    if i == "\n":
+        break
+    else: 
+        cnt += 1
+  header = data[:cnt]
+  header_to_write = header.replace(" ", "_")
+  data_to_write = data[cnt:]
```
joeldodge79, line 32: is this because bq column names can't have spaces? If so, might mention that in the comment.

examples/python/cloud-function-write-to-bigquery/main.py
```diff
@@ -0,0 +1,55 @@
...
+  print("Successfully wrote data to a CSV file stored in temporary disk")
+
+def load_to_bq():
+    # Prepare the table inside BQ in advance. Additional logic can be written to create a table 
+    # Example: https://github.com/googleapis/python-bigquery/blob/35627d145a41d57768f19d4392ef235928e00f72/samples/create_table.py
+    table_id = "myproject.myschema.mytable" 
+    job_config = bigquery.LoadJobConfig(
+        source_format=bigquery.SourceFormat.CSV, skip_leading_rows=1, autodetect=True,
+    )
+    with open("/tmp/table1.csv", "rb") as source_file:
```
joeldodge79, line 46: what is `'table1.csv'` and how does that relate to `'table.csv'` on L34?

joeldodge79 (changes requested):
couple questions

examples/python/cloud-function-write-to-bigquery/main.py
lanlooker: Thank you! I added a comment inside the function and at the top of the script

examples/python/cloud-function-write-to-bigquery/main.py
lanlooker: This is typo from my end. They should be the same (write to table.csv and then read from table.csv) - I made change to the code.

lanlooker:
Hi @joeldodge79, I made changes to the script per your suggestion. Please let me know if you have additional feedback! Thank you! 🙏🏻

PR #910, opened 2021-11-18
feat: OAuthConfigProvider

josephaxisa:
- Replaced `RunItSettings` and `defaultConfigurator`
- Registered `envAdaptor` in the `RunIt` component (for when it's run outside of the API Explorer context)
- Eliminated all `sdk` and `configurator` props. The former is now obtained where needed using the envAdaptor
- Updated various tests that needed it

Note that I introduced a regression here by purposely removing functionality responsible for restoring request form state after OAuth handshake. Local storage interaction is now handled via the envAdaptor making it async and as a result, various tests break. Instead of fixing this temporarily I am going to address it with sagas once the required state is in redux.


packages/run-it/src/components/ConfigForm/ConfigForm.spec.tsx
```diff
@@ -28,27 +28,24 @@ import React from 'react'
...
-  const sdk = initRunItSdk()
+  const adaptor = new BrowserAdaptor(initRunItSdk())
+  registerTestEnvAdaptor(adaptor)
+
   const apiLabel = /API server URL/i
   const authLabel = /OAuth server URL/i
-
   beforeEach(() => {
-    defaultConfigurator.removeStorage(RunItConfigKey)
+    localStorage.removeItem(RunItConfigKey)
```
jkaster, line 43: Should this be using an adaptor storage call or do we presume the unit tests will always be browser based?

jkaster (approved):
LGTM. Just a minor question about testing policy that doesn't need to be address for this PR to be approved.

packages/run-it/src/components/ConfigForm/ConfigForm.spec.tsx
josephaxisa: This PR modifies `ConfigForm` to use `localStorage` directly, not via the envAdaptor. The reason being that localstorage interactions would otherwise be async, requiring a significant amount of changes. Once we have redux in place this will be delegated to sagas, making it much easier. Also `ConfigForm` is entirely for browser apps right now as extensions do not need it.

jkaster (approved):
LGTM

PR #916, opened 2021-12-03
chore: Explicitly call tsc with yarn lint:ts in apix CI

jeremytchang:
Replaced apix ci call to `yarn build-extensions` with call to
`yarn lint:ts` to trigger TSC to catch typescript compilation errors.

Previously we called to `yarn build-extensions` to trigger tsc.


bryans99 (approved):
LGTM

PR #917, opened 2021-12-03
chore: hackathon ci

bryans99:


jkaster (approved):
Good first cut for the hack app. Thanks!

PR #918, opened 2021-12-05
fix: enum naming and registration

jkaster:
Enum names were poorly determined, and in some cases multiple duplicates of the same enum name could result in a type property not getting correctly updated to the correct enum representing its values.

This cleans up the enum registration in the `findName()` function


packages/sdk-codegen/src/sdkModels.ts
```diff
@@ -1836,24 +1841,40 @@ export class EnumType extends Type implements IEnumType {
       )
     }
     if (methodName) {
-      this.description = `Type defined in ${methodName}\n${this.description}`
+      this.description = `${this.description}${
+        this.description ? ' ' : ''
+      }(Enum defined in ${methodName})`
```
joeldodge79, line 1846: I'm curious why the python generated code doesn't get this description addition?

packages/sdk-codegen/src/sdkModels.spec.ts
```diff
@@ -59,1212 +59,1214 @@ import {
...
-      ['foo_bar', 'foo_bar'],
-      ['foo   bar', 'foo_bar'],
-      ['foo -  bar  - - baz', 'foo_bar_baz'],
-      ['Foo  -  Bar', 'Foo_Bar'],
-      ['FOOBAR', 'FOOBAR'],
-    ])('"%s" is "%s"', (actual, expected) => {
-      expect(safeName(actual)).toEqual(expected)
-    })
-  })
+      // Creates unique names
```
joeldodge79, line 122: Without spelunking into the code I would have thought that `name` influenced the generated enum name in models files but it doesn't appear to. How is `name` used?

on second thought, I suppose we just don't have any actual cases in publicly generated code for this? If that's the case, it might be worthwhile adding to the fake section of the test spec and verifying the actual generated output uses a "renamed" name (like `ResultFormat2`)?

joeldodge79 (changes requested):
a question and a suggestion for further testing

packages/sdk-codegen/src/sdkModels.ts
jkaster: it does. See the updated python.gen.spec.ts test

packages/sdk-codegen/src/sdkModels.ts
```diff
@@ -2307,23 +2328,29 @@ export class ApiModel implements ISymbolTable, IApiModel {
...
 
-    if (hash in this.enumTypes) {
-      /**
-       * whatever type this was, it's now the same as the existing enum type
-       */
-      return this.enumTypes[hash]
+    if (type.name in this.types) {
+      const hash = md5(type.asHashString())
+      const matched = this.enumTypes[hash]
+      if (matched && matched.name === type.name) {
```
bryans99, line 2335: matched?.name === type.name ?

packages/sdk-codegen/src/sdkModels.ts
```diff
@@ -2307,23 +2328,29 @@ export class ApiModel implements ISymbolTable, IApiModel {
...
       const method = this.methods[methodName]
       if (method) {
+        // add a type reference for the method
         method.types.add(type.name)
         method.customTypes.add(type.name)
       }
     }
-    this.enumTypes[hash] = type
+    // tail end of dedupe logic
+    // this.enumTypes[hash] = type
```
bryans99, line 2353: Remove commented out code?

packages/sdk-codegen/src/sdkModels.ts
```diff
@@ -2364,6 +2391,13 @@ export class ApiModel implements ISymbolTable, IApiModel {
     return result
   }
 
+  /**
+   * Read-only accessor for private enum collection used within this source file as a "friend"
```
bryans99, line 2395: This is actually protected but maybe I'm misreading the comment. From a purest standpoint as this is an object this is technically updatable by the caller. 
`return {...this,enumTypes}`

bryans99 (commented):
Some minor things. Can approve if you dont feel like changing

packages/sdk-codegen/src/sdkModels.spec.ts
jkaster: test spec updated, test added to typescript.gen.spec.ts

packages/sdk-codegen/src/sdkModels.ts
jkaster: it needs to be updateable

packages/sdk-codegen/src/sdkModels.ts
joeldodge79: guess I missed that because I also didn't see any changes in python generated models (where-as I do for TS generated models) - I feel like when we make a generator change like this we should also updated sdks so that the next person to do it isn't surprised

PR #924, opened 2021-12-16
chore: Added bundle step and cleaned up hackathon CI

jeremytchang:
- Added `yarn bundle` step to make sure hackathon can always be bundled
- Renamed and split up steps to be more descriptive

### DIFF
--- a/.github/workflows/hackathon-ci.yml
+++ b/.github/workflows/hackathon-ci.yml
@@ -48,7 +48,18 @@ jobs:
       - name: Install dependencies
         run: |
           yarn
+
+      - name: Lint
+        run: |
           yarn lint:es --quiet
           yarn lint:ts
+
+      - name: Build all
+        run: |
           yarn build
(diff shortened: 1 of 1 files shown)


jkaster (changes requested):
I think this  may run on every PR as written. See the other workflow files for how they limit the scope of CI to the affected paths.

jeremytchang:
@jkaster This workflow should already only run on wholly-sheet and hackathon path only. Unless i'm misunderstanding this `on` [section](https://github.com/looker-open-source/sdk-codegen/blob/main/.github/workflows/hackathon-ci.yml#L2) this existing workflow should be limited in scope already.

Also if that was the case, the checks for this PR would have already run the Hackathon CI too, but it has not.

jeremytchang:
Also this is only adding steps to the original hackathon CI workflow Bryan and you approved/merged before.

jkaster (approved):
LGTM

PR #925, opened 2021-12-16
feat: add http method to IRawResponse

jkaster:
- added `| null` to the declaration of TypeScript SDK interface properties that are nullable according to the API specification

- restored EOF newline for TS prettier reformatting

- also updated caniuse-lite


bryans99 (approved):
LGTM

PR #932, opened 2021-12-30
chore: Added resources index CI

jeremytchang:
Added CI that triggers on resources index code changes and resource json
changes. It will test/analyze/build the resources index and also check
if resources.json or resource-lock.json are missing changes.

I've tested this locally with act. https://github.com/nektos/act


fabio-looker:
Thanks for running me through the logic to confirm. Looks good!

fabio-looker (approved):

PR #933, opened 2022-01-08
feat: Looker SDK generator for the Dart language

bryans99:
# Description:

A generator for the Dart language. Note, only supports Looker SDK 4.0.

# Changes:

1. Dart generator.
2. Tests for dart generator.
3. Generated dart methods and models.
4. Runtime library for dart.
5. Example dart implementation.
6. Dart unit tests,
7. Dart end to end tests.


bryans99:
Converted to draft. google g3 reports a bunch of lint errors. Turned out I had missed turning on full lint. Have figured out how to do that but it will take a while to fix.

dart/looker_sdk/example/main.dart
```diff
@@ -0,0 +1,108 @@
...
+import 'package:looker_sdk/index.dart';
+import 'package:dotenv/dotenv.dart' show load, env;
+
+void main() async {
+  load();
+  var sdk = await createSdk();
+  await runLooks(sdk);
+  await runDashboardApis(sdk);
+  await runConnectionApis(sdk);
+}
```
jkaster, line 12: Very handy ref. Thanks!

dart/looker_sdk/example/main.dart
```diff
@@ -0,0 +1,108 @@
...
+
+Future<void> runLooks(LookerSDK sdk) async {
+  try {
+    var looks = await sdk.ok(sdk.allLooks());
+    if (looks.isNotEmpty) {
+      for (var look in looks) {
+        print(look.title);
+      }
+      var look = await sdk.ok(sdk.runLook(looks[looks.length - 1].id, 'png'));
```
jkaster, line 29: FWIW if you want to have a test for binary payloads, `content_thumbnail` is pretty quick for PNG

dart/looker_sdk/lib/src/rtl/api_settings.dart
```diff
@@ -0,0 +1,55 @@
...
+        : null;
+  }
+
+  bool isConfigured() {
+    return _baseUrl != null;
+  }
+
+  void readConfig(String section) {
+    throw UnimplementedError('readConfig');
+  }
```
jkaster, line 30: FWIW I think this could be basically the same as the credentialsCallback for consistency with other SDKs

dart/looker_sdk/lib/src/rtl/api_types.dart
```diff
@@ -0,0 +1,18 @@
+class DelimList<T> {
```
jkaster, line 1: I like that `DelimList` is shorter than `DelimArray`. Renaming in other SDKs is probably not worth the breakage.

dart/looker_sdk/lib/src/rtl/auth_token.dart
```diff
@@ -0,0 +1,42 @@
+class AuthAccessToken {
+  final String _accessToken;
+  final String _tokenType;
+  final int _expiresIn;
+
+  AuthAccessToken.fromResponse(Map map, String contentType)
+      : _accessToken = map['access_token'],
+        _tokenType = map['token_type'],
+        _expiresIn = map['expires_in'];
```
jkaster, line 9: Do you think you'll want the refresh token in case someone wants to implement OAuth flow? Or worry about it then?

dart/looker_sdk/lib/src/rtl/constants.dart
```diff
@@ -0,0 +1,5 @@
+const agentPrefix = 'DART-SDK';
+const defaultApiVersion = '3.1';
```
jkaster, line 2:
```suggestion
const defaultApiVersion = '4.0';
```

dart/looker_sdk/lib/src/rtl/sdk.dart
```diff
@@ -0,0 +1,21 @@
...
+    _sdk = LookerSDK(authSession);
+    return _sdk;
+  }
+
+  static LookerSDK getSdk() {
+    if (_sdk == null) {
+      throw Exception('SDK not initialized');
+    }
+    return _sdk;
+  }
```
jkaster, line 20: Handy when we only need 1 API supported

dart/looker_sdk/lib/src/rtl/transport.dart
```diff
@@ -0,0 +1,147 @@
...
+        return handleResponse(response, responseHandler);
+      case HttpMethod.delete:
+        var response = await http.delete(fullPath, headers: headers);
+        return handleResponse(response, responseHandler);
+      case HttpMethod.put:
+        var response = await http.put(fullPath, headers: headers, body: body);
+        return handleResponse(response, responseHandler);
+      case HttpMethod.head:
+        throw UnimplementedError('head');
+    }
```
jkaster, line 102: If you're basing this on what I did for TypeScript, I did it this way because I wasn't clever enough with TypeScript and the request lib to abstract the method and params into a single call

dart/looker_sdk/pubspec.yaml
```diff
@@ -0,0 +1,11 @@
+name: looker_sdk
+version: 0.0.1
+environment:
+  sdk: '>=2.9.0 <3.0.0'
```
jkaster, line 4: Curious what this does. Does this mean use 2.9.x or greater but not 3.x?

jkaster (changes requested):
Looking good. Nothing major but some minor tweaks.

dart/looker_sdk/pubspec.yaml
bryans99: Again very old stuff. My recollection is that this was mainly driven by flutter but I believe your summation is correct.

neiljaywarner: @bryans99  would we expect this SDK branch to be the best choice for using looker in flutter mobile apps, or platform calls to swift/kotlin SDKs, or rest api calls, or embedded urls, or just depending on things?

PR #934, opened 2022-01-11
chore: test JSON deserialization behavior for TS

jkaster:
Verifies that whatever your JSON payload says, that's what you get in the TypeScript SDK (as expected)


bryans99 (approved):
LGTM

PR #939, opened 2022-01-13
chore: refactor handling of LOOKERSDK_CLIENT_ID and _SECRET

drstrangelooker:
to work better with matrix.looker

Versions prior to 21_18 all used a version specific `LOOKERSDK_CLIENT_ID` and `LOOKERSDK_CLIENT_SECRET`. The would be shared by appending something like `__21_16` to `LOOKERSDK_CLIENT_ID` and `LOOKERSDK_CLIENT_SECRET`. Every time a new version was released and new set of secrets needed to be created and added to the proper repositories.

Beginning with 21_18, the secrets stay constant. This update means that regardless of whether the value of `${{ matrix.looker }}` is `21_18`, `21_20`, `22_0`, etc. the secrets for `21_18` will be used. For values of prior to `21_18`, that value is returned. That way we don't have to make any changes to the secrets stored in Github.


drstrangelooker:
I'm using this pattern with gzr. Basically, I wrote a script that determines what version of the secrets to use for a given Looker version so we don't have to add additional secrets to the repo.

.github/workflows/python-ci.yml
```diff
@@ -174,12 +174,14 @@ jobs:
           python -m pip install --upgrade pip
           pip install tox
 
-      - name: Set LOOKERSDK_CLIENT_{ID,SECRET}
-        # api client/secret stored as github secret per release, e.g. LOOKERSDK_CLIENT_ID__7_20
+      - name: Determine credentials version
         run: |
-          echo "LOOKERSDK_CLIENT_ID=${{ secrets[format('LOOKERSDK_CLIENT_ID__{0}', matrix.looker)] }}" >> $GITHUB_ENV
-          echo "LOOKERSDK_CLIENT_SECRET=${{ secrets[format('LOOKERSDK_CLIENT_SECRET__{0}', matrix.looker)] }}" >> $GITHUB_ENV
+          echo "CREDENTIALS_VERSION=$(${{ github.workspace }}/.github/scripts/parse_version.bash <<< ${{ matrix.looker }})" >> $GITHUB_ENV
```
joeldodge79, line 179: maybe a quick comment as to why we even need to worry about secret versions and a TODO to get rid of them once we remove `21_18` as a supported matrix version?

.github/workflows/tssdk-ci.yml
```diff
@@ -139,11 +139,14 @@ jobs:
           yarn
           yarn build
 
-      - name: Set LOOKERSDK_CLIENT_{ID,SECRET}
-        # api client/secret stored as github secret per release, e.g. LOOKERSDK_CLIENT_ID__7_20
+      - name: Determine credentials version
         run: |
-          echo "LOOKERSDK_CLIENT_ID=${{ secrets[format('LOOKERSDK_CLIENT_ID__{0}', matrix.looker)] }}" >> $GITHUB_ENV
-          echo "LOOKERSDK_CLIENT_SECRET=${{ secrets[format('LOOKERSDK_CLIENT_SECRET__{0}', matrix.looker)] }}" >> $GITHUB_ENV
+          echo "CREDENTIALS_VERSION=$(${{ github.workspace }}/.github/scripts/parse_version.bash <<< ${{ matrix.looker }})" >> $GITHUB_ENV
```
joeldodge79, line 144: ditto

joeldodge79 (changes requested):
glad we don't have to add new secrets!

.github/workflows/tssdk-ci.yml
drstrangelooker: done

.github/workflows/python-ci.yml
drstrangelooker: done

.github/scripts/parse_version.bash
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
```
jkaster, line 2: I think it would be worth describing what this does so others who might have to maintain (hopefully never) would know the purpose

.github/workflows/python-ci.yml
```diff
@@ -174,12 +174,24 @@ jobs:
...
-        # api client/secret stored as github secret per release, e.g. LOOKERSDK_CLIENT_ID__7_20
+      - name: Determine credentials version
+        # Prior to 21_18, each version had different credentials and a 
+        # different secret. 21_20 and later all use the same credentials
+        # as 21_18. The parse_version.bash script parses the version and
+        # yields 21_12, 21_14, 21_16 etc for those versions but 21_18 for
+        # anything 21_18, 21_20, 22_0, etc.
+        #
+        # This can be factored out entirely once 21_18 is the earliest
+        # supported looker version.
```
jkaster, line 185: great comment

jkaster (dismissed):
LGTM. Could use another comment, but not a blocker by any means.

.github/scripts/parse_version.bash
drstrangelooker: HA! It should be `/bin/bash`. I'm going to rename the script.

joeldodge79 (approved):

PR #947, opened 2022-01-20
chore: skip ci tests but return success if files not modified

drstrangelooker:


jkaster (approved):
LGTM. I grok the pattern.

Issue #944, opened 2022-01-19
looker-sdk 21.20.0 throws TypeError for python 3.9.10


rowedonalde:
Due to a [change](https://github.com/python/cpython/blame/v3.9.10/Lib/functools.py#L855) in Python's `functools` module that was included in the update from 3.9.9->3.9.10, `looker_sdk` breaks on import. The error seems to be through the `cattr` dependency:

```
  File "/opt/app/src/tvsapi/views.py", line 8, in <module>
    import looker_sdk
  File "/usr/local/lib/python3.9/site-packages/looker_sdk/__init__.py", line 32, in <module>
    from looker_sdk.sdk.api31 import methods, methods as methods31  # noqa:F401
  File "/usr/local/lib/python3.9/site-packages/looker_sdk/sdk/api31/methods.py", line 32, in <module>
    from . import models as mdls
  File "/usr/local/lib/python3.9/site-packages/looker_sdk/sdk/api31/models.py", line 11935, in <module>
    sr.converter31.register_structure_hook(
  File "/usr/local/lib/python3.9/site-packages/cattr/converters.py", line 186, in register_structure_hook
    self._structure_func.register_cls_list([(cl, func)])
  File "/usr/local/lib/python3.9/site-packages/cattr/multistrategy_dispatch.py", line 45, in register_cls_list
    self._single_dispatch.register(cls, handler)
  File "/usr/local/lib/python3.9/functools.py", line 855, in register
      raise TypeError(
TypeError: Invalid first argument to `register()`. ForwardRef('AccessToken') is not a class.
```

rowedonalde:
Relevant discussion in `cattrs` from others who have experienced this outside of `looker_sdk`: https://github.com/python-attrs/cattrs/issues/206

ps-jmorrow:
We're experiencing this issue as well and it's blocking our release via GHA.

joeldodge79:
@ps-jmorrow @rowedonalde Can you pin python at <= 3.9.9 as a workaround till I can take a look?

ps-jmorrow:
That worked as a workaround, thanks.

jeremytchang:
Have you all been able to pull in latest with the PR change and confirm the issue is fixed? https://github.com/looker-open-source/sdk-codegen/pull/949

alekseiloginov:
[...]

PR #949, opened 2022-01-21
fix: restrict python version and update cattrs version restriction

alekseiloginov:
## Description

We need to unpin `cattrs` version to avoid dependency conflicts with other tools using a later version of it. 

`cattrs` version was [pinned](https://github.com/looker-open-source/sdk-codegen/commit/bf8f3f694efd3a2b7af7d5c849694e0c7bd399c5) to `1.1.2` due to https://github.com/Tinche/cattrs/issues/119. Since the issue should be [fixed](https://github.com/python-attrs/cattrs/issues/119#issuecomment-789757914) in `cattrs` `1.3`, we update the restriction to `>=1.3`.

Also `python` version need to be restricted to `<=3.9.9` due to https://github.com/looker-open-source/sdk-codegen/issues/944. Since `Pipfile` doesn't support qualifiers for `python_full_version`, it's set to `3.9.9`.


python/setup.py
```diff
@@ -38,7 +38,6 @@
     "python-dateutil;python_version<'3.7'",
     # Python 3.7+
     "attrs >= 20.1.0;python_version>='3.7'",
-    "cattrs == 1.1.2;python_version>='3.7'",
```
joeldodge79, line 41: oops, actually we need to still specify a cattrs dep here ([ci failed](https://github.com/looker-open-source/sdk-codegen/runs/4902006917?check_suite_focus=true)).

```
cattrs >= 1.3;python_version>='3.7'
```

should have the fix?

joeldodge79 (changes requested):

joeldodge79:
also, lets change the title to something like the following (to fit conventional commit msg standard)

`fix: restrict python version and remove cattrs version`

python/setup.py
alekseiloginov: Done, also updated `Pipfile` with `cattrs = ">=1.3"`.

jkaster (approved):
LGTM

joeldodge79 (approved):

potiuk:
Just one comment here - pinning python version in a library is a very bad idea. What should people do if they already have python 3.9.10 installed ? This - for example - prevents https://github.com/apache/airflow/pull/20882 from merging.

potiuk:
The consequence of this change is that any other package that would be released with looker SDK released with this limitation will not be installable not only on Python 3.9.10 but also any other version it follows. This is extremely bad practice to upper-bound a Python version IMHO. I have never, ever seen a library doing it. It might be fine for an application if it has specific requirements and they are broken (because that application might simply be ok to refuse working on specific python version) but restricting an SDK/library to not being installable on a python version means that you are preventing any other packages that use looker SDK in this version from even installing on newer Python versions.

joeldodge79:
@potiuk you're correct, it's a less than ideal situation. We're working to resolve the issue but the choice is between pinning it ourselves or letting users find out that it's broken on 3.9.10+

The PR title should have reflected that this is a temporary measure to choose between the lesser of two evils while we find the resources to fix the actual issue

PR #953, opened 2022-01-25
chore: new method to skip required checks

drstrangelooker:
The basic problem we have is that we want to make certain checks required, but the tests for them don't run on all requests. For example this test only runs if the indicated paths are changed.

```
name: TypeScript SDK CI
on:
  pull_request:
    paths:
      - packages/sdk/**
      - packages/sdk-rtl/**
      - packages/sdk-node/**
      - packages/extension-sdk/**
      - packages/extension-sdk-react/**
      - packages/extension-utils/**
```

A special `required_checks_hack_ci` workflow would run. It would just sleep then mark the required checks as succeeded if they hadn't failed by 480 seconds. Every PR had to wait 480 seconds regardless.

My first version of this was to remove the path filters on pull requests and pushes, then use the a "check-files" reusable workflow that would set a bunch of flags based on what files changed. In this snippet, the name of each filter set would be used as a name of an output to this job. So `<job-name>.outputs.apix` could equal 'true' or 'false' based on whether changed files matched the `apix` pattern.
```yaml
      - uses: tony84727/changed-file-filter@v0.2.0
        id: filter
          filters: |
            apix:
              - packages/code-editor/**
              - packages/run-it/**
              - packages/api-explorer/**
              - packages/extension-api-explorer/**
              - packages/extension-utils/**
            codegen:
              - packages/sdk-codegen/**
              - packages/sdk-codegen-utils/**
              - packages/sdk-codegen-scripts/**
            hackathon:
              - packages/wholly-sheet/**
              - packages/hackathon/**
            python:
              - python/**
            resources:
              - bin/looker-resources-index/**
              - docs/resources/**
            tssdk:
              - packages/sdk/**
              - packages/sdk-rtl/**
              - packages/sdk-node/**
              - packages/extension-sdk/**
              - packages/extension-sdk-react/**
              - packages/extension-utils/**
```

Then I would check the flags and skip the certain jobs with code like `if: needs.check_files.outputs.tssdk == 'true'`. I would generate an alternate success message for `if: needs.check_files.outputs.tssdk == 'false'`. Now every test runs every time, generating lots of output when it isn't really needed.

@jeremytchang made the point that this generates lots of noise that is confusing.

To hopefully fix this I reset the state of the CI tests so they have the `path:` filters on pushes and pull requests again. That was commit 378df7694a7d9b02f1d4c67061dddda576883bfd. Now tests once again only run if they are needed. Then I modified the `required_checks_hack_ci` workflow to use the `check-files` logic. If the push or pull_request does not have a file pattern that triggers the actual test, the expression like `if: needs.check_files.outputs.tssdk == 'false'` will succeed and the noop job will run, causing the unneeded required checks to pass almost immediately instead of after 480 seconds.

My one major regret with this pattern is that the file patterns for the checks are defined in two places. I'm not sure if github workflows can work around that, however.

### DIFF
--- a/.github/workflows/apix-ci.yml
+++ b/.github/workflows/apix-ci.yml
@@ -1,10 +1,22 @@
 name: API Explorer CI
 on:
   pull_request:
+    paths:
+      - packages/code-editor/**
+      - packages/run-it/**
+      - packages/api-explorer/**
+      - packages/extension-api-explorer/**
+      - packages/extension-utils/**
 
   push:
     branches:
       - main
+    paths:
+      - packages/code-editor/**
+      - packages/run-it/**
+      - packages/api-explorer/**
+      - packages/extension-api-explorer/**
+      - packages/extension-utils/**
 
   workflow_dispatch:
 
@@ -12,13 +24,8 @@ env:
   TS_JUNIT_OUTPUT_DIR: results/apix
 
 jobs:
-  checkfiles:
-    uses: looker-open-source/.github/.github/workflows/check-files.yml@main
-
   unit:
     name: Unit - ${{ matrix.os }} / Node ${{ matrix.node-version }}
-    needs: checkfiles
-    if: needs.checkfiles.outputs.apix == 'true'
     env:
       JEST_JUNIT_OUTPUT_DIR: results/apix
       JEST_JUNIT_OUTPUT_NAME: ${{ matrix.os }}.apix${{ matrix.node-version }}.xml
@@ -72,28 +79,6 @@ jobs:
         uses: actions/download-artifact@v2
         with:
           path: artifacts
-
-      - name: Publish Unit Test Results
-        uses: EnricoMi/publish-unit-test-result-action@v1.12
-        with:
-          # Cosmetic issue with `check_name` being associated to the wrong
-          # workflow: https://github.com/EnricoMi/publish-unit-test-result-action/issues/12
-          check_name: APIX Tests
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          report_individual_runs: true
-          hide_comments: orphaned commits
-          check_run_annotations_branch: '*'
-          files: 'artifacts/apix-test-results/*.xml'
-
-  publish-test-results-on-skip:
-    needs: checkfiles
-    if: needs.checkfiles.outputs.apix == 'false'
-    runs-on: ubuntu-latest
-
-    steps:
-      - name: Make Dummy Artifacts
-        run: |
-          mkdir -p 'artifacts/apix-test-results'
 
       - name: Publish Unit Test Results
         uses: EnricoMi/publish-unit-test-result-action@v1.12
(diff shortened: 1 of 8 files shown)


drstrangelooker:
This doesn't work since a skipped test is seen as successful. Looking at another alternative.

drstrangelooker:
New system figured out. If the CI test won't run, I use the REST API to manually create the required "check" so that the process passes.

.github/workflows/apix-ci.yml
```diff
@@ -52,6 +59,23 @@ jobs:
           yarn build
           yarn dedupe:ci
 
+      # if this job fails before this point the required state check "APIX Tests"
+      #  is never set, so we will set it manually.
+      - name: Report Failure
+        run: |
+          curl --request POST \
```
jkaster, line 66: Love the hack

.github/workflows/required-checks-hack-ci.yml
```diff
@@ -0,0 +1,135 @@
...
+              - packages/extension-utils/**
+      - name: Debug info
+        run: |
+          echo "${{ toJSON(steps.filter.outputs) }}"
+          echo "apix: ${{ steps.filter.outputs.apix }}"
+          echo "codegen: ${{ steps.filter.outputs.codegen }}"
+          echo "hackathon: ${{ steps.filter.outputs.hackathon }}"
+          echo "python: ${{ steps.filter.outputs.python }}"
+          echo "resources: ${{ steps.filter.outputs.resources }}"
+          echo "tssdk: ${{ steps.filter.outputs.tssdk }}"
```
jkaster, line 75: so we'll need to update this whenever we add another CI workflow. Good to note somewhere

docs/ci-testing.md
```diff
@@ -0,0 +1,206 @@
+# Adding required ci testing
+## Create a test-area-ci workflow
+Create a workflow in `.github/workflows/`. The existing workflows can be used as an example.
```
jkaster, line 3: Nice addition!

.github/workflows/required-checks-hack-ci.yml
jkaster: Oh, I see you're way ahead of me. Nice.

jkaster (approved):
LGTM

.github/workflows/required-checks-hack-ci.yml
```diff
@@ -0,0 +1,135 @@
+# The jobs in this workflow are necessary because they are configured as
+# required checks on PRs. Github does not currently support configuring required
+# checks per workflow. A PR that only triggers codegen-ci.yml jobs will only
+# report back with the "Codegen Tests" check. But because there is only a global
+# option for required PR checks, that PR would otherwise never satisfy the
+# "Typescript/Python/APIX Tests" checks without these noop jobs.
```
jeremytchang, line 6: This paragraph really helped me understand the underlying issue. A monorepo does not work well with Github required checks that are repo-wide.

jeremytchang (approved):
LGTM! Thanks for your attention to making the PR check output more readable. One hopes Github will allow required checks per workflow.

PR #966, opened 2022-02-03
feat: SDK for Looker 22_2

drstrangelooker:


joeldodge79:
this is looking good. My only thought is maybe to wait towards the end of the week. That way we give the 22.0 packages we just released a little time to bake and crop up anything we really want to patch fix on them. Also, I'm not sure what timing we try to aim for in releasing the SDKs versus when the Looker version is deployed, I think we usually wait a few weeks?

drstrangelooker:
👍

joeldodge79 (approved):

PR #972, opened 2022-02-07
feat(python rtl): add optional config dictionary init parameter

kevin-wallace-wpro:
Hi all,

We need to talk to multiple Looker instances at once. It is difficult to do this reliably with the current configuration options. This PR adds an optional dictionary parameter containing config values to pass to the `init` functions.

If this is an acceptable approach, I can update the tests and docs.

Thank you!


joeldodge79:
@jkaster doesn't our TS SDK also have the ability to receive configuration from an object rather than reading from a file or envvars? can you point us to that code so we can try to follow the same pattern of where to pass in that config?

jkaster:
I believe you're referring to [`readConfig()`](https://github.com/looker-open-source/sdk-codegen/blob/ae20ad316c3a5e155b952bc92c9050436e5e78a6/packages/sdk-rtl/src/apiSettings.ts#L64) and you can see an implementation in the [OAuth tests](https://github.com/looker-open-source/sdk-codegen/blob/ae20ad316c3a5e155b952bc92c9050436e5e78a6/packages/sdk-rtl/src/oauthSession.spec.ts#L45-L53)

joeldodge79:
ok, this is slightly different. @kevin-wallace-wpro is proposing being able to pass in a bag of config properties into the `ApiSettings` constructor and use it as a final-override (so config file < env-var < bag-o-props) in `ApiSettings.read_config`. I think looks ok, thoughts?

jkaster:
It's certainly an option. The difference would be that the credentials are not fetched on demand and discarded when they are no longer needed. `readConfig()` is intended to be a dynamic point in time fetch without storing any creds beyond authentication token retrieval.

kevin-wallace-wpro:
Thanks for the quick review.

> The difference would be that the credentials are not fetched on demand and discarded when they are no longer needed. readConfig() is intended to be a dynamic point in time fetch without storing any creds beyond authentication token retrieval.

Just to note this PR doesn't change anything about how the python sdk stores the information, the values read from the dict are the same as those from the env and are handled the same way.

joeldodge79:
We talked about it and the consensus is that rather than modify the provided `ApiSettings` implementation we'd prefer folks roll their own if they'd like to provide alternative behavior (like retrieving config from a provided dictionary). 

When you implement your own version of `ApiSettings` (quite possibly subclassing the provided one and overriding `read_config`), note that at a minimum it must implement the `PApiSettings` protocol (and be sure that `read_config` returns at least a client_id and client_secret in it's `Dict[str, str]` when called, it can return other stuff too but the properties defined in `PTransportSettings` must be accessible as instance properties)

One change we could definitely make here is to add a `settings=None` parameter to `looker_sdk.init40` (and init31) to save you the boiler plate of having to manually build the SDK instance with your settings class (otherwise you'll have to copy/paste what those methods do using your custom settings class).

Does that make sense?

PR #974, opened 2022-02-08
fix: clean up package dependencies

jkaster:
we had a stale version in a package dependency


joeldodge79 (approved):
we could add this package to `release-please-config.json` and have it's versioning and dependencies auto-updated (without actually publishing it to npm)

PR #975, opened 2022-02-08
feat: Implemented Go SDK env var handling

jeremytchang:
Also added unit tests for it that will leave existing environment variables undisturbed. (Will store env vars before test and restore env vars after test)

### DIFF
--- a/go/rtl/settings.go
+++ b/go/rtl/settings.go
@@ -3,6 +3,9 @@ package rtl
 import (
 	"fmt"
 	"gopkg.in/ini.v1"
+	"os"
+	"strconv"
+	"strings"
 )
 
 var defaultSectionName string = "Looker"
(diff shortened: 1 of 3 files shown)


go/rtl/settings.go
```diff
@@ -37,3 +44,32 @@ func NewSettingsFromFile(file string, section *string) (ApiSettings, error) {
...
+
+	if v, present  := os.LookupEnv(baseUrlEnvKey); present {
+		settings.BaseUrl = v
+	}
+	if v, present  := os.LookupEnv(apiVersionEnvKey); present {
+		settings.ApiVersion = v
+	}
+	if v, present  := os.LookupEnv(verifySslEnvKey); present {
+		s := strings.ToLower(v)
+		settings.VerifySsl = s == "true" || s == "t" || s == "1" || s == "y" || s == "yes"
```
jkaster, line 59: fwiw other SDKs have either an `isBool()` or `defaultBool()` type of helper function. In this case, I think we want to default `settings.VerifySSL` to `true`

jkaster (approved):
Looks good. Do you want to update the GoLook readme for this now, or in another PR?

go/rtl/settings.go
jeremytchang: Oh good catch! I will fix.

jeremytchang: I just reread the spec. My logic is correct. It already has setting default to true. Only if it is set does it change.
```
true, t, yes, y, or 1 (case insensitive) to enable SSL verification. Any other value is treated as false. Defaults to true if not set.
```

PR #976, opened 2022-02-09
chore: move specs into redux and cleanup oauth related code

josephaxisa:
- Significantly simplified API Explorer's OAuth flow
- Specs now live in a redux slice 
- Spec fetching is now only done through `ApixEnvAdaptor` and `ApixExtensionEnvAdaptor` 
- Routes now drive the state. This addresses issues where DiffScene state (i.e based on the base and comparison spec keys) was out of sync with the url


packages/api-explorer/src/StandaloneApiExplorer.tsx
```diff
@@ -26,79 +26,64 @@
...
+  const [browserAdaptor] = useState(
+    new ApixAdaptor(initRunItSdk(), window.origin)
+  )
+  const location = useLocation()
+  const oauthReturn = location.pathname === `/${oAuthPath}`
+  const sdk = browserAdaptor.sdk
+  const canLogin =
+    (sdk.authSession.settings as OAuthConfigProvider).authIsConfigured() &&
+    !sdk.authSession.isAuthenticated() &&
+    !oauthReturn
```
josephaxisa, line 57: In a following PR I intend to make this cleaner and delegate more logic to OAuthScene

jkaster (approved):
LGTM. Verified stand-alone OAuth behavior and diff link. Only nit is that RunIt is staying open when perhaps it shouldn't, but that's minor. This is a big step forward.

PR #980, opened 2022-02-11
fix: Update go codegen to set object as interface{}

jeremytchang:
Currently the Go SDK Models will break on build because the type in params is set to `*struct` which is not valid. For now set object to `interface{}`. Could do `map[string]interface{}` in the future. 

This change is needed to get integration tests to work.

Generated 4.0 Go model too.


jkaster (approved):
LGTM. Does this mean the Go `v3` folder is no longer being generated, too? Was that an incidental cleanup you got resolved to get the generator running correctly?

jeremytchang:
@jkaster The Go model generation issue stemmed from sdk-codegen script package.json version was incorrect. I had made Jax's change locally. https://github.com/looker-open-source/sdk-codegen/pull/976/files#diff-4b80bf2b50cec2975b7466079f3acd1355591534fd0adc381a257267a682b5d9

PR #982, opened 2022-02-14
chore: revert version numbers accidentally updated in package.json files

drstrangelooker:


joeldodge79 (commented):
got it, let's hold off on this for now though

joeldodge79 (changes requested):
Let's drop this now

drstrangelooker:
drop as in kill this, or drop as in merge it in? :-)

PR #984, opened 2022-02-14
chore: add permissions so use of github api is ok regardless of submitter

drstrangelooker:
By default the `secrets.GITHUB_TOKEN` does not have the permissions to write the static of checks. In order to make the API call work regardless of who submitted the PR, a stanza is needed in the yaml file to request the correct permissions.

```yaml
permissions:
  checks: write
```


.github/workflows/required-checks-hack-ci.yml
```diff
@@ -84,7 +87,7 @@ jobs:
         run: |
           curl --request POST \
           --url https://api.github.com/repos/looker-open-source/sdk-codegen/check-runs \
-          --header "authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+          --header "authorization: token ${{ secrets.GITHUB_TOKEN }}" \
```
joeldodge79, line 90: why this change?

Also, at some point you might checkout [actions/github-script](https://github.com/marketplace/actions/github-script#comment-on-an-issue) to use the octocat sdk instead of using curl

joeldodge79 (approved):
non-blocking question/suggestion

.github/workflows/required-checks-hack-ci.yml
drstrangelooker: I was running down why certain runs were failing with a 403. I found token instead of Bearer in the api docs and decided to use it to see if that was the problem. It wasn't but it matches the github api docs more closely.

drstrangelooker: Hmm. I'm looking at github-script now. I'm always torn on these. Using curl means that anyone looking at this knows what is happening.

PR #989, opened 2022-02-15
fix: Fixed incorrect VerifySSL setting implementation

jeremytchang:
The original implementation was incorrect. Also added comment/doc to help avoid confusion if someone sets custom transport setting outside of the lookersdk verify ssl setting.


jeremytchang:
Also I'd like to eliminate `NewAuthSessionWithTransport`. I feel it muddles our Go SDK interface, and the comment I've added outlines the issue. However since the function already exists and is public it may break downstream dependencies.

PR #996, opened 2022-02-18
feat: python rtl optional config ApiSettings init parameter

kevin-wallace-wpro:
Supersedes #972 

This PR adds an optional `config_settings` parameter to the python sdk init functions. It is up to the sdk user to provide a valid implementation of the `ApiSettings` class.


joeldodge79 (changes requested):
Fantastic! besides my suggested readme change I have one more request to update [this code](https://github.com/looker-open-source/sdk-codegen/blob/42823364d20d5b396c032cc91282f3632e86012f/python/looker_sdk/rtl/api_settings.py#L41)


1. ideally some typing foo that specifies the return dict must have `client_id: str` and `client_secret: str` keys but may have any other keys as well
2. if that's not easily available then at least comment (or docstring?) on that method declaration indicating that those keys are necessary. [PEP665](https://www.python.org/dev/peps/pep-0655/) would have been good but we ned a solution that supports at least 3.7 (and ideally 3.6)

python/README.rst
```diff
@@ -132,6 +137,35 @@ example file:
...
+    import looker_sdk
+    from looker_sdk import api_settings
+
+    class MyApiSettings(api_settings.ApiSettings):
+        def __init__(self, *args, **kw_args):
+            self.my_var = kw_args.pop("my_var")
+            super().__init__(*args, **kw_args)
+
+        def read_config(self) -> t.Dict[str, str]:
+            config = super().read_config()
```
joeldodge79, line 155:
```suggestion
            # Return dict must include at least `client_id` and `client_secret` keys
            config = super().read_config()
```

joeldodge79: can you also add this comment to the `api_settings.ApiSettings.read_config` method if it's not clearly indicated there? This would be in addition to updating the `api_settings.PApiSettings.read_config` declaration I already mentioned

python/looker_sdk/rtl/api_settings.py
```diff
@@ -32,13 +32,21 @@
...
-    from typing import Protocol
+    from typing import Protocol, TypedDict
 else:
-    from typing_extensions import Protocol
+    from typing_extensions import Protocol, TypedDict
 
+from typing_extensions import Required
+
+class SettingsConfig(TypedDict, total=False):
+    client_id: Required[str]
```
kevin-wallace-wpro, line 42: `Required` is a relatively recent addition to `typing_extensions`, looks like about 3 months ago.

pyright support is there which is nice..

settings-config

python/looker_sdk/rtl/api_settings.py
```diff
@@ -32,13 +32,21 @@
...
+    client_secret: Required[str]
+    base_url: Required[str]
+    verify_ssl: str
+    timeout: str
 
 class PApiSettings(transport.PTransportSettings, Protocol):
-    def read_config(self) -> Dict[str, str]:
+    def read_config(self) -> SettingsConfig:
```
kevin-wallace-wpro, line 49: I've added the type here and my editor was happy with it. Adding it to `read_config` would probably require further code changes, probably not worth it?

kevin-wallace-wpro (commented):
Updated comments and added type for `read_config`

python/looker_sdk/rtl/api_settings.py
joeldodge79: cool! I'm approving action runs so we can see how this goes on all the python versions we support

python/looker_sdk/rtl/api_settings.py
```diff
@@ -32,13 +32,21 @@
...
 else:
-    from typing_extensions import Protocol
+    from typing_extensions import Protocol, TypedDict
 
+from typing_extensions import Required
+
+class SettingsConfig(TypedDict, total=False):
+    client_id: Required[str]
+    client_secret: Required[str]
+    base_url: Required[str]
```
joeldodge79, line 44: `base_url` is technically not required. It's required to be an instance property on the settings object and the default implementation gets it from `read_config` but that's not required. I know it's confusing - honestly the only reason `read_config` exists is so that our default implementation doesn't have to store `client_id` or `client_secret` in python memory for security purposes. Otherwise they'd just be required instance properties too and there would be no `read_config`

python/looker_sdk/rtl/api_settings.py
joeldodge79: let's see how the mypy and unittest runs go. If `TypeDict.Required` is available on all our supported python versions then it actually would be nice (actually probably required) to update the default implementation's type signature

joeldodge79 (commented):
You'll need to sign the Google CLA to get this in (see the google/cla check that's failing and click details for instructions on how to sign it)

joeldodge79:
yup, [mypy run](https://github.com/looker-open-source/sdk-codegen/runs/5297688728?check_suite_focus=true) failed with `looker_sdk/rtl/api_settings.py:104: error: Return type "Dict[str, str]" of "read_config" incompatible with return type "SettingsConfig" in supertype "PApiSettings"` before we could run tox tests on other python versions. I'll do a quick experiment to see if that works and is worth pursuing the `Required` stuff

joeldodge79:
If https://github.com/looker-open-source/sdk-codegen/pull/1005 passes CI then I think it would be great if you could fix all the mypy errors so that we can use `def read_config(self) -> SettingsConfig`

joeldodge79:
> If #1005 passes CI

[failed with](https://github.com/looker-open-source/sdk-codegen/runs/5297804600?check_suite_focus=true)
```
tests/conftest.py:8: in <module>
    import looker_sdk  # noqa: E402
looker_sdk/__init__.py:[25](https://github.com/looker-open-source/sdk-codegen/runs/5297804600?check_suite_focus=true#step:5:25): in <module>
    from looker_sdk.rtl import api_settings
looker_sdk/rtl/api_settings.py:40: in <module>
    from typing_extensions import Required
E   ModuleNotFoundError: No module named 'typing_extensions'
```

provided some relevant guidance [inline](https://github.com/looker-open-source/sdk-codegen/pull/996/files#r812956730) I appreciate your effort on this!

python/looker_sdk/rtl/api_settings.py
```diff
@@ -32,13 +32,21 @@
 from looker_sdk.rtl import transport
 
 if sys.version_info >= (3, 8):
-    from typing import Protocol
+    from typing import Protocol, TypedDict
 else:
-    from typing_extensions import Protocol
+    from typing_extensions import Protocol, TypedDict
 
+from typing_extensions import Required
```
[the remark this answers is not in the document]
kevin-wallace-wpro: I tried with 3.11.0a4 but `Required` wasn't in there.

So I've

1. kept the import coming from `typing_extensions`
2. updated setup.py to require typing-extensions >= 4.1.1 regardless of the python version

I think that is the best way? Note we still can't use 3.11 anyway because of #944 (I pulled Required from typing_extensions running 3.11 and got the error referenced in that issue).

python/looker_sdk/rtl/api_settings.py
```diff
@@ -122,6 +135,21 @@ def _bool(val: str) -> bool:
             raise TypeError
         return converted
 
+    def _override_settings(
```
kevin-wallace-wpro, line 138: We can avoid this with some `cast`s, but I couldn't get anything else to infer the correct types without explicitly using the SettingsConfig keys.

python/looker_sdk/rtl/api_settings.py
```diff
@@ -93,23 +102,27 @@ def __init__(
         if sdk_version:
             self.agent_tag += f" {sdk_version}"
 
-    def read_config(self) -> Dict[str, str]:
+    def read_config(self) -> SettingsConfig:
         cfg_parser = cp.ConfigParser()
+        data: SettingsConfig = {
```
kevin-wallace-wpro, line 107: If the user doesn't end up settings these they'll run into this [exception](https://github.com/looker-open-source/sdk-codegen/blob/d5b0c999727e6065723ac6becbb5c53e98550387/python/looker_sdk/rtl/auth_session.py#L136)

python/looker_sdk/rtl/api_settings.py
joeldodge79: 👍

joeldodge79 (changes requested):
Looking good!

I don't know why I always have to approve your check runs. thought that was a one-time thing. anyway, getting closer but `mypy` still failed:
```
looker_sdk/rtl/auth_session.py:276: error: Expression has type "Any"
looker_sdk/rtl/auth_session.py:276: error: TypedDict "SettingsConfig" has no key "redirect_uri"
looker_sdk/rtl/auth_session.py:277: error: Expression has type "Any"
looker_sdk/rtl/auth_session.py:277: error: TypedDict "SettingsConfig" has no key "looker_url"
looker_sdk/rtl/auth_session.py:286: error: Expression type contains "Any" (has type "Tuple[str, Any]")
looker_sdk/rtl/auth_session.py:286: error: Expression has type "Any"
looker_sdk/rtl/auth_session.py:292: error: Expression has type "Any"
looker_sdk/rtl/auth_session.py:294: error: Expression has type "Any"
looker_sdk/rtl/auth_session.py:338: error: Expression has type "Any"
looker_sdk/rtl/auth_session.py:349: error: Expression has type "Any"
```
points out some inconsistencies we have - added "OAuthSession" support that needs `redirect_uri` and `looker_url` so we should add those to both `SettingConfig` and `PApiSettings` as optional.

you should be able to check it locally (might need to install requests-types) with:
```
pipenv run mypy looker_sdk/
```

python/looker_sdk/rtl/api_settings.py
kevin-wallace-wpro: hm actually I see now there is some use of `cast` elsewhere. Personally I like the `cast` solution better, less code changes and personally I feel the code is more future-proof, this way we could easily miss keys we want.

I'll put up a version with that so you can compare

kevin-wallace-wpro:
> points out some inconsistencies we have - added "OAuthSession" support that needs redirect_uri and looker_url so we should add those to both SettingConfig and PApiSettings as optional.

Ah whoops ok I'll take a look at it. I was actually running mypy locally but there were quite a few complaints about `Any` so I just figured these were already there and acceptable without much thought haha.

python/looker_sdk/rtl/api_settings.py
```diff
@@ -146,7 +170,7 @@ def _override_from_env(self) -> Dict[str, str]:
 
         return overrides
 
-    def _clean_input(self, data: Dict[str, str]) -> Dict[str, str]:
+    def _clean_input(self, data: SettingsConfig) -> SettingsConfig:
```
kevin-wallace-wpro, line 173: Reverted back to more or less the original version here with a `cast` at the end.

kevin-wallace-wpro (commented):
Ok added `redirect_uri` and `looker_url` to SettingsConfig. Simplified `_override_settings` to just loop over SettingsConfig's keys (we could do the same thing in _override_from_env if you want?). Reverted _clean_input to original impl.

python/looker_sdk/rtl/api_settings.py
joeldodge79: works for me!

joeldodge79 (approved):
Nice work! really appreciate it.

mypy and unittests are passing. our integration tests won't run/pass because they use a secret that doesn't transfer to forked repos. So I'm gonna update this branch and then override the failed integration test checks and merge it in. If, for some reason, our integration tests fail I'll forward fix them

PR #1000, opened 2022-02-21
feat: initSdk helper fn in @looker/extension-utils

josephaxisa:


packages/extension-utils/src/authUtils.ts
```diff
@@ -108,9 +115,46 @@ export class OAuthConfigProvider extends ApiSettings {
...
+
+export interface InitSdkSettings {
+  /** agent tag to use for the SDK requests */
+  agentTag: string
+  /** Local storage key for retrieving auth configuration settings */
+  configKey: string
+  /** Id used to register this application on the Looker server */
+  clientId: string
+  /** A callback function that returns an sdk instance */
+  createSdkCallback: (session: BrowserSession) => APIMethods
```
jkaster, line 133: Can this return `IAPIMethods` instead? Is there something in `APIMethods` that's not documented in `IAPIMethods`?

jkaster (changes requested):
This looks really good, but I think you should be able to use `IAPIMethods` instead of `APIMethods` for the method contract

PR #1018, opened 2022-03-11
ci: change workflows to use reusable action for looker versions

drstrangelooker:
The looker versions for CI are all in one place now.

### DIFF
--- a/.github/workflows/go-ci.yml
+++ b/.github/workflows/go-ci.yml
@@ -65,7 +65,7 @@ jobs:
           path: ${{ env.JUNIT_OUTPUT_DIR }}
 
   setup:
-    uses: looker-open-source/reusable-actions/.github/workflows/supported-versions.yml@v0.0.1
+    uses: looker-open-source/reusable-actions/.github/workflows/supported-versions.yml@main
   integration:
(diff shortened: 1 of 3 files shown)


jkaster (approved):
LGTM. Nice improvement

drstrangelooker:
@jkaster Typescript tests fail with 22.4 on an issue with strings passed where interferes are expected.

I'm thinking that I'll override the required check then we'll do a separate PR to change the tests.

drstrangelooker:
@jkaster I backed the supported-versions array to `["21.18", "21.20", "22.0", "22.2"]` for now.

I also changed the action to reference `supported-versions@main` instead of `supported-versions@v0.0.1`. I realized if I used a specific tag, I would need to change `sdk-codegen` every time the version array was changed. That was the exact opposite of what we wanted. So I needed to reference the main branch instead.

I also created tags `supported-versions-v22.2` and `supported-versions-v22.4` to allow a caller to specify the max version they want to deal with.

jkaster:
> @jkaster Typescript tests fail with 22.4 on an issue with strings passed where interferes are expected.

I think you mean "integers" above, and I'll look at that.

PR #1020, opened 2022-03-15
chore: update unit test for string ID types

jkaster:
Unit test tweak to support Looker 22.4 API type changes

### DIFF
--- a/packages/sdk-codegen-scripts/src/exampleMiner.spec.ts
+++ b/packages/sdk-codegen-scripts/src/exampleMiner.spec.ts
@@ -75,7 +75,7 @@ describe('example mining', () => {
     it('getRemoteHttpOrigin', () => {
       const actual = getRemoteHttpOrigin()
       expect(actual).toEqual(
-        'https://github.com/looker-open-source/sdk-codegen'
+        'https://github.com/looker-open-source/sdk-codegen.git'
       )
     })
   })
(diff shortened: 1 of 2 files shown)


packages/sdk-codegen-scripts/src/exampleMiner.spec.ts
```diff
@@ -75,7 +75,7 @@ describe('example mining', () => {
     it('getRemoteHttpOrigin', () => {
       const actual = getRemoteHttpOrigin()
       expect(actual).toEqual(
-        'https://github.com/looker-open-source/sdk-codegen'
+        'https://github.com/looker-open-source/sdk-codegen.git'
```
joeldodge79, line 78: CI doesn't like this: [action output](https://github.com/looker-open-source/sdk-codegen/runs/5547364560?check_suite_focus=true)

packages/sdk-node/test/methods.spec.ts
```diff
@@ -754,7 +754,7 @@ describe('LookerNodeSDK', () => {
           const query = await sdk.ok(sdk.create_query(request))
           expect(query).toBeDefined()
           expect(query.id).toBeDefined()
-          expect(query.id).toBeGreaterThan(0)
+          expect(Number(query.id)).toBeGreaterThan(0)
```
joeldodge79, line 757: hm, now that IDs really are strings, maybe

```suggestion
          expect(query.id.length).toBeGreaterThan(0)
```

joeldodge79 (changes requested):

packages/sdk-node/test/methods.spec.ts
jkaster: TypeScript didn't like that so I'm using a `String()` cast

packages/sdk-codegen-scripts/src/exampleMiner.spec.ts
jkaster: modified the test statement

Issue #795, opened 2021-08-11
Golang SDK v3 does not return error context from API


Jberlinsky:
When issuing a request to the Golang API that triggers an HTTP 4XX error (e.g. a 422 validation error), the Swagger API will return useful error context, e.g.:

```
{
  "message": "Validation Failed",
  "errors": [
    {
      "field": "dialect",
      "code": "missing_field",
      "message": "This field is required.",
      "documentation_url": "http://docs.looker.com/"
    },
    {
      "field": "database",
      "code": "missing_field",
      "message": "This field is required.",
      "documentation_url": "http://docs.looker.com/"
    },
    {
      "field": "db_timezone",
      "code": "invalid",
      "message": "Must specify a database timezone when user timezones are activated.",
      "documentation_url": "http://docs.looker.com/"
    },
    {
      "field": "max_connections",
      "code": "invalid",
      "message": "Max connections must be an integer between 5 and 100.",
      "documentation_url": "http://docs.looker.com/"
    }
  ],
  "documentation_url": "http://docs.looker.com/"
}
```

However, the Golang SDK doesn't seem to decode these error messages when a failure is encountered (see https://github.com/looker-open-source/sdk-codegen/blob/main/go/rtl/auth.go#L138). This leaves a user of the SDK in the unfortunate position of having to:

1. Dump the request object to JSON
2. Go to the Looker Swagger UI and enter the transposed JSON request
3. Inspect errors there
4. Make changes in the Golang SDK caller
5. Repeat as needed

This also makes it difficult to write effective integration tests for erroneous states.

jeremytchang:
Hey! This is a great call out. 

We already have generated [Error](https://github.com/looker-open-source/sdk-codegen/blob/main/go/sdk/v4/models.go#L1138) and [ValidationError](https://github.com/looker-open-source/sdk-codegen/blob/main/go/sdk/v4/models.go#L3617) struct types.

We can make use of these types and json decode the error responses. To follow error conventions and not break downstream dependencies, we'd need to define a new Error struct that implements the `error` [interface](https://pkg.go.dev/builtin#error) .

Something like this:
```
type ResponseError struct {
  StatusCode int
  Error v4.Error
  ValidationError v4.ValidationError
}

func (r *ResponseError) Error() string {
	return fmt.Sprintf("...")
}
```

Though this would couple a custom Go SDK typing to our swagger spec. So not ideal, but also it's unlikely the swagger spec errors will change.

PR #1021, opened 2022-03-16
fix: Make Go SDK parse response by result type

jeremytchang:
Changes: 
- Go SDK Parse response based of type defined in go sdk models, the result type. This fixes methods returning string type. This handles all method result types, `string`, `struct`, `interface{}`.
- Added additional integration tests around queries, looks, and
  dashboards. This fills out the basic integration test requirements.
- Added unit tests
- Return error body in addition to error status code for more descriptive response errors.

Ideally we want the content type header to be the source of truth when parsing the response. However, currently, go sdk model generation doesn't support union type for results, so we cannot parse according to Content Type Header. For now, parse based off result type.

Take for example [RunQuery](https://developers.looker.com/api/explorer/4.0/methods/Query/run_query). It returns 4 content type headers `text`, `application/json`, `image/png`, and `image/jpeg`. Result types should be `string`, `map[string]interface{}`, `byte[]`. Right now result type is only `string` [here](https://github.com/looker-open-source/sdk-codegen/blob/main/go/sdk/v4/methods.go#L5002). Passing in a string type to the decoder means we cannot parse to `map[string]interface{}` or `byte[]`.
In this case, it is an antipattern to force the Go SDK caller to cast the `string` result to what they need, but it at least works.

For the future:
We need to update sdk model generation to use interface{} (any) for union types, and then use Content Type Header to parse response accordingly. This will break downstream dependencies. [Issue tracking this](https://github.com/looker-open-source/sdk-codegen/issues/1022).

### DIFF
--- a/go/rtl/auth.go
+++ b/go/rtl/auth.go
@@ -137,11 +137,26 @@ func (s *AuthSession) Do(result interface{}, method, ver, path string, reqPars m
 	defer res.Body.Close()
 
 	if res.StatusCode < 200 || res.StatusCode > 226 {
-		return fmt.Errorf("response error: %s", res.Status)
+		b, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+				return fmt.Errorf("response error. status=%s. error parsing error body", res.Status)
+		}
+
+		return fmt.Errorf("response error. status=%s. error=%s", res.Status, string(b))
 	}
 
-	extra.RegisterFuzzyDecoders()
-	err = json.NewDecoder(res.Body).Decode(&result)
+	// TODO: Make parsing content-type aware. Required change to go model generation to use interface{} for all union types. Github Issue: 
+	switch v := result.(type) {
+	case *string:
+			b, err := ioutil.ReadAll(res.Body)
+			if err != nil {
+					return err
+			}
+			*v = string(b)
+	default:
+		extra.RegisterFuzzyDecoders()
+		return json.NewDecoder(res.Body).Decode(&result)
+	}
 
 	return nil
 }
--- a/go/rtl/auth.go
+++ b/go/rtl/auth.go
@@ -145,7 +145,8 @@ func (s *AuthSession) Do(result interface{}, method, ver, path string, reqPars m
 		return fmt.Errorf("response error. status=%s. error=%s", res.Status, string(b))
 	}
 
-	// TODO: Make parsing content-type aware. Required change to go model generation to use interface{} for all union types. Github Issue: 
+	// TODO: Make parsing content-type aware. Requires change to go model generation to use interface{} for all union types.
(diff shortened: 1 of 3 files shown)


jeremytchang:
@jkaster The TS SDK exhibits the antipattern, returning string for text/json/image content types. https://github.com/looker-open-source/sdk-codegen/blob/main/packages/sdk/src/4.0/methods.ts#L8253

jkaster:
> @jkaster The TS SDK exhibits the antipattern and returns string for text/json/image content types. https://github.com/looker-open-source/sdk-codegen/blob/main/packages/sdk/src/4.0/methods.ts#L825

Not quite. See the existing tests that verify the image payloads for various SDK methods. The `string` in that case is a byte array a UTF-8 decoded string.

jkaster:
For more strictly typed languages, you may need to indicate either byte array or string as the return type. For kotlin and swift, I hacked together a solution we've never cleaned up that implements binary payloads via the "streaming" parallel functions for the SDKs.

go/rtl/auth.go
```diff
@@ -137,11 +137,27 @@ func (s *AuthSession) Do(result interface{}, method, ver, path string, reqPars m
 	defer res.Body.Close()
 
 	if res.StatusCode < 200 || res.StatusCode > 226 {
-		return fmt.Errorf("response error: %s", res.Status)
+		b, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+				return fmt.Errorf("response error. status=%s. error parsing error body", res.Status)
+		}
+
+		return fmt.Errorf("response error. status=%s. error=%s", res.Status, string(b))
```
panoskoug, line 145: I am wondering if this could break any code that relies on the string.

panoskoug (approved):

jkaster:
This is a major improvement. Can you add a test for content_thumbnail that verifies the PNG and SVG payload? See the TS tests for PNG bytestream verification code.

go/rtl/auth.go
jeremytchang: Yeah this could, though consumers _reallllllly_ shouldn't be matching on strings. I'll be implementing better error context in the future so consumers can match on the status code and/or specific errors. https://github.com/looker-open-source/sdk-codegen/issues/795#issuecomment-1069742721

The possibility of breakage is not great, but while we try to fully support the Go SDK, consider it in beta and we will provide best effort to not break downstream dependencies.

jeremytchang:
@jkaster Added png/svg integration test. Made use of golang standard library for checking. https://pkg.go.dev/net/http#DetectContentType

jkaster (approved):
This looks great. Thanks for adding the PNG/SVG tests. Nice that Go has a MIME type detection function

jeremytchang:
@panoskoug Thanks for going out of your way for the approval!

PR #1024, opened 2022-03-17
feat: Looker 22.4 API updates

drstrangelooker:


drstrangelooker:
python is not working on my machine so I can't do local tests.

drstrangelooker:
@jkaster I'm having an error pop up in packages/sdk-node/test/methods.spec.ts:1059 during the TSSDK-ci runs. Not sure how to handle it.

jkaster:
> I'm having an error pop up in packages/sdk-node/test/methods.spec.ts:1059 during the TSSDK-ci runs. Not sure how to handle it.

I can look at it now

jkaster:
I can't reproduce this locally. Ran those tests for 21.18, 22.0, 22.4. If this persists we may need to skip the test

jkaster:
There's something off with this branch. PR #1025 passes

joeldodge79:
is this the looker bindings release for 22.4? if so, we may want to think about how it shows up in the changelogs. I used to make them like `feat: Looker 21.12 bindings` that way they show up under the "Features" section in teh changelog. We need to hard-code the version anyway in release-please-config.json so having it be a `fix:` doesn't really buy us anything and it makes it look like a bugfix :-)

joeldodge79 (changes requested):
oh, hm. let's try to fix the python situation so `black` formats the files correctly. otherwise it looks like they changed a whole lot when in fact they didn't. happy to help with that (either help get python working for you or I can just format them myself and push to this branch)

joeldodge79 (commented):
Python changes now LGTM (@jkaster even after your file formatting commit the methods.py files weren't `black` formatted correctly so I just added a commit formatting those - we may need to dig into why your `yarn gen py` didn't `black` format those files)

I'll let someone else approve the PR (for other languages etc)

joeldodge79 (approved):

PR #1025, opened 2022-03-17
fix: declaration miner (internal utility)

jkaster:
Turns out bazel build files have weird `stat`s that break node fs interrogation and hence break declaration miner.

This ignores errors reported on directory checking for those offending paths so indexing completes successfully

Hope to find a cleaner solution in the future but because this is an internal tool rarely used I don't want to spend more time on it right now


joeldodge79 (approved):
I'd highly suggest changing the pr title (and making sure the actual squash commit takes it) to `fix: declaration miner something something` - this seems like an actual bug fix and that way it will go in changelogs and bump the version. otherwise it won't as a `chore:`

PR #1046, opened 2022-04-05
fix(typescript-sdk): stringify hashes in query parameters

jackscott6174:
Because javascript's default `toString()` turns hashes into `[object Object]` weirdness, we need to explicitly `JSON.stringify` any hash-type objects that are passed into the query string. Right now, this only applies to the `filters` parameter in the `model_fieldname_suggestions` endpoint, but this code should ensure that any future object-type query parameter will be properly stringified as well.

### DIFF
--- a/packages/sdk-codegen/src/typescript.gen.spec.ts
+++ b/packages/sdk-codegen/src/typescript.gen.spec.ts
@@ -545,6 +545,13 @@ describe('typescript generator', () => {
       const args = gen.httpArgs('', method).trim()
       expect(args).toEqual('null, body, options')
     })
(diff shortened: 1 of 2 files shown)


jackscott6174:
@jkaster First test failure looks like a timeout – how do I retest?

jkaster:
The workflow can be restarted. However, I think this change can be addressed in the sdk-rtl package rather than a private override in the code generator.  Did you look at `encodeParam()` in `transport.ts` to see if the change could be made there? If not, the change should be made in `typeMap()` in `typescript.gen.ts`.

PR #1047, opened 2022-04-06
fix: encodeParam returned [object Object] for object types

jkaster:
`@looker/sdk-rtl` was encoding object values like

```ts
{ created_date: 'this year to second' }
```

as 
```ts
[object Object]
```

and now the value is JSON stringified instead

### DIFF
--- a/packages/sdk-rtl/src/transport.ts
+++ b/packages/sdk-rtl/src/transport.ts
@@ -380,7 +380,9 @@ export function encodeParam(value: any) {
   if (value instanceof Date) {
     value = value.toISOString()
   }
-  let encoded = value.toString()
+  // check for object type to prevent "[object Object]" as the value.toString()
+  let encoded =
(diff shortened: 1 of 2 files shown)


joeldodge79 (approved):

packages/sdk-rtl/src/transport.ts
```diff
@@ -379,8 +380,12 @@ export type Values = { [key: string]: any } | null | undefined
 export function encodeParam(value: any) {
   if (value instanceof Date) {
     value = value.toISOString()
+  } else if (value instanceof DelimArray) {
```
bryans99, line 383: This looks so familiar. I thought we'd fixed it

bryans99 (commented):
LGTM

packages/sdk-rtl/src/transport.ts
jkaster: at least we've fixed it now

PR #1054, opened 2022-04-08
fix: Register Go Fuzzy JSON Decoders once on init

jeremytchang:
The RegisterFuzzyDecoders method can lead to a concurrent map write error if called by multiple threads.
Also it does not need to be called every time JSON needs to be decoded. It only needs to be called once on package init.
Change still passes the fuzzy decode unit tests. No additional tests were needed or needed to be updated.


go/rtl/auth.go
```diff
@@ -15,6 +15,12 @@ import (
 	extra "github.com/json-iterator/go/extra"
 )
 
+func init() {
```
bryans99, line 18: Is there an example or documentation of how this is to be called? 

Seems a little magic but I haven't pulled the code to look at the details of what is going on.

jeremytchang: It's called once on the package being imported. 
https://medium.com/golangspec/init-functions-in-go-eac191b3860a
https://www.digitalocean.com/community/tutorials/understanding-init-in-go

bryans99: TIL about init in go!!!!!

Thanks for the info.

bryans99 (approved):

PR #1055, opened 2022-04-08
fix: package refs for API Explorer extension

jkaster:
Needed up update package refs for API Explorer extension


bryans99 (approved):
LGTM

PR #1058, opened 2022-04-12
chore: upgrade eslint to fix header copyright year

bryans99:
Upgraded @looker/eslint-config-oss to version 1.7.14. Attempted to upgrade
to latest which is 1.7.15 but it wanted me to add an add package @looker/eslint-config
but that package is marked as no longer supported. Decided to leave that investigation
for another day.

Automatic fix fixed most of the errors which I assume is the header. I have not looked at
every file.

There were a number of no-nested-ternary errors for which I added eslint ignores. To much
room for error to fix. A couple were embedded in really large JSX snippets so had to ignore
all of the JSX.

The was one error to do with lodash. Nice easy fix on that one.

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "@babel/preset-typescript": "^7.13.0",
     "@babel/runtime": "^7.14.0",
     "@babel/runtime-corejs3": "^7.14.0",
-    "@looker/eslint-config-oss": "^1.7.1",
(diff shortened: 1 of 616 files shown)


bryans99:
Closed in favor of https://github.com/looker-open-source/sdk-codegen/pull/1060

PR #1060, opened 2022-04-12
chore: upgrade @looker/eslint-config-oss

bryans99:
1. pinned to version 1.7.14 as 1.7.15 requires unsupported package.
2. overrode header rule to use a pattern for date.
3. fixed one lodash issue that arose.
4. ignored nested ternary issues. should be fixed but outside the
   scope of what I want to do with this change.

The next published version of @looker/eslint-config-oss will include
the overridden header rule so the overridden header rule can be
removed then. Will investigate the unsupported package then.

### DIFF
--- a/packages/run-it/src/components/ConfigForm/ConfigForm.tsx
+++ b/packages/run-it/src/components/ConfigForm/ConfigForm.tsx
@@ -259,6 +259,7 @@ export const ConfigForm: FC<ConfigFormProps> = ({
     await adaptor.login()
   }
 
+  /* eslint-disable */
   return (
     <SpaceVertical gap="u2">
       <RunItHeading>{title}</RunItHeading>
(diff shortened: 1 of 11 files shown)


jkaster (approved):
LGTM

packages/run-it/src/components/ConfigForm/ConfigForm.tsx
```diff
@@ -259,6 +259,7 @@ export const ConfigForm: FC<ConfigFormProps> = ({
     await adaptor.login()
   }
 
+  /* eslint-disable */
```
whscullin, line 262: Can make this specifically `no-nested-ternary`, so that other errors can be caught, or may, since nested ternaries seem to be popular around here, just allow it via a global override?

bryans99: > since nested ternaries seem to be popular around here, just allow it via a global override?

I like that. I will spin up another PR later on today.

PR #1069, opened 2022-04-21
fix: Refactored Go SDK to use Go oauth2 client, added `x-looker-appid` header to Go requests

### LINKED ISSUE #1027: Support `x-looker-appid` for Go SDK identification like the other SDKs have
Looks like the transport layer for Go doesn't set `x-looker-appid` which we use for tracking SDK usage internally to help prioritize the SDK roadmap.

Let's get this added.

jeremytchang:
Fixes #1027 

- Refactored to use Go's [standard oauth2 client](https://pkg.go.dev/golang.org/x/oauth2) to handle our [client credentials flow](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4). The client will automatically fetch and refresh the token when it is expired. Removed all the manual authorization implementation.
- Also updated the the AuthSession to maintain only one client instead of creating a new one per request. As recommended by Go docs. `Clients and Transports are safe for concurrent use by multiple goroutines and for efficiency should only be created once and re-used.` [client docs](https://pkg.go.dev/net/http)
- Added `x-looker-appid` header with standard go http middleware, the RoundTripper interface. [Article](https://lanre.wtf/blog/2017/07/24/roundtripper-go/) because the docs are sparse on this one. 

- Removed auth implementation specific tests. Added couple auth sanity tests even though the go client itself is well tested. Updated tests to properly call NewAuthSession to properly create the client.


jkaster (approved):
LGTM. Nice cleanup.

PR #1072, opened 2022-05-03
feat: implement receive tile host message

bryans99:
The major change here is that it became apparent that the ExtensionConnector
did not have an up to date copy of the provider data. This was okay because all of
the provider components merged the data received from the ExtensionConnector
with the existing data. As the ExtensionConnector was only passing partial
data this was okay. For extensions as tile however, the ExtensionConnector
needed to update other provider and was failing because it did not have the
latest data. This was due to the callbacks being called from the chatty api
and being unaware changes made to properties in the react render cycle. To
solve this provider data is now stored in a reference so the latest data
is always available. In addition, provider data updates are now explicitly
called out as partial updates.

The main aim (although in the end a smaller change) is to make the extension
aware of changes made by the host to the tile. This involved some simplification
of states (DashboardRunState now has UNKNOWN, RUNNING, NOT_RUNNING), renaming
to be more dashboard specific (isDashboardEditing, isDashboardCrossFilteringEnabled,
dashboardFilters) and addition of new fields (dashboardId and isExploring).

isExploring indicates that the extension is running inside of an explore. In this
case the dashboard fields are not populated. It is possible that isExploring
may be moved to its own message in a future change.

Note that this branch is branched off of feature/extension-dashboard-tile and
will subsequently be merged into that branch

### DIFF
--- a/packages/extension-sdk-react/src/components/ExtensionConnector/ExtensionConnector.spec.tsx
+++ b/packages/extension-sdk-react/src/components/ExtensionConnector/ExtensionConnector.spec.tsx
@@ -198,16 +198,16 @@ describe('ExtensionConnector component', () => {
     expect(await screen.findByText('Mock Extension')).toBeInTheDocument()
     expect(connectedCallback).toHaveBeenCalled()
     tileHostDataChangedCb!({
-      isEditing: false,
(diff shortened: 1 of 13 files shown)


jkaster (approved):
LGTM

PR #1073, opened 2022-05-04
test: Diff scene e2e tests

fabio-looker:
Adds the following requested e2e tests:

- Diff Scene:
  - loads the default scene (/diff/3.1)
    -  "Base" input element (defaulting to 3.1)
    -  "Comparison" input element
    -  Switch button (disabled)
  - loads a comparison scene (/diff/3.1/4.0) and navigates from it
    - "Base" input element (set to 3.1)
    - "Comparison" input element (set to 4.0)
    - Switch button (enabled)
    - Diff results (present and containing an expected value)
    - Expanding a result
    - Navigating from a method result and validating the destination
  - updates when a comparison is chosen or switched
- Default/methods scene
  - Navigate to an APIX endpoint with examples (e.g. all_dashboards), click on one of the examples, which should open GitHub and show code that includes the operation ID (all_dashboards)

Additoinally, three previous e2e tests were failing (partially due to out of date tests from before 4.0 GA release). Two of them have been fixed, and one which was proving cumbersome has been commented out for the time being.


packages/api-explorer/e2e/e2e.spec.ts
```diff
@@ -164,31 +166,35 @@ describe('API Explorer', () => {
...
-        text: 'Looker API 4.0 (Beta) Reference',
-      })
-    })
+    // This test was broken during the 4.0 GA spec changes, and needs to be fixed
+    // it('changes specs', async () => {
+    //   await expect(page).toMatchElement('h2', {
+    //     text: 'Looker API 4.0 Reference',
+    //   })
+    //   await expect(page).toClick('input[value="4.0"]')
+    //   await Promise.all([
```
josephaxisa, line 175: is this intentionally commented out? if yes, do we have a ticket to track its fix?

josephaxisa (commented):
lgtm, just have one question before approving

packages/api-explorer/e2e/e2e.spec.ts
fabio-looker: Yes, it was broken when I merged main back in, along with two others that I was able to fix.

I just created this issue: https://b.corp.google.com/issues/231742409

josephaxisa (approved):

PR #1076, opened 2022-05-06
fix: remove generated files from examples index

jkaster:
Added more `build` folders to ignore for the code miners. API Explorer was showing links to files that were build artifacts and don't get checked in to the repository, so they were dead links on the developer portal.

### DIFF
--- a/packages/sdk-codegen-scripts/src/declarationMiner.ts
+++ b/packages/sdk-codegen-scripts/src/declarationMiner.ts
@@ -32,7 +32,7 @@ import {
   ExampleMiner,
   readFile,
   getCommitHash,
-  getRemoteHttpOrigin,
+  getRemoteOrigin,
 } from './exampleMiner'
 
 /**
@@ -94,7 +94,8 @@ export class DeclarationMiner {
   constructor(
     public readonly sourcePath: string,
     public readonly methodProbe: IProbe,
-    public readonly typeProbe: IProbe
+    public readonly typeProbe: IProbe,
+    private readonly originOverride: string
   ) {}
 
   execute(sourcePath?: string, methodProbe?: IProbe, typeProbe?: IProbe) {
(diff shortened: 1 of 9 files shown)


packages/sdk-codegen-scripts/src/declarationMiner.ts
```diff
@@ -127,8 +128,11 @@ export class DeclarationMiner {
   }
 
   get remoteOrigin(): string {
+    if (this.originOverride) {
+      return this.originOverride
+    }
     process.chdir(this.sourcePath)
-    return getRemoteHttpOrigin()
+    return getRemoteOrigin()
```
bryans99, line 135: Is this what we really want. I looked at the code and it executes
`git remote get-url origin`
which returns
`git@github.com:looker-open-source/sdk-codegen.git`
Looke like getRemoteHttpOrigin attemps to convert it

packages/sdk-codegen-scripts/src/exampleMiner.ts
```diff
@@ -80,7 +80,17 @@ export const filterCodeFiles = (fileName: string) => {
...
 
-const skipFolder = (name: string, excludeList: string[] = ['node_modules']) =>
+export const IGNORE_PATHS = [
+  'node_modules',
+  'lib',
+  'dist',
+  'bazel-bin',
+  'build',
+  'bin',
+  '.build',
```
bryans99, line 90: .direnv, .github, .vscode, .idea, results?

Probably overkill

bryans99 (approved):
LGTM with a couple of questions

packages/sdk-codegen-scripts/src/declarationMiner.ts
jkaster: Yes. Note this is declaration miner, not example miner. The address for the helltool code repo is no longer https when you ask for the remote origin. That's why I added the `origin_override` configuration and alternate path composition.

packages/sdk-codegen-scripts/src/exampleMiner.ts
jkaster: Easy enough to add. I'll do that now.

jkaster: Added

bryans99 (approved):

PR #1084, opened 2022-05-25
feat: additional tile host data

bryans99:
Dashboard run start/end times now provided along with success information.
Allows extension to determin if the dashboard run button was pressed (or
auto run).

Example of usage provided.


jkaster (approved):
LGTM

PR #1085, opened 2022-05-26
feat: extension tile information

bryans99:
Provides additional information about the dashboard tile.
1. element id.
2. lastRunSourceElementId - populated when dashboard run is
   initiated by an extension tile.

Also, minor fix to documentation.
Also, minor adjustment to tests.


josephaxisa (approved):
lgtm

PR #1086, opened 2022-05-26
fix: description sync issues in diff scene

josephaxisa:
also fixes a bug where we weren't rendering the first delta

Before:
before

After
after


jkaster (approved):
LGTM

PR #1087, opened 2022-05-31
feat: extension tile update filter

bryans99:
1. extension sdk support for update filter
2. extension tile example update filter example


jkaster (approved):
LGTM

PR #1098, opened 2022-06-21
chore: separate unit and integration test configs

josephaxisa:
prior to this change, debugging a unit test from inside the IDE was not possible.

To verify the fix:
1) run `yarn test:apix` from the root of the monorepo
2) run `yarn test:apix-e2e` from the root of the monorepo
3) try debugging a test


jkaster (approved):
Works for me. 2 failing e2e tests can be fixed separately

PR #1115, opened 2022-07-13
feat: generate SDKs for Looker 22.10

jkaster:


bin/sdk_gen
```diff
@@ -172,16 +172,15 @@ const pullci = async (release) => {
...
-        // 'git add -A',
-        // `git commit -m "feat: generate SDKs for Looker ${release}"`,
-        /*
-        Test these out after first cut is merged to main
-        'git push'
-        create PR
+        'git add -A',
+        `git commit -m "feat: generate SDKs for Looker ${release}"`,
+        `git push origin ${branch}`,
+        /* create PR
```
josephaxisa, line 183: I believe we mentioned this will be removed in a separate PR so I won't let it block this.

josephaxisa (approved):
lgtm - thanks for your work on this John!

bin/sdk_gen
jkaster: yeah, I'll get rid of the commented out PR notes for the next round. Thanks!

PR #1120, opened 2022-07-19
feat: copy link button for search

patnir41:
Upon typing in the search box and hovering the mouse on the input box itself, a link button will appear for the user to copy the current URL and share their page view with others. When the user hovers away from the search box, the link button will disappear.


packages/api-explorer/src/components/SideNav/SideNav.tsx
```diff
@@ -160,18 +165,33 @@ export const SideNav: FC<SideNavProps> = ({ headless = false, spec }) => {
...
+            isClearable
+          />
+          <CopyLinkButton
+            top={'24px'}
+            right={'58px'}
+            visible={showCopyLinkButton}
+          />
+        </Span>
+        <SearchMessage search={searchResults} />
+      </Box2>
```
josephaxisa, line 194: sidenav unit tests are missing

packages/api-explorer/src/components/CopyLinkButton/CopyLinkButton.tsx
```diff
@@ -0,0 +1,68 @@
...
+  )
+}
+
+const CopyLink = styled('span')<{
+  top: string
+  right: string
+  visible: boolean
+}>`
+  position: absolute;
+  top: ${({ top }) => top};
```
josephaxisa, line 65: would margins work here? 

in general, we should avoid custom spacing property values and leverage @looker/components built-in spacing values. See this [example](https://github.com/looker-open-source/sdk-codegen/blob/main/packages/api-explorer/src/components/SideNav/SideNavMethodTags.tsx#L54) on how to reference `theme` and available spacing props.

Also, I could see a use for having margin top/bottom/left/right. Consider supporting that now.

In the end, you should be able to use this component in ways similar to below:
```ts
<CopyLinkButton visible={true} mt="small" ml="large" />
```

josephaxisa (changes requested):
looking good. Some feedback below

packages/api-explorer/src/components/CopyLinkButton/CopyLinkButton.tsx
patnir41: Currently blocked here, facing a problem where the copy button appears aligned to the far-left side and unable to be right-aligned.

jkaster (commented):
Minor comment

packages/run-it/src/components/CopyLinkWrapper/CopyLinkWrapper.tsx
```diff
@@ -0,0 +1,78 @@
...
+ * @param visible
+ */
+export const CopyLinkWrapper = ({
+  children,
+  visible = true,
+}: CopyLinkWrapperProps) => {
+  const [tooltipContent, setTooltipContent] = useState(COPY_TO_CLIPBOARD)
+  const [showCopyLinkButton, setShowCopyLinkButton] = useState(false)
+  const handleCopyLink = async () => {
+    // const adaptor = getEnvAdaptor()
```
jkaster, line 52: code cruft?

jkaster (commented):
Haven't tested this yet, so just a comment about a commented out code line

packages/run-it/src/components/CopyLinkWrapper/CopyLinkWrapper.tsx
```diff
@@ -0,0 +1,78 @@
...
+interface CopyLinkWrapperProps {
+  children: ReactNode | ReactNodeArray
+  visible?: boolean
+}
+
+const COPY_TO_CLIPBOARD = 'Copy to clipboard'
+
+/**
+ * Wrapper component that can render a hideable copy link button next to any
+ * passed in children components
```
josephaxisa, line 40:
```suggestion
 * Displays a copy link button on hover
 *
```

packages/run-it/src/components/CopyLinkWrapper/CopyLinkWrapper.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+ */
+import type { ReactNode, ReactNodeArray } from 'react'
+import React, { useState } from 'react'
+import { IconButton, Space } from '@looker/components'
+import { Link } from '@styled-icons/material-outlined/Link'
+
+interface CopyLinkWrapperProps {
+  children: ReactNode | ReactNodeArray
+  visible?: boolean
```
josephaxisa, line 33: documentation?

jkaster:
FWIW this isn't working with the extension framework. Let me know if you need help setting up the development version of the extension with your local Looker instance.

packages/run-it/src/components/CopyLinkWrapper/CopyLinkWrapper.tsx
```diff
@@ -0,0 +1,77 @@
...
+ */
+export const CopyLinkWrapper = ({
+  children,
+  visible = true,
+}: CopyLinkWrapperProps) => {
+  const [tooltipContent, setTooltipContent] = useState(COPY_TO_CLIPBOARD)
+  const [showCopyLinkButton, setShowCopyLinkButton] = useState(false)
+  const handleCopyLink = async () => {
+    // const adaptor = getEnvAdaptor()
+    await navigator.clipboard.writeText(location.href)
```
patnir41, line 52: TODO: set this to support browser vs extensions framework

packages/extension-utils/src/extensionAdaptor.ts
```diff
@@ -59,6 +59,12 @@ export class ExtensionAdaptor
     )
   }
 
+  async copyToClipboard(location: { pathname: string; search: string }) {
```
patnir41, line 62: Commenting to make visible to @josephaxisa @bryans99 : Per my discussion with Bryn today, I need a way to access the browser window.location object in this adaptor in order to append its pathname/search params to the hostname/url contained by the extensionSdk. I do this by passing in the `location` returned from `useLocation` when called in CopyLinkWrapper.

I experimented with a couple of different approaches for passing in the parameter, but settled on this typed object. This made it easier than dealing with the location object overall and its type issues when passing it in through the spec file as opposed to after getting it from the `useLocation` hook in CopyLinkWrapper. Thoughts on this approach?

bryans99: Change to this. Route is available in lookerHostData. No need to pass it in
```

  async copyToClipboard() {
    const routeContext = this.extensionSdk.lookerHostData
    const { lookerHostData } = this.extensionSdk
    if (lookerHostData) {
      const { extensionId, route } = lookerHostData
      const url = `${routeContext?.hostUrl}/extensions/${extensionId}${route}`
      await this.extensionSdk.clipboardWrite(url)
    }
  }
```

Update `packages/extension-utils/src/adaptorUtils.ts` as follows

```
export interface IEnvironmentAdaptor extends IAuthAdaptor {
  /** Method for determining whether running in a browser or extension environment */
  isExtension(): boolean
  /** Method for retrieving a keyed value from local storage */
  localStorageGetItem(key: string): Promise<string | null>
  /** Method for setting a keyed value in local storage */
  localStorageSetItem(key: string, value: string): void
  /** Method for removing a keyed value from local storage */
  localStorageRemoveItem(key: string): void
  /** Theme settings for extension */
  themeOverrides(): ThemeOverrides
  /** Open a new browser window with the given url and target  */
  openBrowserWindow: (url: string, target?: string) => void
  /** error logger */
  logError: (error: Error, componentStack: string) => void
  /** copy to clipboard */
  copyToClipboard: () => Promise<void>
}
```

And CopyLinkWrapper.tsx as follows
```
  const handleCopyLink = async () => {
    await getEnvAdaptor().copyToClipboard()
    // await navigator.clipboard.writeText(location.href)
    setTooltipContent('Copied to clipboard')
  }
```

bryans99 (changes requested):
See comments

bryans99 (commented):
I've suggested cleaning up the extension copy method. 
1. handles possibility that fields are undefined (should not happen but typing says it could)
2. uses non deprecated field name
3. does not implicitly rename fields

bryans99 (commented):
LGTM. 

Will let Jax give the final stamp

jkaster (commented):
Looks like it's ready to test. I'll do that before EOD.

josephaxisa (approved):
lgtm. Thanks Niral!

jkaster (approved):
Works for me!

bryans99 (approved):

PR #1123, opened 2022-07-22
feat: SDK language selection as URL search parameter

patnir41:
Upon having a selected SDK language in the API Explorer, an 'sdk' search parameter will be present on the URL that will allow developers to share method and type descriptions with the appropriate language automatically selected.


packages/api-explorer/src/components/SelectorContainer/SdkLanguageSelector.spec.tsx
```diff
@@ -82,4 +97,20 @@ describe('SdkLanguageSelector', () => {
       })
     })
   })
+  test('changing SDK language causes parameter to be pushed to URL', async () => {
+    renderWithRouterAndReduxProvider(<SdkLanguageSelector />)
```
patnir41, line 101: discuss with Joseph: whitespace between the 3 main steps, render, getting components, interacting with components.

Be generous with it between different tests.

packages/api-explorer/src/utils/hooks.spec.ts
```diff
@@ -60,19 +60,14 @@ describe('Navigate', () => {
     })
   })
 
-  test('clears existing params when given params are an empty object', () => {
```
patnir41, line 63: Wanted to discuss if we should still account for the case of an empty object. We currently use `undefined` as a value to indicate we are keeping the key that already exists in the URL, and `null` to indicate that we are getting rid of parameters. If you look at the way I changed the code, individual parameters can be removed, as well as all of them, if passing in `null` entirely. 

Let me know thoughts or we can discuss in call.

packages/api-explorer/src/components/SelectorContainer/SdkLanguageSelector.tsx
```diff
@@ -55,6 +57,20 @@ export const SdkLanguageSelector: FC = () => {
     setSdkLanguageAction({ sdkLanguage: language })
   }
 
+  useEffect(() => {
+    const searchParams = new URLSearchParams(history.location.search)
+    const urlSdk = searchParams.get('sdk')
+    if (urlSdk && allSdkLanguages.find((lang) => lang.label === urlSdk)) {
+      setSdkLanguageAction({ sdkLanguage: urlSdk! })
+    }
+  }, [])
```
josephaxisa, line 66: if `APIExplorer.tsx` is handling reading query params and driving state accordingly shouldn't this entire effect be redundant?

packages/api-explorer/src/components/SelectorContainer/SdkLanguageSelector.tsx
```diff
@@ -55,6 +57,20 @@ export const SdkLanguageSelector: FC = () => {
...
+
+  useEffect(() => {
+    navigate(history.location.pathname, {
+      sdk: selectedSdkLanguage.toLowerCase(),
+    })
```
josephaxisa, line 71: based on the implementation we discussed for navigate I would expect this call to be along the lines of 
```
navigate(history.location.pathname, {...existingParams, sdk: selectSdkLanguage.toLowerCase()}
```

packages/api-explorer/src/components/SelectorContainer/SdkLanguageSelector.tsx
```diff
@@ -55,6 +57,20 @@ export const SdkLanguageSelector: FC = () => {
...
+  }, [selectedSdkLanguage])
```
josephaxisa, line 72: what about when "All" is selected? This should default to no sdk query param.

packages/api-explorer/src/utils/hooks.spec.ts
josephaxisa: I'm ok with this. @jkaster @bryans99 thoughts?

I don't see tests unit that cover this though.
1. Unit test to show the additiveness. i.e if current query params are `foo=foo` and I pass in `bar=bar` then I end up with both.
2. Unit test to show that if current params are `foo=foo` and `bar=bar`, passing in `bar=null` should leave me with `foo=foo`.
3. Unit test to show that passing in null clears all existing query params

josephaxisa (changes requested):

packages/api-explorer/src/components/SelectorContainer/SdkLanguageSelector.tsx
patnir41: If we just pass in `sdk: sdk: selectSdkLanguage.toLowerCase()`, the remaining params (or the `existingParams` which you note) would be undefined, so they would remain in the URL. It would be redundant to have existing params then wouldn't it?

josephaxisa: agreed. I wrote this before I saw and replied to your comment [here](https://github.com/looker-open-source/sdk-codegen/pull/1123#discussion_r930358895)

packages/api-explorer/src/components/SelectorContainer/SdkLanguageSelector.tsx
patnir41: Agree, I had a lapse of judgement here.

packages/api-explorer/src/components/SideNav/SideNav.tsx
```diff
@@ -113,19 +113,20 @@ export const SideNav: FC<SideNavProps> = ({ headless = false, spec }) => {
...
-  }, [location.search, debouncedPattern])
+  }, [debouncedPattern])
 
   useEffect(() => {
     let results
     let newTags
     let newTypes
     let newTypeTags
     const api = spec.api || ({} as ApiModel)
+    setSearchPattern(searchPattern)
```
patnir41, line 129: I have to remove `location.search` as a dependency from the useEffect and  setSearchPattern(searchPattern) in the searchPattern listener for routing backwards/forwards. Imagine there you have it blank, then search ‘test’ in the box. If you press the back button, APIExplorer will have its listener trigger and set our search pattern to ‘’.  *The problem, however, if `location.search` is a dependency of the useEffect based on `debouncedPattern`, the pattern will not have debounced immediately and the `debouncedPattern` will still point to ‘test’. Hence the error.

Why we need `setSearchPattern(searchPattern`: once APIExplorer executes `setSearchPatternAction`, this useEffect will only update the filters that will be visible in the sideNav. We need the search bar to also reflect the new search change, if that makes sense.

packages/api-explorer/src/components/SideNav/SideNav.tsx
```diff
@@ -113,19 +113,20 @@ export const SideNav: FC<SideNavProps> = ({ headless = false, spec }) => {
...
       searchParams.set('s', debouncedPattern)
-      navigate(location.pathname, { search: searchParams.toString() })
+      navigate(location.pathname, { s: searchParams.get('s') })
     } else if (!debouncedPattern && searchParams.get('s')) {
       searchParams.delete('s')
-      navigate(location.pathname, { search: searchParams.toString() })
+      navigate(location.pathname, { s: null })
     }
-  }, [location.search, debouncedPattern])
+  }, [debouncedPattern])
```
patnir41, line 121: @jkaster 
I have to remove `location.search` as a dependency from this useEffect due to routing backwards/forwards. Imagine there you load APIExplorer, leave it blank, then search ‘test’ in the box. After it searches, if you press the back button, APIExplorer will have its listener trigger and set our search pattern to ‘’.  *The problem, however, is if `location.search` is a dependency of this useEffect.  The pattern (of value '') will not have debounced immediately and the `debouncedPattern` will refer to ‘test’. Hence the URL will not remove the 's=' parameter as it should

jkaster: I think that may have been feedback from @bryans99 instead of me? If it was me, sorry I don't remember!

patnir41: Totally correct, my apologies ! This is for Bryn.

bryans99: If its causing a problem with it and not without then lets remove it. Please add a comment as to why its not included. At some point I will look at it more closely

packages/api-explorer/src/utils/hooks.spec.ts
patnir41: @josephaxisa See new tests, also I thought the test starting in line 56 (clears existing params when given params are null) addressed your 3rd point.

packages/api-explorer/src/components/SelectorContainer/SdkLanguageSelector.spec.tsx
```diff
@@ -60,25 +75,19 @@ describe('SdkLanguageSelector', () => {
     })
   })
 
-  test('it stores the selected language in localStorage', async () => {
```
patnir41, line 63: Deleted this test, leaving for integration testing which will be implemented within this month. Cannot have this test within the unit test since it relies on an action being dispatched in APIExplorer.tsx.

packages/api-explorer/src/components/SelectorContainer/SdkLanguageSelector.tsx
```diff
@@ -52,9 +53,25 @@ export const SdkLanguageSelector: FC = () => {
...
 
   const handleChange = (language: string) => {
-    setSdkLanguageAction({ sdkLanguage: language })
+    navigate(location.pathname, { sdk: language.toLowerCase() })
   }
 
+  useEffect(() => {
+    const searchParams = new URLSearchParams(location.search)
+    const sdkParam =
+      selectedSdkLanguage === 'All' ? null : selectedSdkLanguage.toLowerCase()
```
patnir41, line 62: NOTE: Currently working on a bug with SDK choosing 'All'. Working on ensuring the sdk parameter is removed when choosing All, and that this works when going back or forward in the browser history. Will update when fixed and added a test for choosing SDK as "All".

packages/api-explorer/src/ApiExplorer.tsx
```diff
@@ -116,6 +127,26 @@ export const ApiExplorer: FC<ApiExplorerProps> = ({
     return () => unregisterEnvAdaptor()
   }, [])
 
+  useEffect(() => {
+    // reconcile local storage state with URL or vice versa
```
josephaxisa, line 131: as we discussed, in the future this might not be an either or scenario. Adding this comment for other reviewers

packages/api-explorer/src/state/settings/slice.ts
```diff
@@ -43,7 +43,7 @@ export interface SettingState extends UserDefinedSettings {
 }
 
 export const defaultSettings = {
-  sdkLanguage: 'Python',
+  sdkLanguage: 'py',
```
josephaxisa, line 46: it's worth renaming this key to `sdkLanguageAlias` and its associated selector.

packages/api-explorer/src/test-data/sdkLanguages.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+export const sdkLanguageMapping = {
+  py: 'Python',
+  ts: 'TypeScript',
+  cs: 'C#',
+  kt: 'Kotlin',
+  swift: 'Swift',
+  go: 'Go',
+  all: 'All',
+}
```
josephaxisa, line 34: note that this was purposely hardcoded for testing purposes. Supporting a new language in API Explorer requires a minor manual effort since we have dependencies such as `prism` where a human needs to confirm correctness.

packages/api-explorer/src/utils/hooks.spec.ts
```diff
@@ -60,19 +60,32 @@ describe('Navigate', () => {
...
-    navigate(route, {})
+  test('removes particular parameter when object is passed in with its value as null', () => {
```
josephaxisa, line 63:
```suggestion
  test('null query params are removed, () => {
```

packages/api-explorer/src/ApiExplorer.tsx
```diff
@@ -116,6 +127,26 @@ export const ApiExplorer: FC<ApiExplorerProps> = ({
...
+      const sdkLanguage = searchParams.get('sdk')
+      if (sdkLanguage && getLanguageByAlias(sdkLanguage)) {
+        // sync store with URL
+        setSdkLanguageAction({
+          sdkLanguage: sdkLanguage.toLowerCase(),
+        })
+      } else {
+        // sync URL with store
+        const sdk =
+          getAliasByLanguage(selectedSdkLanguage) || selectedSdkLanguage
```
josephaxisa, line 142: note that this was needed because we changed what's stored in `sdkLanguage` to be the alias. As a result, existing customers with the full language name in `sdkLanguage` would run into issues if we didn't account for this. If we change `sdkLanguage` to `sdkLanguageAlias` this will no longer be an issue.

packages/api-explorer/src/ApiExplorer.tsx
```diff
@@ -116,6 +127,26 @@ export const ApiExplorer: FC<ApiExplorerProps> = ({
     return () => unregisterEnvAdaptor()
   }, [])
 
+  useEffect(() => {
+    // reconcile local storage state with URL or vice versa
+    if (initialized) {
+      const sdkLanguage = searchParams.get('sdk')
+      if (sdkLanguage && getLanguageByAlias(sdkLanguage)) {
```
josephaxisa, line 134: This check means that if a user passes in `sdk=invalid` we are defaulting to what they previously had in the store. Alternatively we could just push them to `all`. @jkaster  any preference?

josephaxisa (commented):
this is looking great. some comments below. Also it seems like the `SDK Call` tab in `RunIt` is not rerendering properly on language change.

packages/api-explorer/e2e/e2e.spec.ts
```diff
@@ -255,7 +255,9 @@ describe('API Explorer', () => {
       await expect(page).toMatchElement('button', { text: 'Types (0)' })
       await expect(page).toClick('a', { text: 'Get Workspace' })
       await expect(page).toMatchElement('h2', { text: 'Get Workspace' })
-      await expect(page.url()).toEqual(`${v40}/methods/Workspace/workspace`)
+      await expect(page.url()).toEqual(
+        `${v40}/methods/Workspace/workspace?sdk=kt&s=get+workspace`
```
josephaxisa, line 259: seems like we have some state being carried over between tests. Let's clear local storage between tests using the `beforeEach` api.

packages/api-explorer/e2e/e2e.spec.ts
```diff
@@ -36,6 +36,7 @@ const v40 = `${BASE_URL}/4.0`
 
 describe('API Explorer', () => {
   beforeEach(async () => {
+    localStorage.clear()
```
patnir41, line 39: This shouldn't be here, WIP as to how to clear state for tests without causing failures.

packages/api-explorer/src/ApiExplorer.tsx
patnir41: Noting the new rename here, we should be able to remove this check now?

packages/api-explorer/src/utils/sdkLanguage.ts
```diff
@@ -0,0 +1,60 @@
...
+ * Searches for alias by sdk language name
+ * @param language name to search by
+ * @returns alias
+ */
+export const getAliasByLanguage = (language: string) => {
+  const languages = allSdkLanguages()
+  const match = Object.keys(languages).find(
+    (alias) => languages[alias].toLowerCase() === language.toLowerCase()
+  )
+  return match
```
jkaster, line 59: This function should be `getSDKAlias` and support either language or alias as the value to match. Also, this shouldn't be doing a table scan by default. `allSDKLanguages()` is keyed by alias, so a generic `findSDK() function that checks for languages[alias] first then checks for language matching with a `.find()` can be used. Also, the case conversions can be avoided by using [localeCompare()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare). And now that I'm thinking about it, we should fix [findGenerator()](https://github.com/looker-open-source/sdk-codegen/blob/9909206e326b60a4465dbd50fbcf0ddb71749ae7/packages/sdk-codegen/src/codeGenerators.ts#L129) for with localeCompare() the same way.

packages/api-explorer/src/test-data/sdkLanguages.ts
jkaster: This needs to be created from iterating through the existing language definitions in sdk-codegen before this PR can be merged. Let's not have a different hard-coded list of languages.

packages/api-explorer/src/components/DocSdkUsage/DocSdkUsage.tsx
```diff
@@ -61,7 +62,8 @@ interface DocSdkUsageProps {
  */
 export const DocSdkUsage: FC<DocSdkUsageProps> = ({ method }) => {
   const examples = useSelector(selectExamplesLode)
-  const sdkLanguage = useSelector(selectSdkLanguage)
+  const alias = useSelector(selectSdkLanguage)
```
jkaster: is this selector still `selectSdkLanguage`, or is it `selectSdkAlias` or just `selectSdk`?

packages/api-explorer/src/ApiExplorer.tsx
jkaster: I think an invalid language should clear any configured preference and reset to the default behavior

jkaster (changes requested):
Great progress. I'm looking forward to this landing.

jkaster (commented):
Some comments since there's pending feedback requested from others

packages/api-explorer/src/components/SelectorContainer/SdkLanguageSelector.spec.tsx
```diff
@@ -24,61 +24,85 @@
...
   })
 
-  test('it has the correct default language selected', () => {
-    renderWithReduxProvider(<SdkLanguageSelector />)
-    expect(screen.getByRole('textbox')).toHaveValue(
-      defaultSettingsState.sdkLanguage
-    )
+  test('it has the correct default language selected', async () => {
+    renderWithRouterAndReduxProvider(<SdkLanguageSelector />, undefined, store)
+    expect(screen.getByRole('textbox')).toHaveValue('Python')
```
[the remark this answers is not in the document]
patnir41: Test is now reverted back to previous state given new change to store full language name in localStorage and no longer deal with storing an alias as well.

packages/api-explorer/e2e/diffScene.spec.ts
```diff
@@ -151,7 +151,7 @@ describe('Diff Scene', () => {
           page.evaluate((el) => el.innerText.match(/^[a-z_]*/)[0], resultCard)
         )
       )
-      expect(page1Methods).toHaveLength(15)
+      expect(page1Methods).toHaveLength(16)
```
patnir41, line 154: This was failing in main, fixed so that it is up to date and accurately reflecting new additional method.

packages/api-explorer/src/ApiExplorer.tsx
patnir41: @josephaxisa Is there anything left to discuss right now here?

josephaxisa: I don't think so

packages/api-explorer/e2e/e2e.spec.ts
```diff
@@ -164,6 +164,9 @@ describe('API Explorer', () => {
         'Kotlin'
       )
       await expect(page).toMatchElement('h3', { text: 'Kotlin Declaration' })
+      await page.evaluate(() => {
```
patnir41, line 167: Observed this cause unpredictable behavior, new commit will move this to an `afterEach` within this test block.

josephaxisa: I can see the `afterEach` but this is still here. Is this intended?

packages/api-explorer/src/ApiExplorer.tsx
```diff
@@ -116,6 +121,29 @@ export const ApiExplorer: FC<ApiExplorerProps> = ({
...
 
+  useEffect(() => {
+    // reconcile local storage state with URL or vice versa
+    if (initialized) {
+      const sdkParam = searchParams.get('sdk') || ''
+      const sdk = findSdk(sdkParam)
+      const paramMatchesSdk =
+        !sdkParam.localeCompare(sdk.alias, 'en', { sensitivity: 'base' }) ||
+        !sdkParam.localeCompare(sdk.language, 'en', { sensitivity: 'base' })
+      if (sdkParam && sdk && paramMatchesSdk) {
```
josephaxisa, line 132: is `sdk` redundant in this check? Also perhaps `sdkParam && validSdkParam` would be more descriptive here.

packages/api-explorer/src/components/SelectorContainer/SdkLanguageSelector.spec.tsx
```diff
@@ -24,61 +24,88 @@
...
-import { defaultSettingsState, settingsSlice } from '../../state'
-import { renderWithReduxProvider } from '../../test-utils'
+import {
+  createTestStore,
+  renderWithRouterAndReduxProvider,
+} from '../../test-utils'
+import { findSdk } from '../../utils'
+import { languages } from '../../test-data'
 import { SdkLanguageSelector } from './SdkLanguageSelector'
+import { defaultSettingsState } from '@looker/api-explorer'
```
josephaxisa, line 37: import needs to be relative.

packages/api-explorer/src/utils/sdkLanguage.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+        const actual = findSdk(alias)
+        expect(actual?.language).toEqual(language)
+      }
+    )
+
+    test.each(Object.entries(sdkLanguageMapping))(
+      `it finds alias for language %s`,
+      (alias, language) => {
+        const actual = findSdk(language)
+        expect(actual?.alias).toEqual(alias)
```
josephaxisa, line 59: the safe navigation operator `?` shouldn't be needed anymore. 

also add a unit test to confirm that `all` is returned for a bogus/unsupported language.

packages/api-explorer/src/utils/sdkLanguage.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+  test('allSdkLanguages gets all sdk languages', () => {
+    const actual = allSdkLanguages()
+    const aliases = Object.keys(actual)
+    expect(Object.keys(actual)).toHaveLength(codeGenerators.length + 1)
+    aliases.forEach((alias) =>
+      expect(actual[alias]).toEqual(sdkLanguageMapping[alias])
+    )
+  })
+
+  describe('getSdkLanguage', () => {
```
josephaxisa, line 41:
```suggestion
  describe('findSdk', () => {
```

packages/api-explorer/src/components/SelectorContainer/utils.ts
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * @returns sdk language options for render
+ */
+export const allSdkLanguageOptions = (): SelectOptionProps[] => {
+  const options = [] as SelectOptionProps[]
+  const languages = allSdkLanguages()
+  Object.entries(languages).forEach(([alias, language]) => {
+    const option = { value: alias, label: language }
+    if (alias === 'all') {
```
josephaxisa, line 37: use `allAlias` from `utils/sdkLanguage.ts`?

packages/api-explorer/src/ApiExplorer.tsx
```diff
@@ -116,6 +121,29 @@ export const ApiExplorer: FC<ApiExplorerProps> = ({
...
+        // sync store with URL
+        setSdkLanguageAction({
+          sdkLanguage: sdk.language,
+        })
+      } else {
+        // sync URL with store
+        const { alias } = findSdk(selectedSdkLanguage)
+        navigate(location.pathname, {
+          sdk: alias === 'all' ? null : alias,
```
josephaxisa, line 141: use `allAlias` from `utils/sdkLanguage.ts`?

packages/api-explorer/src/components/SelectorContainer/SdkLanguageSelector.tsx
```diff
@@ -24,43 +24,36 @@
...
-      {
-        value: 'All',
-      },
-    ],
-  })
-
-  const handleChange = (language: string) => {
-    setSdkLanguageAction({ sdkLanguage: language })
+  const handleChange = (alias: string) => {
+    navigate(location.pathname, { sdk: alias === 'all' ? null : alias })
```
josephaxisa, line 44: use `allAlias` from `utils/sdkLanguage.ts`?

josephaxisa (changes requested):

packages/api-explorer/src/ApiExplorer.tsx
patnir41: Yeah, I think we can get by with just `validSdkParam` for the conditional check, as it will encapsulate all situations. If there is no sdk parameter, then sdkParam will be '' and the comparison will be false. If there is a parameter then we need to know if it matches what was found when we searched.

packages/api-explorer/e2e/e2e.spec.ts
patnir41: The most recent commit maintains the intended structure, there are two afterEach's that handle the clearing of local storage.

josephaxisa (approved):
lgtm, thanks Niral!

jkaster (approved):
LGTM

PR #1129, opened 2022-07-28
chore: reset release config CalVer

jkaster:
- also change TS SDK workflow to use node 14, 16, and 18

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -2,7 +2,6 @@
   "plugins": ["node-workspace"],
   "bump-minor-pre-major": true,
   "bump-patch-for-minor-pre-major": true,
-  "release-as": "",
   "packages": {
     ".": { "release-as": "" },
     "packages/api-explorer": { "release-as": "" },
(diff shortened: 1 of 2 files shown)


release-please-config.json
```diff
@@ -2,7 +2,7 @@
   "plugins": ["node-workspace"],
   "bump-minor-pre-major": true,
   "bump-patch-for-minor-pre-major": true,
-  "release-as": "22.10.0",
+  "release-as": "",
```
joeldodge79, line 5: I know it's bizarre, but an empty string actually means ["use semver"](https://docs.google.com/document/d/1-FBgwiL0j4KyNFV6yk7J_FKlF6QcWlYUouIyI9CblQw/edit?resourcekey=0-mpwoYOWpXZqCDF5SzqIckQ#bookmark=id.qkmui5mqtcd4) and because this is a top level config it will be the default for any child package not configured explicitly with `release-as`

The correct thing to do here is delete the key (though perhaps setting to `null` would have the same effect, not sure, we could find out by doing it an examining what the subsequent release PR looks like)

joeldodge79 (dismissed):

release-please-config.json
jkaster: Thanks, I was wondering about that. I've removed the key. Not going to bother testing `undefined` or `null` values for it

joeldodge79 (approved):
👍

PR #1134, opened 2022-08-01
fix: add field details for SDK error

praphun:
Looker API returns field-level information in the error. This change surfaces the information when using the SDK.


packages/sdk-rtl/src/transport.ts
```diff
@@ -470,32 +471,35 @@ function bufferString(val: any) {
...
     // Try to get most specific error first
     if ('error' in error) {
       const result = bufferString(error.error)
-      return new Error(result)
+      return new LookerSdkError(result)
     }
     if ('message' in error) {
-      return new Error(response.error.message.toString())
+      return new LookerSdkError(response.error.message.toString(), {
+        errors: error.errors,
```
jkaster, line 488: what happens if `error.errors` is not defined?

packages/sdk-rtl/src/lookerSdkError.ts
```diff
@@ -0,0 +1,179 @@
...
+  documentation_url: string | null
+}
+
+// This specifies SDK custom error options
+interface LookerSdkErrorOptions {
+  errors?: IErrorDetail[]
+  documentation_url?: string | null
+}
+
+export interface LookerSdkError extends Error {
```
jkaster, line 60: We've typically used `SDK` for names, like `LookerSDKError`. The interfaces in our SDK also have been prefixed with `I` but I'm not that concerned about it with low-level run-time code. However, the TS SDK methods expect an `IError` or `IValidationError` response currently, so I think we may need to change the code generator to support this interface. `IError` and `IValidationError` are used based on the API specification for the method, like this:

```ts
  async update_alert(
    alert_id: string,
    body: Partial<IWriteAlert>,
    options?: Partial<ITransportSettings>
  ): Promise<SDKResponse<IAlert, IError | IValidationError>> {
    alert_id = encodeParam(alert_id)
    return this.put<IAlert, IError | IValidationError>(
      `/alerts/${alert_id}`,
      null,
      body,
      options
    )
  }
```

jkaster (commented):
Excellent progress. Just feedback at this point. The `sdk-rtl` readme should be updated to describe the new error capturing behavior also, or a new examples/typescript showing how validation errors can be caught and explored.

I updated your PR comment to be a fix rather than a feature because we don't want the minor version to increment for the SDK.

packages/sdk-rtl/src/transport.ts
praphun: `errors` will be an empty list if `error.errors` is not defined.

packages/sdk-node/test/methods.spec.ts
```diff
@@ -1088,6 +1089,46 @@ describe('LookerNodeSDK', () => {
     )
   })
 
+  describe('Template', () => {
+    it('validate_template returns ok on valid template', async () => {
```
jkaster, line 1093:
```suggestion
    it('validate_theme returns ok on valid template', async () => {
```

packages/sdk-node/test/methods.spec.ts
```diff
@@ -1088,6 +1089,46 @@ describe('LookerNodeSDK', () => {
...
+            tile_shadow: false,
+            font_family: 'Arial',
+          },
+        })
+      )
+      expect(result).toBeDefined()
+      expect(result).toEqual('')
+    })
+
+    it('validate_template throws error with details', async () => {
```
jkaster, line 1110:
```suggestion
    it('validate_theme throws error with details', async () => {
```

packages/sdk-node/test/methods.spec.ts
```diff
@@ -1088,6 +1089,46 @@ describe('LookerNodeSDK', () => {
...
+              tile_shadow: false,
+              font_family: 'Arial;',
+            },
+          })
+        )
+      } catch (e: any) {
+        expect(e).toBeInstanceOf(LookerSdkError)
+        expect(e.message).toBeDefined()
+        expect(e.errors).toBeDefined()
+        expect(e.errors).toHaveLength(3)
```
jkaster, line 1127: very nice

packages/sdk-rtl/src/lookerSdkError.ts
jkaster: I see from the new test that the behavior looks fine.

jkaster (changes requested):
minor description corrections on the tests

praphun:
@jkaster I updated the PR as follows:
* changed `LookerSdkError` to `LookerSDKError`
* updated interface to have `I` prefix. `LookerSDKError` itself has to match the function that serves as the constructor. I added comment on that in the file.
* fixed the test names: `template` to `theme`

Sorry, I should have added a fix commit there and squash at the end.

jkaster (approved):
Looks good! Thanks.

packages/sdk-node/src/nodeTransport.ts
```diff
@@ -194,7 +194,7 @@ export class NodeTransport extends BaseTransport {
       }
     } else {
       try {
-        result = Buffer.from(result).toString('binary')
+        result = Buffer.from(result || '').toString('binary')
```
AprilArcus, line 197: I generally prefer the `??` nullish coalescing operator to the boolean union `||` to avoid `OR`ing `0` or `NaN` into the empty string.

Unlike `||` which returns its right-hand-side if the left-hand-side is any falsy value (JavaScript has eight falsy values: `0`, `-0`, `NaN`, `false`, `null`, `undefined`, the empty string, and `document.all`) `??` only returns the right-hand side in the event that the left-hand side is `null` or `undefined` ("nullish").

packages/sdk-rtl/src/lookerSDKError.ts
```diff
@@ -0,0 +1,185 @@
...
+interface IErrorDetail {
+  field?: string | null
+  code?: string | null
+  message?: string | null
+  documentation_url: string | null
+}
+
+// This specifies SDK custom error options
+interface ILookerSDKErrorOptions {
+  errors?: IErrorDetail[]
```
AprilArcus, line 56: `someError.errors` is a non-obvious name for this field. How about `someError.details`?

packages/sdk-rtl/src/lookerSDKError.ts
```diff
@@ -0,0 +1,185 @@
...
+  documentation_url?: string | null
```
AprilArcus, line 57: JavaScript object properties should always be in `snakeCase`, so `documentationUrl`.

packages/sdk-rtl/src/lookerSDKError.ts
```diff
@@ -0,0 +1,185 @@
...
+      // Object.setPrototypeOf() is necessary when extending built-ins,
+      // since Error.call(this, message, errorOptions, ...rest) doesn't
+      // set up the prototype chain the way it would with a user-defined
+      // class.
+      Object.setPrototypeOf(
+        error,
+        this ? Object.getPrototypeOf(this) : LookerSDKError.prototype
+      )
+
+      // imitates the non-enumerability of the 'message' and 'stack'
```
AprilArcus, line 128: Have you thought about whether you want these to be enumerable? Is it better for your API to match the behavior of the non-enumerable `message` and `stack` fields, or not?

Should `Object.keys(someError)` return `['errors', 'documentation_url']`, or should it return `[]`?

Do you want `for (const field in someError) {}` to iterate your custom fields, or to be a no-op?

I don't have a position on this, but you should talk about it with the team before committing one way or the other.

packages/sdk-rtl/src/lookerSDKError.ts
```diff
@@ -0,0 +1,185 @@
...
+}
+
+export const LookerSDKError: ILookerSDKErrorConstructor =
+  /* #__PURE__*/ (() => {
+    'use strict'
+    const LookerSDKErrorConstructor = function LookerSDKError(
+      this: LookerSDKError | undefined,
+      ...[
+        message,
+        { errors = [], documentation_url = '', ...errorOptions } = {},
```
AprilArcus, line 89: The default values you use in this destructuring will only apply in the case that `errors` and `documentation_url` are `undefined` or not present in the input object. How do you want to handle `null`?

jkaster: I think `errors` should end up as an empty array if it's not in the payload. That's what I've done for the Kotlin SDK, fwiw, and consuming the result becomes much simpler that way. Our API error payloads should always have `documentation_url` but I think that should also default to an empty string if not provided

PR #1136, opened 2022-08-04
feat: error code document retrieval

jkaster:
Retrieves the best match error code document for the provided error url from an API error payload

See `errorDoc.spec.ts` for an implementation example


josephaxisa (approved):
lgtm

PR #1139, opened 2022-08-09
feat: Tag scene filters reflected as URL search parameter

patnir41:
Provides the ability to share a link to a method/type tag landing page with an applied filter. For example, being able to load a tag scene already filtered on `POST` in the methods tab or `SPECIFICATION` in the types tab.

**3 Points to Address // Blocked Currently:**

1. If the URL has an invalid parameter like `v=invalid`, we need to be able to get this filter, VALIDATE IT (check if it matches a valid http options or type options), and then update the URL accordingly when changing an invalid filter. This means there needs to be a `navigate` invoked (in APIExplorer?) to update the URL if we have an invalid parameter.

- Could we use the single effect that fires when `initialized` is true in the SDK sharing PR for this?

2. When setting a filter in the tag scene page, then selecting methods under the tag in the SideNav, what is expected behavior with the tag filter?

- Can remove so that when they select a new tag the filter is ALL
- Can persist it in the URL, which would make sense with, for example, a filtered SideNav (see point below); basically choosing a filter sets it for all tags until user selects a new one.

3. Should we filter the SideNav items to those corresponding to a filter when selected? Meaning the SideNav methods/types will only display those that correspond to the filter.

- This is currently implemented as you can see with my changes in `SideNavMethods` and `SideNavTypes`. You can play around with this and see what you think.


packages/api-explorer/src/scenes/MethodTagScene/MethodTagScene.spec.tsx
```diff
@@ -29,16 +29,28 @@ import { screen, waitFor } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 
 import { api } from '../../test-data'
-import { renderWithRouter } from '../../test-utils'
+import { renderWithRouterAndReduxProvider } from '../../test-utils'
 import { MethodTagScene } from './MethodTagScene'
 
 const opBtnNames = /ALL|GET|POST|PUT|PATCH|DELETE/
```
patnir41, line 35: Would be better to have these opBtnNames (also hard-coded in `TypeTagScene.spec.tsx`) in a util file and use for validating?

PR #1140, opened 2022-08-09
feat: Tag scene filter selection as URL search parameter

patnir41:
Provides the ability to share a link to a method/type tag landing page with an applied filter. For example, being able to load a tag scene and have its methods already filtered on `POST`.

### DIFF
--- a/packages/api-explorer/src/ApiExplorer.tsx
+++ b/packages/api-explorer/src/ApiExplorer.tsx
@@ -91,7 +91,8 @@ export const ApiExplorer: FC<ApiExplorerProps> = ({
   const specs = useSelector(selectSpecs)
   const spec = useSelector(selectCurrentSpec)
   const { initLodesAction } = useLodeActions()
-  const { initSettingsAction, setSearchPatternAction } = useSettingActions()
+  const { initSettingsAction, setSearchPatternAction, setTagFilterAction } =
+    useSettingActions()
   const { initSpecsAction, setCurrentSpecAction } = useSpecActions()
 
   const location = useLocation()
@@ -126,7 +127,15 @@ export const ApiExplorer: FC<ApiExplorerProps> = ({
   useEffect(() => {
     const searchParams = new URLSearchParams(location.search)
     const searchPattern = searchParams.get('s') || ''
-    setSearchPatternAction({ searchPattern: searchPattern! })
+    const verbParam = searchParams.get('v') || 'ALL'
+    // TODO: need to validate verbParam, checking against all available
+    //       httpMethod and metaType options, default to ALL if not valid
+    setSearchPatternAction({
+      searchPattern: searchPattern!,
+    })
+    setTagFilterAction({
+      tagFilter: verbParam.toUpperCase(),
+    })
   }, [location.search])
 
   useEffect(() => {
--- a/packages/api-explorer/src/ApiExplorer.tsx
+++ b/packages/api-explorer/src/ApiExplorer.tsx
@@ -68,6 +68,7 @@ import {
   selectSpecs,
   selectCurrentSpec,
   selectSdkLanguage,
+  selectTagFilter,
 } from './state'
 import {
   getSpecKey,
@@ -99,6 +100,7 @@ export const ApiExplorer: FC<ApiExplorerProps> = ({
   const specs = useSelector(selectSpecs)
   const spec = useSelector(selectCurrentSpec)
   const selectedSdkLanguage = useSelector(selectSdkLanguage)
+  const selectedTagFilter = useSelector(selectTagFilter)
   const { initLodesAction } = useLodeActions()
   const {
     initSettingsAction,
@@ -136,10 +138,13 @@ export const ApiExplorer: FC<ApiExplorerProps> = ({
     // reconcile local storage state with URL or vice versa
     if (initialized) {
       const sdkParam = searchParams.get('sdk') || ''
+      const verbParam = searchParams.get('v') || ''
       const sdk = findSdk(sdkParam)
       const validSdkParam =
         !sdkParam.localeCompare(sdk.alias, 'en', { sensitivity: 'base' }) ||
         !sdkParam.localeCompare(sdk.language, 'en', { sensitivity: 'base' })
+      const validVerbParam = isValidFilter(location, verbParam)
+
       if (validSdkParam) {
         // sync store with URL
         setSdkLanguageAction({
@@ -150,6 +155,14 @@ export const ApiExplorer: FC<ApiExplorerProps> = ({
         const { alias } = findSdk(selectedSdkLanguage)
         navigate(location.pathname, {
           sdk: alias === allAlias ? null : alias,
+        })
+      }
+
+      if (validVerbParam) {
+        setTagFilterAction({ tagFilter: verbParam.toUpperCase() })
+      } else {
+        navigate(location.pathname, {
+          v: selectedTagFilter === 'ALL' ? null : selectedTagFilter,
         })
       }
     }
@@ -171,8 +184,6 @@ export const ApiExplorer: FC<ApiExplorerProps> = ({
     const { language: sdkLanguage } = findSdk(sdkParam)
     setSearchPatternAction({ searchPattern })
     setSdkLanguageAction({ sdkLanguage })
-    // TODO: need to validate verbParam, checking against all available
-    //       httpMethod and metaType options, default to ALL if not valid
     setTagFilterAction({
       tagFilter: isValidFilter(location, verbParam)
         ? verbParam.toUpperCase()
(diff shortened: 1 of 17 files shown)


packages/api-explorer/src/components/SelectorContainer/SelectorContainer.tsx
```diff
@@ -51,6 +51,9 @@ export const SelectorContainer: FC<SelectorContainerProps> = ({
   ...spaceProps
 }) => {
   const searchParams = new URLSearchParams(location.search)
+  // TODO: noticing that there are certain pages where we must delete extra params
+  //       before pushing its link, what's a way we can handle this?
```
patnir41, line 55: If this was not included here, navigating to the diffScene after selecting a filter will persist that parameter in the URL. I worry that there will be a growing number of scenes to render where certain  parameters are no longer necessary and must be explicitly removed/nullified to avoid bloating the URL. 
- What if we include an optional `toRemove` parameter to the `buildPath` functions, which is a list of strings corresponding to search params which we will remove from the URL? This could clean up changes in `SideNavMethods` and `SideNavTypes`.

packages/api-explorer/src/components/SideNav/SideNavMethods.tsx
```diff
@@ -85,12 +85,15 @@ export const SideNavMethods = styled(
...
             <li key={method.name}>
               <Link
-                to={`${buildMethodPath(
-                  specKey,
-                  tag,
-                  method.name,
-                  searchParams.toString()
-                )}`}
+                to={() => {
+                  searchParams.delete('v')
```
patnir41, line 89: Relates to prior comment about removing parameters from the URL in a cleaner fashion, pointing out for reviewers

packages/api-explorer/src/ApiExplorer.tsx
```diff
@@ -141,6 +157,14 @@ export const ApiExplorer: FC<ApiExplorerProps> = ({
           sdk: alias === allAlias ? null : alias,
         })
       }
+
```
patnir41, line 160: This gives code smell, as I can see it continually expanding as new parameters are added. I'm wondering if there is any insight on how to go about syncing multiple parameters at once in this initialization effect.

packages/api-explorer/src/components/SelectorContainer/SelectorContainer.tsx
```diff
@@ -51,6 +51,9 @@ export const SelectorContainer: FC<SelectorContainerProps> = ({
...
+  searchParams.delete('v')
```
josephaxisa, line 56: could we replace the link component with say a `Span` that behaves like a link and calls `navigate` on click?

packages/api-explorer/src/scenes/TypeTagScene/TypeTagScene.tsx
```diff
@@ -58,16 +57,31 @@ export const TypeTagScene: FC<TypeTagSceneProps> = ({ api }) => {
...
 
+  useEffect(() => {
+    setTagFilter(selectedTagFilter)
+  }, [selectedTagFilter])
+
   if (!types) {
     return <></>
   }
 
+  const setValue = (filter: string) => {
```
josephaxisa, line 68:
```suggestion
  const handleChange = (filter: string) => {
```

packages/api-explorer/src/scenes/TypeTagScene/TypeTagScene.tsx
```diff
@@ -58,16 +57,31 @@ export const TypeTagScene: FC<TypeTagSceneProps> = ({ api }) => {
...
   const metaTypes = getMetaTypes(types)
   return (
     <ApixSection>
       <DocTitle>{`${tag.name}: ${tag.description}`}</DocTitle>
-      <ButtonToggle mb="small" mt="xlarge" value={value} onChange={setValue}>
+      <ButtonToggle
+        mb="small"
+        mt="xlarge"
+        value={tagFilter}
+        onChange={setValue}
```
josephaxisa, line 83:
```suggestion
        onChange={handleChange}
```

packages/api-explorer/src/scenes/TypeTagScene/TypeTagScene.tsx
```diff
@@ -79,9 +93,20 @@ export const TypeTagScene: FC<TypeTagSceneProps> = ({ api }) => {
       </ButtonToggle>
       {Object.values(types).map(
         (type, index) =>
-          (value === 'ALL' ||
-            value === type.metaType.toString().toUpperCase()) && (
-            <Link key={index} to={buildTypePath(specKey, tag.name, type.name)}>
+          (selectedTagFilter === 'ALL' ||
+            selectedTagFilter === type.metaType.toString().toUpperCase()) && (
```
josephaxisa, line 97: I'm leaning towards not having this check. If user passes in an invalid value for a tag filter

packages/api-explorer/src/utils/hooks.ts
```diff
@@ -37,7 +37,11 @@ export const useNavigation = () => {
 
   const navigate = (
     path: string,
-    queryParams?: { s?: string | null; sdk?: string | null } | null
+    queryParams?: {
+      s?: string | null
+      sdk?: string | null
+      v?: string | null
+    } | null
```
josephaxisa, line 44: abstract into separate interface and document each property

packages/api-explorer/src/utils/path.ts
```diff
@@ -128,3 +132,35 @@ export const getSpecKey = (location: HLocation | Location): string | null => {
...
 }
+
+/**
+ * Gets the scene type of the current page
+ * @param location browser location
+ * @returns string representing the scene type
+ */
+export const getSceneType = (location: HLocation | Location) => {
+  const match = matchPath<{ tabType: string }>(location.pathname, {
+    path: '/:specKey/:tabType',
```
josephaxisa, line 143: did you mean `tagType`?

packages/api-explorer/src/utils/path.ts
```diff
@@ -128,3 +132,35 @@ export const getSpecKey = (location: HLocation | Location): string | null => {
   }
   return match?.groups?.specKey || null
 }
+
+/**
+ * Gets the scene type of the current page
+ * @param location browser location
+ * @returns string representing the scene type
+ */
+export const getSceneType = (location: HLocation | Location) => {
```
josephaxisa, line 141: I'd personally prefer if you passed in path as a string in here and performed the regex match on it.

packages/api-explorer/src/test-data/tagFilters.ts
```diff
@@ -0,0 +1,28 @@
...
+ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+
+ */
+
+export const methodFilters = ['GET', 'POST', 'PUT', 'PATCH', 'DELETE']
```
josephaxisa, line 27: nit: if you don't think these will be used in any other test file then it might be better to put them in the test file itself.

packages/api-explorer/src/ApiExplorer.tsx
josephaxisa: I think a good way to do this will be to augment `initSettings` so that url parameters can be passed in and override persisted settings when applicable. As a result, this effect will then become a one liner that syncs the url with the store using `navigate`.

packages/api-explorer/src/utils/path.ts
```diff
@@ -128,3 +132,35 @@ export const getSpecKey = (location: HLocation | Location): string | null => {
...
+  return match ? match!.params.tabType : ''
+}
+
+/**
+ * Confirms if filter is valid for the page scene type
+ * @param location browser location
+ * @param filter filter tag for page
+ */
+export const isValidFilter = (
+  location: HLocation | Location,
```
patnir41, line 154: Given you preferred having the string pathname for the method `getSceneType`, would you prefer changing this from a location object to a string?

packages/api-explorer/src/scenes/TypeTagScene/TypeTagScene.tsx
patnir41: Did you mean to finish this comment? If a user passes in an invalid value for a tag filter, the expectation is that we handle this and default the user to have selected 'ALL', thus all of the options would show.

PR #1141, opened 2022-08-10
fix: add field details for SDK error

praphun:
Looker API returns field-level information in the error. This change surfaces the information when using the SDK.


packages/sdk-node/src/nodeTransport.ts
```diff
@@ -194,7 +194,7 @@ export class NodeTransport extends BaseTransport {
       }
     } else {
       try {
-        result = Buffer.from(result).toString('binary')
+        result = Buffer.from(result ?? '').toString('binary')
```
praphun, line 197: Changed from `||` to `??` per @AprilArcus's comment to return empty string only when `null` or `undefined`.

packages/sdk-rtl/src/lookerSDKError.ts
```diff
@@ -0,0 +1,191 @@
...
+          configurable: true,
+          enumerable: true,
+        },
+        documentation_url: {
+          value: documentation_url ?? '',
+          writable: true,
+          configurable: true,
+          enumerable: true,
+        },
+      })
```
praphun, line 148: * Change enumerability per discussion with @AprilArcus and @jkaster. `Object.keys(someError)` now returns `['message', 'errors', 'documentation_url']`

* Add `null` handling for `errors` and `documentation_url` according to the comment in the previous PR: https://github.com/looker-open-source/sdk-codegen/pull/1134

packages/sdk-rtl/src/lookerSDKError.ts
```diff
@@ -0,0 +1,191 @@
...
+  field?: string | null
+  code?: string | null
+  message?: string | null
+  documentation_url: string | null
+}
+
+// This specifies SDK custom error options
+interface ILookerSDKErrorOptions {
+  errors?: IErrorDetail[]
+  documentation_url?: string | null
```
praphun, line 57: While snake-casing is not javascript style, we decided to leave this topic out of this PR.

praphun, line 57: Renaming `errors` to `details` would require updating related documentation, and the snake-case `documentation_url` is not Javascript style. We decided to leave these concerns out of this PR.

jkaster (approved):
Nice updates, thanks!

packages/sdk-rtl/src/lookerSDKError.ts
```diff
@@ -0,0 +1,191 @@
...
+}
+
+export const LookerSDKError: ILookerSDKErrorConstructor =
+  /* #__PURE__*/ (() => {
+    'use strict'
+    const LookerSDKErrorConstructor = function LookerSDKError(
+      this: LookerSDKError | undefined,
+      ...[
+        message,
+        { errors = [], documentation_url = '', ...errorOptions } = {},
```
AprilArcus, line 89: it's sort of guilding the lily to supply default arguments in both the call signature *and* the point of usage.

packages/sdk-rtl/src/lookerSDKError.ts
AprilArcus: I'd suggest writing this in the simpler style:

```
Object.defineProperty(error, 'message', { enumerable: true })
error.errors = errors
error.documentation_url = documentation_url
```

three lines vs. sixteen.

praphun: @AprilArcus I had tried this based on the discussion on the previous PR. I have got the error below, so I went with the code above.
```
error TS2339: Property 'errors' does not exist on type 'Error'.
```

AprilArcus: Do a cast first. `(error as LookerSDKError).errors = errors`

packages/sdk-rtl/src/lookerSDKError.ts
praphun: Changed as suggested in the new revision. Thanks @AprilArcus

AprilArcus (approved):

PR #1142, opened 2022-08-10
feat: ability to share a tag scene with a filter applied

patnir41:
Provides the ability to share a link to a method/type tag landing page with an applied filter. For example, being able to load a tag scene and have its methods already filtered on POST.


packages/api-explorer/src/ApiExplorer.tsx
```diff
@@ -86,23 +85,20 @@ export const ApiExplorer: FC<ApiExplorerProps> = ({
...
   const spec = useSelector(selectCurrentSpec)
-  const selectedSdkLanguage = useSelector(selectSdkLanguage)
   const { initLodesAction } = useLodeActions()
   const { initSettingsAction, setSearchPatternAction, setSdkLanguageAction } =
     useSettingActions()
   const { initSpecsAction, setCurrentSpecAction } = useSpecActions()
 
   const location = useLocation()
-  const navigate = useNavigation()
+  useGlobalSync()
```
patnir41, line 98: Wanted to follow up with @bryans99 on how to handle this return value.

packages/api-explorer/src/components/SideNav/SideNavMethods.tsx
```diff
@@ -85,12 +85,16 @@ export const SideNavMethods = styled(
...
             <li key={method.name}>
               <Link
-                to={`${buildMethodPath(
-                  specKey,
-                  tag,
-                  method.name,
-                  searchParams.toString()
-                )}`}
+                to={() => {
+                  // TODO: span behaving like link with custom navigate?
```
patnir41, line 89: Wanted to follow up with @josephaxisa in call to understand the approach for handling this.

packages/api-explorer/src/ApiExplorer.tsx
```diff
@@ -178,7 +155,7 @@ export const ApiExplorer: FC<ApiExplorerProps> = ({
     neededSpec = spec?.key
   }
 
-  return (
+  return isSynced ? (
```
patnir41, line 158: Adding this here to discuss with @bryans99 . I don't know if this check here is needed or how to go about using this synced variable, considering that without it I get the same behavior and tests passing.

packages/api-explorer/src/components/SideNav/SideNav.tsx
```diff
@@ -84,12 +84,12 @@ export const SideNav: FC<SideNavProps> = ({ headless = false, spec }) => {
     if (parts[1] === 'diff') {
       if (parts[3] !== tabNames[index]) {
         parts[3] = tabNames[index]
-        navigate(parts.join('/'))
+        navigate(parts.join('/'), { v: null })
```
patnir41, line 87: This would use navigateWithGlobal instead

packages/api-explorer/src/components/SideNav/SideNavMethods.tsx
```diff
@@ -85,12 +83,9 @@ export const SideNavMethods = styled(
           {Object.values(methods).map((method) => (
             <li key={method.name}>
               <Link
-                to={`${buildMethodPath(
-                  specKey,
-                  tag,
-                  method.name,
-                  searchParams.toString()
-                )}`}
+                to={buildPathWithGlobal(
```
patnir41, line 86: Pointing out for reviewers

packages/api-explorer/src/scenes/utils/hooks/tagStoreSync.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+
+ */
+
+// TODO: testing the hook outside of a component?
```
patnir41, line 27: How can I approach testing a hook independent from a component?

packages/api-explorer/src/utils/hooks/globalStoreSync.spec.tsx
```diff
@@ -0,0 +1,115 @@
...
+    })
+    expect(history.push).toHaveBeenCalledWith({
+      pathname: '/',
+      search: 'sdk=kt',
+    })
+  })
+
+  /**
+   * TODO: test cases
+   * 1 - url valid params are synced to store
```
patnir41, line 93: Just need to understand how to watch store actions / payload and ensure store is being updated correctly (the goal here is to check both that the action has been dispatched and then check that the store is updated as well)

packages/api-explorer/src/scenes/utils/hooks/tagStoreSync.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+  })
+
+  test('does nothing if uninitialized', () => {
+    const { push } = useHistory()
+    const wrapper = ({ children }: any) => withReduxProvider(children)
+    renderHook(() => useTagStoreSync(), { wrapper })
+    expect(push).not.toHaveBeenCalled()
+  })
+
+  describe.each([
```
josephaxisa, line 57: test.each?

packages/api-explorer/src/scenes/TypeTagScene/TypeTagScene.spec.tsx
```diff
@@ -74,30 +86,28 @@ describe('TypeTagScene', () => {
     ).toHaveLength(2)
   })
 
-  test('it filters methods by operation type', async () => {
+  test('it pushes filter to URL on toggle', async () => {
```
josephaxisa, line 89: what about a test asserting that the right tag is selected when the scene is rendered with a filter value already in the store?

packages/api-explorer/src/utils/hooks/navigation.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+  }
+
+  /**
+   * Builds path to a scene and removes any scene-specific URL parameters
+   *
+   * @param path the destination path
+   * @returns a path excluding scene-specific search parameters
+   */
+  const buildPathWithGlobal = (path: string) => {
```
josephaxisa, line 78:
```suggestion
  const buildPathWithGlobalParams = (path: string) => {
```

packages/api-explorer/src/utils/hooks/navigation.ts
```diff
@@ -0,0 +1,134 @@
...
+    params.delete('v')
+    return `${path}?${params.toString()}`
+  }
+
+  /**
+   * Navigates to a scene removing any scene-specific URL parameters
+   *
+   * @param path Pathname to navigate to
+   */
+  const navigateWithGlobal = (path: string) => {
```
josephaxisa, line 89:
```suggestion
  const navigateWithGlobalParams = (path: string) => {
```

packages/api-explorer/src/utils/hooks/navigation.ts
```diff
@@ -0,0 +1,134 @@
...
+  //  */
+  // const buildTypeScenePath = (
+  //   specKey: string,
+  //   tag: string,
+  //   typeName: string
+  // ) => {
+  //   const params = new URLSearchParams(location.search)
+  //   params.delete('v')
+  //   return `/${specKey}/types/${tag}/${typeName}${params.toString()}`
+  // }
```
josephaxisa, line 127: cruft?

packages/api-explorer/src/utils/path.ts
```diff
@@ -128,3 +132,35 @@ export const getSpecKey = (location: HLocation | Location): string | null => {
...
+  return match ? match!.params.tagType : ''
+}
+
+/**
+ * Confirms if filter is valid for the page scene type
+ * @param location browser location
+ * @param filter filter tag for page
+ */
+export const isValidFilter = (
+  location: HLocation | Location,
```
josephaxisa, line 154: pass in the path instead?

packages/api-explorer/src/utils/path.ts
```diff
@@ -128,3 +132,35 @@ export const getSpecKey = (location: HLocation | Location): string | null => {
...
+ */
+export const getSceneType = (path: string) => {
+  const match = matchPath<{ tagType: string }>(path, {
+    path: '/:specKey/:tagType',
+  })
+  return match ? match!.params.tagType : ''
+}
+
+/**
+ * Confirms if filter is valid for the page scene type
```
josephaxisa, line 149:
```suggestion
 * Confirms if filter is valid for a given method/type tag
```

packages/api-explorer/src/utils/path.ts
```diff
@@ -128,3 +132,35 @@ export const getSpecKey = (location: HLocation | Location): string | null => {
...
+) => {
+  const sceneType = getSceneType(location.pathname)
+  if (!sceneType) return false
+  else if (!filter.localeCompare('all', 'en', { sensitivity: 'base' }))
+    return true
+  else if (sceneType === 'methods') {
+    return methodFilterOptions.test(filter)
+  } else {
+    return typeFilterOptions.test(filter)
+  }
```
josephaxisa, line 165: consider using a switch statement or having a variable with the result and one return statement.

josephaxisa (changes requested):
good job. a few comments below

packages/api-explorer/src/scenes/utils/hooks/tagStoreSync.spec.ts
patnir41: I would be nesting tests if I replaced this with test.each. The intention was to account for the 2 different tabs, methods and types, where this filtering logic must correctly apply. Each tab has their own respective filters, and I was initially considering the case where we are in a tab (say methods) with the URL containing a type tag filter (say v=specification). I decided against having that explicit of a test though because I believe this is covered by my testing of "invalid" filters.

Thoughts?

packages/api-explorer/src/scenes/TypeTagScene/TypeTagScene.spec.tsx
patnir41: In a scenario where the URL has no verb filter param, the store has a selected verb filter, and we are rendering the page: upon execution of the reconciliation hook, because there is no verb filter param in the URL, the page will render having selected the 'ALL' filter. Is this the assertion you want me to make?

packages/api-explorer/src/utils/hooks/navigation.ts
```diff
@@ -59,5 +68,30 @@ export const useNavigation = () => {
...
+   * @param path Pathname to navigate to
+   */
+  const navigateWithGlobalParams = (path: string) => {
+    history.push(buildPathWithGlobalParams(path))
+  }
+
+  return {
+    navigate,
+    navigateWithGlobal: navigateWithGlobalParams,
+    buildPathWithGlobal: buildPathWithGlobalParams,
```
patnir41, line 95: Will fix in next commit: functions should have only a single name/reference

packages/api-explorer/src/scenes/utils/hooks/tagStoreSync.spec.ts
josephaxisa: I misread, this is fine as is.

packages/api-explorer/src/scenes/TypeTagScene/TypeTagScene.tsx
```diff
@@ -43,21 +46,40 @@ interface TypeTagSceneParams {
...
+      v: filter === 'ALL' ? null : filter.toLowerCase(),
+    })
+  }
 
   useEffect(() => {
-    /** Reset ButtonToggle value on route change */
-    setValue('ALL')
-  }, [typeTag])
+    const searchParams = new URLSearchParams(location.search)
+    let verbParam = searchParams.get('v') || 'ALL'
```
josephaxisa, line 63: `v` doesn't seem very applicable to type tag scenes. Perhaps we should use `f` for filter on this scene instead. @jkaster thoughts?

josephaxisa (commented):
tested locally and it works as expected. left one question below that's worth discussing before this is merged.

packages/api-explorer/src/scenes/TypeTagScene/TypeTagScene.tsx
jkaster: What would we be filtering? `f` would be fine, but I don't know what it's used for. Maybe `t` for type?

jkaster: or `t` for tag, depending on what's being filtered 🤷🏼

josephaxisa: @patnir41 let's go with `t` for type on type tag scenes.

patnir41: @bryans99 We discussed the above off-thread and came across 2 potential decisions:

1. instead of using `v`, have the parameter be `t` for both MethodTag and TypeTag scenes. 
2. Customize the `tagStoreSync` hook to take a character argument, either `v` or `t`, which will determine which parameter the hook is syncing when rendering a scene.  We then have TypeTag scenes involve logic dealing with a `t` parameter value, and MethodTag scenes involve logic dealing with `v` in the URL.

Do you have any thoughts?

packages/api-explorer/src/utils/index.ts
```diff
@@ -30,4 +30,5 @@ export { getLoded } from './lodeUtils'
 export { useWindowSize } from './useWindowSize'
 export * from './apixAdaptor'
 export * from './adaptorUtils'
-export { useNavigation } from './hooks'
+export { useNavigation } from './hooks/navigation'
```
patnir41, line 33: Export what you need from hooks into `index.ts` and fix

packages/api-explorer/src/utils/hooks/navigation.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+      },
+    }),
+  }
+})
+
+describe('useNavigation', () => {
+  const history = useHistory()
+  const { navigate, navigateWithGlobalParams, buildPathWithGlobalParams } =
+    useNavigation()
+  const curParams = new URLSearchParams(history.location.search) // 's=test&sdk=py&v=get'
```
patnir41, line 48: Remove comment

packages/api-explorer/src/utils/hooks/navigation.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+        v: newParams.get('v'),
+      })
+      expect(mockHistoryPush).lastCalledWith({
+        pathname: route,
+        search: newParams.toString(),
+      })
+    })
+  })
+
+  describe('buildPathWithGlobal', () => {
```
patnir41, line 95: Fix describe names

packages/api-explorer/src/scenes/TypeTagScene/TypeTagScene.tsx
patnir41: Moving forward with renaming parameter from 'v' to 'f' for both scenes.

tsconfig.common.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "lib": ["ES2019", "dom"],
+    "lib": ["ES2019", "dom", "dom.iterable"],
```
patnir41, line 3: Needed in order to register URLSearchParams as an iterable object for `buildPathWithGlobalParams` in `navigation.ts`.

packages/api-explorer/src/utils/hooks/navigation.ts
```diff
@@ -76,7 +76,11 @@ export const useNavigation = () => {
    */
   const buildPathWithGlobalParams = (path: string) => {
     const params = new URLSearchParams(history.location.search)
-    params.delete('v')
+    for (const key of params.keys()) {
+      if (key !== 's' && key !== 'sdk') {
```
josephaxisa, line 80: store all global params in an array and check that array for validity. i.e this check becomes:
`if (!globalParams.includes(key) params.delete(key)`

josephaxisa (changes requested):

packages/api-explorer/src/scenes/TypeTagScene/TypeTagScene.tsx
```diff
@@ -28,8 +28,11 @@ import React, { useEffect, useState } from 'react'
 import { Grid, ButtonToggle, ButtonItem } from '@looker/components'
 import type { ApiModel } from '@looker/sdk-codegen'
 import { useParams } from 'react-router-dom'
+import { useSelector } from 'react-redux'
 import { ApixSection, DocTitle, DocTypeSummary, Link } from '../../components'
-import { buildTypePath, useNavigation } from '../../utils'
+import { buildTypePath, isValidFilter, useNavigation } from '../../utils'
+import { selectTagFilter, useSettingActions } from '../../state'
+import { useTagStoreSync } from '../utils/hooks/tagStoreSync'
```
josephaxisa, line 35: you shouldn't need to reach for imports. It should be from `../utils`.

josephaxisa (approved):
lgtm

PR #1143, opened 2022-08-12
fix: add ErrorDoc class to the Kotlin SDK

jkaster:
Utility class to retrieve available error documentation from the API error docs CDN


kotlin/src/main/com/looker/rtl/ErrorDoc.kt
```diff
@@ -0,0 +1,224 @@
...
+    /** api version of the error link */
+    var apiVersion: String
+    /** HTTP status code */
+    var statusCode: String
+    /** REST API Path */
+    var apiPath: String
+ }
+
+ interface IErrorDoc {
+    /** Index of all know error codes. Call load() to populate it */
```
praphun, line 45: typo:
```suggestion
    /** Index of all known error codes. Call load() to populate it */
```

packages/sdk-rtl/src/errorDoc.spec.ts
```diff
@@ -41,52 +41,11 @@ import type {
...
-    url: 'run_inline_query_422.md',
-  },
-  '404/get/looks/{look_id}/run/{result_format}': {
-    url: 'run_look_404.md',
-  },
-}
-
+const config = TestConfig()
+const errorIndex = `${config.testPath}errorCodesIndex.json`
+const sampleIndex: ErrorCodeIndex = JSON.parse(readFile(errorIndex))
```
praphun, line 48: 👍 extract to a separate json file

praphun (approved):
LGTM

I've noticed some formatter (or IDE) was doing some stuff to import statements (changing individual to '*', reordering, both individual and '*' import, etc). Maybe we should set up checkstyle in the future.

jkaster:
@praphun yes, an opinionated Linter for Kotlin is one of the things we'd like to have. I think it's already a ticket in our backlog. The code changes in this PR were mainly caused by the "spotless kotlin" gradle process.

PR #1146, opened 2022-08-15
fix: IErrorDetail fields should not be null

praphun:
fields should not be null if we have `errors` info in the response


jkaster (approved):
LGTM

PR #1147, opened 2022-08-16
docs: Readme - Deprecate 3.x

fabio-looker:
Adds notes to readme clarifying that 3.x is deprecated, 4.0 is stable, and linking to 3.0 deprecation notice.


josephaxisa (approved):
lgtm. I'd wait for John to review as well before merging

README.md
```diff
@@ -26,37 +26,38 @@ An [OpenAPI specification](https://github.com/OAI/OpenAPI-Specification) describ
 
 A Looker SDK has several parts:
 
-- **Looker API** OpenAPI specification (e.g., found at
-  `https://<your-looker-endpoint>:19999/api/3.1/swagger.json`, although this is still the Swagger 2.x representation)
+- The **Looker API**, described by an OpenAPI specification (e.g., the Swagger 2.x representation found at
+  `https://<your-looker-domain>:19999/api/4.0/swagger.json`). The 4.0 API is our current & stable API. As of June 2022, [3.x is deprecated](https://developers.looker.com/api/advanced-usage/version-3x-deprecation).
 
 - The **Looker API Docs viewer**, provided in the Looker web app directly from our version-specific OpenAPI specification, available on each Looker server instance.
```
jkaster, line 32: The API Docs viewer is gone now, and has been replaced by API Explorer, which is available to install for each instance and also on the developer portal

README.md
```diff
@@ -26,37 +26,38 @@ An [OpenAPI specification](https://github.com/OAI/OpenAPI-Specification) describ
...
 - **Language SDKs**, "smarter" client language classes and methods to improve the experience of calling the Looker API in various popular coding languages.
 
-## Multi-API support with Looker 7.2 and later
+## SDK multi-API-version support
 
-Looker 7.2 introduced an **Experimental** version of API 4.0. Since that release, some SDKs now support multiple API versions in the same SDK package.
+The 4.0 version of the API is the current and stable version of the API, in addition to the 3.x API which is now [deprecated](https://developers.looker.com/api/advanced-usage/version-3x-deprecation).
 
-For all SDKs but Swift, API-specific SDKs are now created and put in the same SDK package, and share the same run-time code.
+Many SDKs support and expose both API versions in the same SDK package, including all [Looker-supported](https://docs.looker.com/reference/api-and-integration/api-sdk-support-policy#looker_supported) SDKs other than Swift.
```
jkaster, line 40: I'm pretty sure we've dropped 3.1 from Kotlin and Go as well. We should also drop it from Look# if we haven't yet because no one should use API 3.x on a statically typed language.

README.md
```diff
@@ -26,37 +26,38 @@ An [OpenAPI specification](https://github.com/OAI/OpenAPI-Specification) describ
...
+| [Python](python)           | `looker_sdk.init31()`                                                                           | `looker_sdk.init40()`                                                    |                                                                                                                                                            |
+| [TypeScript](packages/sdk) | `Looker31SDK()`, `LookerNodeSDK.init31()`, or `LookerBrowserSDK.init31()`                       | `Looker40SDK()`, `LookerNodeSDK.init40()` or `LookerBrowserSDK.init40()` | **Important** - See information on the [typescript SDK dependencies](#very-important-note-regarding-the-looker-typescript-sdk) at the bottom of this file. |
+| [Kotlin](kotlin)           | Do not use                                                                                      | `LookerSDK()`                                                            | API 4.0 was created to better serve strongly-typed languages, like Kotlin. Kotlin only works with API 4.0. Also, the initializer uses an unversioned name. |
```
jkaster, line 57: If 3.1 is gone from Kotlin, you can make it "Not applicable" as well.

README.md
```diff
@@ -176,7 +177,7 @@ To generate a language currently not supported by Looker's SDK code generator wi
 
 - configure the desired language in [`codeGenerators.ts`](packages/sdk-codegen/src/codeGenerators.ts). Currently, only Go and C# have legacy language generation configured, and we now have a prototype Look# SDK that can be used instead of the legacy C# generator.
 
-- the legacy generator defaults to using the API 4.0 specification, which is more accurate for strongly typed languages. To use API 3.1, put `api_version=3.1` in the `Looker` section of your `looker.ini`
+- the legacy generator defaults to using the API 4.0 specification. To use deprecated API 3.1, specify `api_version=3.1` in the `Looker` section of your `looker.ini`
```
jkaster, line 180: Did you actually test this? I haven't tried the legacy generator in literally a year or more.

README.md
```diff
@@ -281,7 +282,7 @@ The following table describes the environment variables. By default, the SDK "na
 | Variable name           | Description                                                                                                                                                           |
 | ----------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | LOOKERSDK_BASE_URL      | A URL like `https://my.looker.com:19999`. No default value.                                                                                                           |
-| LOOKERSDK_API_VERSION   | Version of the Looker API to use. Use `3.1` for now, which is the default and used to produce this SDK.                                                               |
+| LOOKERSDK_API_VERSION   | Version of the Looker API to use. Use `4.0` for the current API version.                                                                                              |
```
jkaster, line 285: I wonder where we're actually using this now. Worth spending a few minutes verifying whether we still use it or not.

jkaster (changes requested):
Great job sweeping through the docs. This will be a good update.

README.md
fabio-looker: Updated to API Explorer

README.md
fabio-looker: I've cleaned this up a bit by focusing the language primarily around the "Looker supported" SDKs

README.md
fabio-looker: Updated all strongly typed SDKs to 3.1 "Not supported"

README.md
fabio-looker: I tested it and it threw an opaque error from sdk-rtl/lib/transport.js:221... although I couldn't immediately tell if this was a resolvable network/instance/config issue, it seems like your comment is suggesting that this section should be removed, so I've tentatively removed it.

README.md
fabio-looker: Actually used in only 1 place:
- In a "check'' against the API login endpoint during a "wait for looker" github script.

Programmatically passed-through / typed in 2 places
- Docker config passes it around
- Go type delarations include it

Doc'd in 4 places:
- This main readme
- Two env sample files
- A Python notebook example

From the one use, I think the default value (4.0) should be fine (and indistinguishable in functionality from 3.1) in virtually every situation besides trying to build against a _very_ old Looker version, so maybe we leave the variable there but remove it from the documentation?

README.md
jkaster: I'm good with removing it

README.md
jkaster: I think removing it from the doc is a good step

jkaster (commented):
Good updates. I think there's one pending removal you called out, but this is looking good so far.

README.md
```diff
@@ -22,41 +22,40 @@ This repository contains:
...
+## SDK multi-API-version support
 
-## Multi-API support with Looker 7.2 and later
+The 4.0 version of the API is the current and stable version of the API, in addition to the 3.x API which is now [deprecated](https://developers.looker.com/api/advanced-usage/version-3x-deprecation).
 
-Looker 7.2 introduced an **Experimental** version of API 4.0. Since that release, some SDKs now support multiple API versions in the same SDK package.
+Some SDKs support and expose both API versions in the same SDK package, including all [Looker-supported SDKs](https://docs.looker.com/reference/api-and-integration/api-sdk-support-policy#language_sdks).
 
-For all SDKs but Swift, API-specific SDKs are now created and put in the same SDK package, and share the same run-time code.
+For SDKs that supports multiple API versions, there will be `methods.*` and `models.*` collections generated for each API version. Each API version is exposed under a distinct class name from which to instantiate an initial SDK object.
```
jkaster, line 40:
```suggestion
For SDKs that support multiple API versions, there will be `methods.*` and `models.*` collections generated for each API version. Each API version is exposed under a distinct class name from which to instantiate an initial SDK object.
```

README.md
```diff
@@ -170,16 +169,6 @@ python
 
 **Note:** If you're unable to download the API specification file because you're using an instance of Looker that is self-signed and errors are thrown, you can explicitly turn off SSL verification by putting `verify_ssl=false` in the `looker.ini` file configuration section.
 
-### Using the Legacy generator
```
jkaster, line 173: We should keep this section in because the legacy generator can be used to generate language SDKs we don't have.

jkaster (changes requested):
We're very close

README.md
fabio-looker: I added the section back, but omitted the commentary around which SDKs are "currently" available and around changing the API versions (the looker-sample.ini file still explains the apiVersions property with a comment in that file)

jkaster (approved):
Looks great!

PR #1158, opened 2022-08-18
feat: ability to share a preconfigured diff scene

patnir41:
Selecting comparison options in diff scenes will update an `opts` parameter in the URL, which is used to drive the state of the selected options in the scene.


patnir41:
(For @josephaxisa) I created a new branch from the previous [PR](https://github.com/looker-open-source/sdk-codegen/pull/1140). After making some final changes to that PR and merging it into `main`, I merged `main` into this branch. I thought conflicts were resolved until I pushed and saw there were still some conflicts which I resolved in the GitHub editor.

packages/api-explorer/src/components/SideNav/SideNav.tsx
```diff
@@ -84,12 +84,12 @@ export const SideNav: FC<SideNavProps> = ({ headless = false, spec }) => {
     if (parts[1] === 'diff') {
       if (parts[3] !== tabNames[index]) {
         parts[3] = tabNames[index]
-        navigateWithGlobalParams(parts.join('/'))
+        navigate(parts.join('/'), { t: null })
```
patnir41, line 87: The reason for this change: if we are in a diff scene when we have `opts` in the url, and we open the SideNav to switch to the `Types` or `Method` tab, the diff scene shouldn't change - we should maintain the current parameter as is instead of resetting it. So in this case, I feel it's appropriate that navigation should specifically remove the tag filter variable only.

packages/api-explorer/src/components/SideNav/SideNavMethods.tsx
```diff
@@ -54,9 +54,9 @@ export const SideNavMethods = styled(
       const _isOpen = !isOpen
       setIsOpen(_isOpen)
       if (_isOpen) {
-        navigate(`/${specKey}/methods/${tag}`)
+        navigate(`/${specKey}/methods/${tag}`, { opts: null })
```
patnir41, line 57: The scenario to consider here (and with the same change in `SideNavTypes`: we are in a diff scene with selected options, and `opts` param in the url. When we open the sideNav and open a method/tag scene, this opts parameter should no longer persist. The reason we don't use `navigateWithGlobalParams` to remove this parameter is because if we started at a tag scene with a filter selected and want to select another tag, this tag should still persist when we open the new tag scene.

So, technically, the purpose of this line in plain english is more like "navigate to the tag scene and ONLY keep the tag filter variable if we have one in the URL" 

This makes me reconsider what we initially had proposed where we have a specific `navigateTo` methods. I think in the long term there will be 2 cases:

1. We navigate to a scene, remove all scene-specific params first, then push global params (currently the case with `navigateWithGlobalParams`, etc)
2. We navigate to a scene, keeping ONLY specific parameters, and removing anything else in the URL. 

To better account for case 2, I think having custom navigate methods makes the most sense. Here we'd explicitly have a `navigateToTagScene` instead of having to explicitly nullify any extra parameters that could persist when navigating to a tag scene from the SideNav.

packages/api-explorer/src/scenes/utils/hooks/diffStoreSync.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  useEffect(() => {
+    if (initialized) {
+      const params = new URLSearchParams(location.search)
+
+      // syncing diff options on diff scene page
+      const diffOptionsParam = getDiffOptionsFromUrl(params.get('opts'))
+      if (diffOptionsParam) {
+        setDiffOptionsAction({ diffOptions: diffOptionsParam })
+        navigate(location.pathname, { opts: diffOptionsParam.join(',') })
```
patnir41, line 57: The reason for this `navigate` here, which contrasts from `tagStoreSync` or `globalStoreSync` where we just dispatch to the store, is in the case where we have a list of options in the url, but one or more of those are not actually valid comparison options for the diff scene.

`getDiffOptionsFromUrl` does the job of filtering out invalid comparison options from the URL and updating the URL to only have the valid options out of those listed in the parameter.

packages/api-explorer/src/scenes/DiffScene/DiffScene.spec.tsx
```diff
@@ -0,0 +1,147 @@
...
+        name: 'Missing',
+      })
+    )
+    await waitFor(() => {
+      expect(push).toHaveBeenCalledWith({
+        pathname: '/',
+        search: 'opts=missing',
+      })
+    })
+    // TODO: test URL change leads to store dispatch? - change mock history push implementation to change our location
```
patnir41, line 102: WIP

packages/api-explorer/src/scenes/DiffScene/DiffScene.spec.tsx
```diff
@@ -0,0 +1,162 @@
...
+  // })
+  // ;(getApixAdaptor as jest.Mock).mockReturnValue(mockApixAdaptor)
+  Element.prototype.scrollTo = jest.fn()
+  Element.prototype.scrollIntoView = jest.fn()
+
+  const toggleNavigation = () => false
+  test.only('updating url dispatches store action', () => {
+    // ;(useDispatch as jest.Mock).mockReturnValue(mockDispatch)
+    // TODO issue: executing createTestStore here doesn't init default settings
+    const store = createTestStore({
```
patnir41, line 98: Issue: Having a test that just executes `createTestStore` in `DiffScene.spec.tsx` will fail to initialize the store with default settings, only taking values from the overrides.

packages/api-explorer/e2e/diffScene.spec.ts
```diff
@@ -265,4 +270,70 @@ describe('Diff Scene', () => {
     expect(diff40to31Page1Methods).toHaveLength(15)
     expect(diff40to31Page1Methods).toContain('delete_board_item')
   })
+
+  it('updates when a comparison option is toggled', async () => {
```
patnir41, line 274: Can remove this, but I thought it would be nice to have an integration test ensuring the DiffScene operates properly other than the unit tests which I couldn't make comprehensive due to the spec file bug.

packages/api-explorer/src/scenes/DiffScene/diffUtils.ts
```diff
@@ -118,3 +118,19 @@ export const diffToSpec = (
...
+export const getValidDiffOptions = (opts: string | null) => {
+  if (!opts) return null
+  const diffOptions: string[] = []
+  for (const option of opts.split(',')) {
+    const op = option.toLowerCase()
+    if (allDiffToggles.includes(op) && !diffOptions.includes(op)) {
+      diffOptions.push(option.toLowerCase())
+    }
+  }
+  return diffOptions.length ? diffOptions : null
```
josephaxisa, line 135: why not always return an array? i.e why not return an empty array if there are no valid options?

packages/api-explorer/src/scenes/DiffScene/diffUtils.ts
```diff
@@ -118,3 +118,19 @@ export const diffToSpec = (
   result.types = {}
   return result
 }
+
+/**
+ * Returns all valid diff options contained in list
+ * @param opts url diff options parameter value
+ */
+export const getValidDiffOptions = (opts: string | null) => {
```
josephaxisa, line 126: can opts just be made an optional string arg here? is `null` explicitly needed?

josephaxisa (changes requested):
thanks Niral. I noted some feedback below

packages/api-explorer/src/components/SideNav/SideNavMethods.tsx
patnir41: @josephaxisa

packages/api-explorer/src/scenes/DiffScene/diffUtils.ts
patnir41: I pass in `searchParams.get('opts')` in both DiffScene and diffStoreSync. When the param doesn't exist in the URL it will return null, so this was just a way to be able to handle it without doing any explicit checks before passing it in as input.

packages/api-explorer/src/scenes/DiffScene/diffUtils.ts
patnir41: I think I faced a weird case earlier when working on this before some of our major changes in main, but yes this makes sense.

PR #1159, opened 2022-08-18
feat: change diff scene route

josephaxisa:
from `/diff/:specKey/:compareSpecKey` to `/:specKey/diff/:compareSpecKey `in order for it to follow the same pattern as all other routes

### DIFF
--- a/packages/api-explorer/e2e/diffScene.spec.ts
+++ b/packages/api-explorer/e2e/diffScene.spec.ts
@@ -44,8 +44,9 @@ describe('Diff Scene', () => {
     await jestPuppeteer.resetBrowser()
     await page.setDefaultNavigationTimeout(120000)
   })
-  it('loads the default scene (/diff/3.1)', async () => {
-    await goToPage(`${BASE_URL}/diff/3.1`)
+
+  it('loads the default scene (/3.1/diff)', async () => {
+    await goToPage(`${BASE_URL}/3.1/diff`)
     const body = await page.$('body')
 
     // "Base" input element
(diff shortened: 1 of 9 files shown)


packages/api-explorer/e2e/diffScene.spec.ts
```diff
@@ -44,8 +44,9 @@ describe('Diff Scene', () => {
     await jestPuppeteer.resetBrowser()
     await page.setDefaultNavigationTimeout(120000)
   })
-  it('loads the default scene (/diff/3.1)', async () => {
-    await goToPage(`${BASE_URL}/diff/3.1`)
+
+  it('loads the default scene (/3.1/diff)', async () => {
```
jkaster, line 48: reminds me we need to change this default .. it should be 4.0 already based on spec status

packages/api-explorer/src/ApiExplorer.tsx
```diff
@@ -149,7 +149,7 @@ export const ApiExplorer: FC<ApiExplorerProps> = ({
   const themeOverrides = adaptor.themeOverrides()
 
   let neededSpec = location.pathname?.split('/')[1]
-  if (!neededSpec || neededSpec === diffPath) {
+  if (!neededSpec) {
```
jkaster, line 152: much better

jkaster (approved):
Nice cleanup. I like the consistent pattern, too.

packages/api-explorer/e2e/diffScene.spec.ts
josephaxisa: agreed

PR #1163, opened 2022-08-23
fix: DiffScene Page Displays Accurate Number of Diff Entries

patnir41:
When comparing two specs in APIExplorer's Diff Scene, the scene will display the correct number of diff entries according to the page size, which is 15 by default.

### DIFF
--- a/packages/api-explorer/src/scenes/DiffScene/DocDiff/DocDiff.spec.tsx
+++ b/packages/api-explorer/src/scenes/DiffScene/DocDiff/DocDiff.spec.tsx
@@ -135,4 +135,35 @@ describe('DocDiff', () => {
     const notDisplayedRow = delta[pageSize]
     expect(screen.queryByText(notDisplayedRow.name)).not.toBeInTheDocument()
   })
+
+  it('final diff entry of one page does not appear in next page', async () => {
+    const delta = diffSpecs(leftApi, rightApi, standardDiffToggles)
+    const pageSize = 5
+    renderWithReduxProvider(
+      <DocDiff
+        leftKey={leftKey}
+        leftSpec={leftApi}
+        rightKey={rightKey}
+        rightSpec={rightApi}
+        delta={delta}
+        pageSize={pageSize}
+      />
+    )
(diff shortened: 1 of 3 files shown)


packages/api-explorer/src/scenes/DiffScene/DocDiff/DocDiff.spec.tsx
```diff
@@ -109,4 +109,30 @@ describe('DocDiff', () => {
       ).toBeInTheDocument()
     })
   })
+
+  it('paginates with correct number of entries per page', async () => {
```
jkaster, line 113: is it worth checking the last row isn't the first row on the next page?

patnir41: Yes I agree and feel it wouldn't hurt. See the new additional test.

packages/api-explorer/src/scenes/DiffScene/DocDiff/DocDiff.spec.tsx
```diff
@@ -109,4 +109,61 @@ describe('DocDiff', () => {
...
+        leftKey={leftKey}
+        leftSpec={leftApi}
+        rightKey={rightKey}
+        rightSpec={rightApi}
+        delta={delta}
+        pageSize={pageSize}
+      />
+    )
+
+    for (let i = 0; i < pageSize; i++) {
```
josephaxisa, line 127: nit: our TS guide suggests using `for..of` or `forEach` for looping over iterables.
```suggestion
    for (row of delta) {
```

josephaxisa (commented):
thanks Niral. One minor suggestion below

packages/api-explorer/src/scenes/DiffScene/DocDiff/DocDiff.spec.tsx
josephaxisa: actually, I see how this is necessary given the parameters.

josephaxisa (approved):
lgtm

jkaster (approved):
LGTM

PR #1165, opened 2022-08-24
chore: spec linter improvements

jkaster:
- fix broken csv formatting
- auto update specs to OpenAPI format
- show settings used for run


praphun (approved):
LGTM

I have a general question. There are a couple other places that still use `ApiModel.fromString(readFileSync())` - do we need to `upgradeSpecObject` there as well?

jkaster:
`upgradeSpec()` and `upgradeSpecObject()` are for processing specifications that may be Swagger 2.x instead of OpenAPI 3.x (aka Swagger 3.x). Files that contain `.oas.` in them are already converted to OpenAPI 3.x, and the other usages of `ApiModel.fromString()` all look correct to me.

PR #1167, opened 2022-08-26
feat: error documentation redirect and support for detailed error content

josephaxisa:
Navigating to `/:specKey/err/:errorCode/:verb/*` will redirect users to the correct method scene with an error query param containing the error code. 

To test, navigate to `/4.0/err/400/get/content_metadata/:content_metadata_id`. This change will:
- automatically redirect users to the correct method scene (`h/4.0/methods/Content/content_metadata?e=400` based on the example above)
- scroll the `Responses` section into view
- automatically select the tab based on the error code in the url 
- fetch and render detailed error markdown

### DIFF
--- a/packages/api-explorer/src/scenes/ErrorDetailScene/ErrorDetailScene.tsx
+++ b/packages/api-explorer/src/scenes/ErrorDetailScene/ErrorDetailScene.tsx
@@ -1,0 +1,72 @@
+/*
+
+ MIT License
+
+ Copyright (c) 2022 Looker Data Sciences, Inc.
+
+ Permission is hereby granted, free of charge, to any person obtaining a copy
+ of this software and associated documentation files (the "Software"), to deal
+ in the Software without restriction, including without limitation the rights
+ to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ copies of the Software, and to permit persons to whom the Software is
+ furnished to do so, subject to the following conditions:
+
+ The above copyright notice and this permission notice shall be included in all
+ copies or substantial portions of the Software.
+
+ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+
+ */
+import { Redirect, useRouteMatch } from 'react-router-dom'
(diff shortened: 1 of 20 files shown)


packages/api-explorer/src/scenes/ErrorDetailScene/ErrorDetailScene.tsx
```diff
@@ -0,0 +1,72 @@
...
+export const ErrorDetailScene = ({ api }: ErrorDetailSceneProps) => {
+  const match = useRouteMatch<{
+    specKey: string
+    statusCode: string
+    verb: string
+  }>(`/:specKey/err/:statusCode/:verb/*`)
+  const methodPath = match?.params[0]
+  const adaptor = getEnvAdaptor()
+  const errorDoc = new ErrorDoc(adaptor.sdk)
+  const restPath = errorDoc.specPath(methodPath)
```
jkaster, line 51: Seems we should make ErrorDoc.specPath a class function (as you and I discussed) but that can be done later

packages/api-explorer/src/scenes/MethodScene/MethodScene.tsx
```diff
@@ -115,6 +118,17 @@ export const MethodScene: FC<MethodSceneProps> = ({ api }) => {
     checkRunIt()
   }, [adaptor, setOn])
 
+  useEffect(() => {
+    if (method.responses && errorCode) {
+      const maybe = true
+      if (maybe && docResponsesRef.current) {
```
jkaster, line 124: what is `maybe` for here?

packages/api-explorer/src/utils/hooks/navigation.ts
```diff
@@ -73,17 +73,35 @@ export const useNavigation = () => {
...
+      additionalParams += `${key}=${value}`
+    })
+
+    let queryString = ''
+    if (globalParams) {
+      queryString = globalParams
+    }
+
+    if (additionalParams) {
+      queryString += globalParams ? `&${additionalParams}` : additionalParams
```
jkaster, line 101: looks globalParams is being added twice here

packages/code-editor/src/Markdown/common.tsx
```diff
@@ -69,6 +69,7 @@ export const MDListItem = styled.li`
   color: ${({ theme }) => theme.colors.text5};
   max-width: 600px;
   margin-bottom: 4px;
+  line-height: 1.5;
```
jkaster, line 72: Fixes scrunching of list items when they have inline code snippets

jkaster (changes requested):
Duplication of query params

packages/api-explorer/src/utils/hooks/navigation.ts
josephaxisa: `globalParams` is a condition here, it's not being assigned

packages/api-explorer/src/scenes/ErrorDetailScene/ErrorDetailScene.tsx
josephaxisa: agreed.

packages/api-explorer/src/scenes/MethodScene/MethodScene.tsx
josephaxisa: I think it can be removed. I'll address.

packages/api-explorer/src/utils/hooks/navigation.ts
jkaster: thanks. misread.

jkaster (approved):
LGTM

PR #1168, opened 2022-08-29
fix: rerendering issues on DiffScene options change

patnir41:
In DiffScene, upon selecting new comparison options, scene will accurately render new diff methods and content and total page count.


josephaxisa (approved):
lgtm

PR #1170, opened 2022-08-30
fix: add error detail handling for Python SDK

praphun:
* fix response handling when response payload is empty (e.g. 204)
* add error detail handling for Python SDK


jkaster (approved):
LGTM

PR #1174, opened 2022-09-19
chore: disable python integ tests on 22.12, 22.14 and 22.16

praphun:
disable python integ tests on 22.12, 22.14 and 22.16

### DIFF
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -112,6 +112,12 @@ jobs:
         os:
           - ubuntu
         looker: ${{ fromJson(needs.setup.outputs.matrix_json) }}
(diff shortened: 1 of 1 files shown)


praphun:
I've already changed the list of supported version in [reusable-actions](https://github.com/looker-open-source/reusable-actions/blob/main/.github/workflows/supported-versions.yml#L50) to include 22.12, 22.14, 22.16.

Without updating this exclude list for new versions added to supported-versions, Python CI workflow would fail. I hope for this to be a forcing mechanism to make us run the tests locally and consciously add more exclusions. The friction would also be a forcing mechanism to make us investigate and solve this properly.

jkaster (approved):
LGTM

PR #1190, opened 2022-10-14
fix: various path navigation issues

jkaster:
- sdk selection was broken on the developer portal, because some scenes weren't using the react dom `location` for `location.pathname`, which caused the path to include `/api/explorer` when it shouldn't have
- This also fixes a crash bug when clicking on a method summary page from a type summary page
- also introduces **funSDK** syntax into the TypeScript `makeTheCall()` that now shows in RunIt
- also updates yarn-deduplicate


fabio-looker (approved):
LGTM - very clean fix!

PR #1194, opened 2022-10-17
chore: simplify APIX bundle deployment

jkaster:
- added `yarn deploy:apix`


fabio-looker (approved):
Thanks for adding it to the package scripts! Pathing looks cleaner too :) LGTM

PR #1210, opened 2022-11-18
fix: Hack App Wholly Artifact Refactor

jeremytchang:
- The biggest issue, was business logic relied on Hacker's `id` field which was the Looker User Id. Wholly Artifact required Id with `TABLE_NAME:` prefixed. We needed to maintain all relationships between other tables to the User table.  Added userRecord field to the Hacker class and made it's `id` field refer to the User record's key.
  - Refactored hacker list to keep showing Looker Id instead of actual User Id. SUPER HACKY. Needs to be fixed later. Hack is commented in the code.
    - On getHacker() all hackers are loaded and initialized (with their associated user record).
    - If User record doesn't exist will be created/saved. So first call to getHacker after multiple users are added to looker instance will take a long while saving all the new user records. Possibly setup admin add looker user functionality to also create user records in the future.
    - Moved registerUser logic from sheetData to sheets_client
- Updated most references from whollysheet to wholly artifact. However kept some wholly sheet references so tests can past lint/build. 
- Transitioned from Tech Ids being shown in the UI, to technology descriptions. (Project's `technologies` field should be  `tech_ids`, and Project's `$techs` should be `$tech_names`, anddd Technology's `description` should be `name` or we should add a `name` field) <-Refactor for another time. Project class also now have technology names so that they're displayed correctly in the table list. 
- Current initial load of data is around 10 round trips or so. Super slow. Need to update to load all data in one roundtrip and then process the data into rows and put them in the models/Whollyartifacts. 
- Included a python notebook to help migrate data from sheets to artifact api.


packages/hackathon/src/data/sheets_client.ts
```diff
@@ -454,22 +469,9 @@ class SheetsClient {
...
-    }
-
-    const transport = new ExtensionProxyTransport(extSDK, options)
-    const gSession = new GAuthSession(extSDK, options, transport)
-    const sheetSDK = new SheetSDK(gSession, sheetId)
-    const doc = await sheetSDK.index()
-    this.sheetData = initActiveSheet(sheetSDK, doc)
+    const sheetData = initActiveSheet(new SheetData())
+    await sheetData.init()
+    this.sheetData = sheetData
```
jkaster, line 474: I like the cleanup

packages/hackathon/src/models/Projects.ts
```diff
@@ -258,7 +281,7 @@ export class Project extends SheetRow<Project> {
   }
 
   async addJudge(hacker: Hacker) {
-    if (!hacker.canJudge) throw new SheetError(`${hacker.name} is not a judge`)
+    if (!hacker.canJudge) throw new Error(`${hacker.name} is not a judge`)
```
jkaster, line 284: We should probably create an `ArtifactError` class or `HackathonError` class or something and use that instead.

packages/hackathon/src/models/SheetData.spec
```diff
@@ -82,19 +82,6 @@ describe('SheetData', () => {
...
-      if (hackathon) {
-        const mockUser = mockAHacker('1')
-        const actual = await data.registerUser(hackathon, mockUser)
-        expect(actual._user_id).toEqual(mockUser.id)
-        expect(actual.hackathon_id).toEqual(hackathon._id)
-        expect(actual._updated).not.toEqual(noDate)
-        expect(actual.date_registered).not.toEqual(noDate)
-        expect(actual.attended).toEqual(true)
-      }
-    })
```
jkaster, line 97: Why did you have to remove this? Do we have a different method for registering users in the UI now?

packages/hackathon/src/models/SheetRow.ts
```diff
@@ -57,4 +57,8 @@ export class SheetRow<T extends ISheetRow> extends RowModel<T> {
       user.permissions.has('update')
     )
   }
+
+  namespace() {
+    return 'hackathon'
+  }
```
jkaster, line 63: nice. this is the thing I was talking about yesterday (?). We should rename refactor to HackathonData and HackathonRow in the future.

jkaster (changes requested):
I'd like to have the `.skip` implemented for the Judging spec file and the rename reverted. The rest could wait for another PR.

packages/hackathon/src/data/sheets_client.ts
```diff
@@ -259,17 +255,24 @@ class SheetsClient {
...
+      const data = await this.getSheetData()
+
+      await this.loadHackers(data)
+
+      const hacker: Hacker | undefined = this.hackers?.rows.find(
+        (h) => h.user.id === lookerUser.id
+      )
+
+      if (!hacker) {
+        throw new Error('Failed to load hacker')
```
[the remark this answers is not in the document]
jeremytchang: Resolved.

packages/hackathon/src/data/sheets_client.ts
```diff
@@ -366,11 +379,13 @@ class SheetsClient {
...
   getHackersHeadings(): HackersHeadings {
-    const template = new Hacker()
-    const lookerSdk = getCore40SDK()
-    const hackersContainer = new Hackers(lookerSdk)
-    const headers = hackersContainer.displayHeaders
-    return sheetHeader(headers, template)
+    if (!this.hackers) {
+      throw new Error('Cannot get hackers heading. Hackers not loaded')
+    }
+    if (!this.hacker) {
```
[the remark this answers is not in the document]
jeremytchang: The messages are different, since it was hard to tell, i've update the messaging.

packages/hackathon/src/models/Projects.ts
jeremytchang: Yeah a TODO for the future. It needs to be made in wholly artifact package like SheetError is in whollySheet.

packages/hackathon/src/models/SheetData.spec
jeremytchang: Its moved to sheets_client. Separation of concerns. The user registering flow was oddly split between sheets_client and sheetdata. Now it's consolidated. Noted this in the PR description.

jkaster: thanks for the explanation

jkaster (dismissed):
LGTM. This PR works, so we should get it merged and move on.

packages/hackathon/README.md
```diff
@@ -1,83 +1,12 @@
...
-      "role_groups"
-      ]
-    oauth2_urls: []
-    scoped_user_attributes: ["sheet_id", "token_server_url"]
-  }
-}
-```
-
-**Note** that http://localhost:8081/\* points to the access token server. Change to the access token server URL you are using.
+Until the Hackathon extension is available via the Looker Marketplace, a Hackathon manifest needs to be used. [Example Hackathon manifest]()
```
jkaster, line 9: I think I found the missing link!

packages/hackathon/manifest.lkml
```diff
@@ -0,0 +1,58 @@
...
+      "https://*.github.com/*",
+      "https://*.youtube.com/*",
+      "https://*.youtu.be/*",
+      "https://*.vimeo.com/*"
+    ]
+    use_form_submit: yes
+    use_embeds: no
+    use_iframes: no
+    use_clipboard: no
+    external_api_urls: ["http://localhost:8081/*"]
```
jkaster, line 24: I think this is only needed for the auth token server, which we don't need any more

jkaster (changes requested):
README update is missing a link

jkaster (approved):
:shipit:

PR #1211, opened 2022-11-24
chore: support new cloud redirect host

jkaster:
ErrorDoc implementations support both docs.looker.com and cloud.google.com redirect host patterns

### DIFF
--- a/kotlin/src/main/com/looker/rtl/ErrorDoc.kt
+++ b/kotlin/src/main/com/looker/rtl/ErrorDoc.kt
@@ -116,7 +116,7 @@ class ErrorDoc(val sdk: APIMethods, val cdnUrl: String = ErrorCodesUrl): IErrorD
 
         /** API error document_url link pattern */
         private const val ErrorDocPatternExpression =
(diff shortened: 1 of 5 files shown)


kotlin/src/main/com/looker/rtl/ErrorDoc.kt
```diff
@@ -116,7 +116,7 @@ class ErrorDoc(val sdk: APIMethods, val cdnUrl: String = ErrorCodesUrl): IErrorD
 
         /** API error document_url link pattern */
         private const val ErrorDocPatternExpression =
-            """(?<redirector>https:\/\/docs\.looker\.com\/r\/err\/)(?<apiVersion>.*)\/(?<statusCode>\d{3})(?<apiPath>.*)"""
+            """(?<redirector>https://docs\.looker\.com/r/err/|https://cloud\.google\.com/looker/docs/r/err/)(?<apiVersion>.*)/(?<statusCode>\d{3})(?<apiPath>.*)"""
```
praphun, line 119: Do we no longer need to escape here `\/`?

praphun (approved):
LGTM

kotlin/src/main/com/looker/rtl/ErrorDoc.kt
jkaster: tests work, IDEA complained about the escaping. Maybe it's a Kotlin thing.

PR #1212, opened 2022-11-26
fix: WhollyArtifact/Hack App table name fix

jeremytchang:
WhollyArtifact's RowModel derives it's tablename from the name of the constructor `constructor.name`. Turns out a constructor's name is generated by minified code, so that's a no go. Found out about this while migrating our hack.looker.com hack app, since it requires bundled/minified js. 

- Updated WhollyArtifact package to remove Rowmodel's tablename implementation
- Updated hack app to specify tablename for each RowModel.
- Also updated artifact notebook for convenience.

### DIFF
--- a/packages/hackathon/artifact.ipynb
+++ b/packages/hackathon/artifact.ipynb
@@ -9,7 +9,9 @@
     "DATA = {\n",
     "    'client_id':'',\n",
     "    'client_secret':''\n",
-    "}"
+    "}\n",
+    "\n",
+    "BASE_URL = 'http://localhost:19999'"
    ]
   },
   {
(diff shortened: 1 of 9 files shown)


packages/hackathon/artifact.ipynb
```diff
@@ -21,18 +23,18 @@
     "# Login\n",
     "import requests \n",
     "\n",
-    "response = requests.post('http://localhost:19999/login',data=DATA)\n",
-    "\n",
+    "response = requests.post(BASE_URL + '/api/4.0/login',data=DATA)\n",
```
jkaster, line 26: This would have been a lot simpler with the Python SDK 😛

jkaster (approved):
Pluralization will get you every time.

PR #1213, opened 2022-11-28
fix: Multiple hack app changes

jeremytchang:
- Updated Judging fields to new fields. 
- Updated judging form to user Looker components form and fields and also update descriptions/details of each field to be more descriptive so judges have less questions. Setup separate tab with project view, instead of a preview, which isn't useful. 
- Fixed join/leave project logic to reference hacker's id only. 
- Updated project/judging list to show modal on row click for convenience.
- Updated artifact notebook to make use of looker sdk, pull judging results, and turn artifact tables into CSV for convenience.


packages/hackathon/artifact.ipynb
```diff
@@ -2,56 +2,119 @@
...
+    "def getRows(tableName):\n",
+    "    artifacts = sdk.search_artifacts(key=tableName+'%', namespace='Hackathon')\n",
+    "    rows = []\n",
+    "    for a in artifacts:\n",
+    "        row = json.loads(a.value)\n",
+    "        row['key'] = a.key\n",
+    "        rows.append(row)\n",
+    "    return rows\n",
+    "\n",
+    "def toCSV(rows, fileName):\n",
```
jkaster, line 45: Is this a temporary utility? By adding more Python functionality we're not providing functionality that could be in the Hackathon app admin section.

packages/hackathon/src/scenes/ProjectsScene/components/ProjectView.tsx
```diff
@@ -32,25 +32,35 @@ import type { IProjectProps } from '../../../models'
...
 
-export const ProjectView: FC<ProjectViewProps> = ({ project }) => {
+export const ProjectView: FC<ProjectViewProps> = ({
+  description,
+  technologies,
+  members,
+  title,
+  project_type,
+  contestant,
+}) => {
```
jkaster, line 52: What prompted using individual properties rather than the project object? This makes any future changes we make to project more fragile, requiring updates in multiple places. What is the benefit of breaking the encapsulation?

packages/hackathon/src/scenes/JudgingEditorScene/components/JudgingForm.tsx
```diff
@@ -71,111 +68,102 @@ export const JudgingForm: FC<JudgingFormProps> = ({ judging, readonly }) => {
...
+      </Tab2>
+      <Tab2 id="project" label="Project details">
+        <ProjectView
+          title={judging.$title}
+          description={judging.$description}
+          technologies={judging.$technologies}
+          members={judging.$members}
+          project_type={judging.$project_type}
+          contestant={judging.$contestant}
+        />
```
jkaster, line 165: This could have passed a composed `project` object compatible with the `Project` interface without having to change `ProjectView`, I think.

packages/hackathon/src/scenes/ProjectsScene/components/ProjectView.tsx
jkaster: I see where you made this change for judging. See my comment there.

packages/hackathon/src/scenes/JudgingEditorScene/components/JudgingForm.tsx
```diff
@@ -71,111 +68,102 @@ export const JudgingForm: FC<JudgingFormProps> = ({ judging, readonly }) => {
...
-                  Return to judging
-                </ButtonOutline>
-                <Button type="submit" onClick={handleSave} disabled={readonly}>
-                  Save judging
-                </Button>
-              </Space>
+    <Tabs2 defaultTabId="form">
+      <Tab2 id="form" label="Judging form">
+        {readonly && <Span>You cannot judge this project</Span>}
+        <Form width="70vh">
```
jkaster, line 86: Nice you get to use Form! It didn't exist yet when we wrote this.

packages/hackathon/artifact.ipynb
jkaster: Faster for you, perhaps. However, it's easy enough to replace later.

jkaster (approved):
If this is blocking additional work, got ahead and submit it and we can address my comments after Hackathon completes.

packages/hackathon/src/scenes/JudgingEditorScene/components/JudgingForm.tsx
jeremytchang: Yeah the most comprehensive change would be reverting this change and removing Judging's generated fields. It's duplicated info that we have to worry about syncing (i'm not even sure if they're kept up to sync). Our app already contains projects/team members/etc, we should be exposing that info through redux/saga to react, and not duplicating it for convenience.

packages/hackathon/artifact.ipynb
jeremytchang: Also a lot of this functionality will be hopefully explores/dashboards. Shouldn't need to be in our hack app.

PR #1214, opened 2022-12-02
fix: Multiple hack app fixes and changes

jeremytchang:
- Fixed race condition where project members was being simultaneously create/deleted and fetched. This lead to member count in frontend being incorrect. Updated loading spinner logic to correctly stop spinning after project members have been created/deleted and then fetched. 
- Updated resources page to use multi select and also filter on both type and domain for nicer/sleeker UX.
- Updated resources page's card's font to be smaller and also added one more column.
- Update project/judging list to have UI that points out the options/actions. 
- Updated the message bar to always be floatin on top, instead of at top of page. Ltos of times when you do an actions you dont see the result message because you have to scroll up. This makes a confusing UX, so always have the message bar float at the top of the page. 
- Updated project preview display to say "team members" instead of "by".
- Fix the bug where the technologies field would show lock icon when project is locked.

### DIFF
--- a/packages/hackathon/src/data/sheets_client.ts
+++ b/packages/hackathon/src/data/sheets_client.ts
@@ -25,7 +25,6 @@
  */
 import omit from 'lodash/omit'
 import type { ValidationMessages } from '@looker/components'
-import type { ITabTable } from '@looker/wholly-artifact'
 import { getCore40SDK } from '@looker/extension-sdk-react'
 import { initActiveSheet, SheetData } from '../models/SheetData'
 import type {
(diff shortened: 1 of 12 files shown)


packages/hackathon/src/data/sheets_client.ts
```diff
@@ -80,14 +79,17 @@ class SheetsClient {
 
   async getCurrentProjects(hackathonId?: string): Promise<IProjectProps[]> {
     const data = await this.getSheetData()
+    // Need to refresh teamMembers for latest team members
+    await data.teamMembers.refresh()
     await data.projects.refresh()
     const hackathon = await this.getSheetHackathon(hackathonId)
     const rows = data.projects.filterBy(hackathon)
-    // Create a projects object from the filtered rows
+    // Create a projects object fom the filtered rows
```
jkaster, line 87: lol. introduced a typo here, but not a blocker 😄

jkaster (approved):
LGTM. Nice cleanup!

PR #1218, opened 2022-12-06
fix: Fix admin add user to correctly generate API3 keys.

jeremytchang:
Turns out API will return empty array instead of null/undefined defined in the type.


jkaster (approved):
LGTM

PR #1219, opened 2022-12-06
fix: Hack app fixes

jeremytchang:
- Updated notebook to add users by email to group and also add api3 creds to user in a group
- Updated hack app to display users in alphabetical first name order with attended first
- Updated add user saga to properly check if api3 keys exist.

### DIFF
--- a/packages/hackathon/src/data/sheets_client.ts
+++ b/packages/hackathon/src/data/sheets_client.ts
@@ -282,6 +282,26 @@ class SheetsClient {
     return this.hacker
   }
 
+  firstNameSort(a: IHackerProps, b: IHackerProps): number {
+    if (a.firstName < b.firstName) {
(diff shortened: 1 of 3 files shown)


packages/hackathon/src/data/sheets_client.ts
```diff
@@ -282,6 +282,26 @@ class SheetsClient {
     return this.hacker
   }
 
+  firstNameSort(a: IHackerProps, b: IHackerProps): number {
```
jkaster, line 285: fwiw this could use localeCompare https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

jkaster (approved):
LGTM. Code review comment is not a blocker.

PR #1222, opened 2022-12-08
fix: oauth timeout

bryans99:
Fixes an issue with OAUTH calls whereby the call initially fail because the chatty timeout (30 seconds) is exceeded.

The fix is to rely on the new AbortController signal feature added to chatty. When a signal is provided, chatty does not timeout and instead relies on signal abort to cancel the call. For OAUTH a signal is provided which bypasses the timeout. The abort is never called but this does not matter as the OAUTH processing monitors the OAUTH child window and knows when a user has terminated the flow in a manner other than a successful login.


jkaster (dismissed):
LGTM

jkaster (approved):
LGTM

PR #1223, opened 2022-12-08
feat: add request time tracking

jkaster:
The TypeScript SDKs have two new properties for tracking the "raw" request time:

- `requestStarted` timestamp right before the request begins
- `responseCompleted` timestamp right after the response is downloaded, and before the payload is parsed (in most cases)

API Explorer will also display the elapsed time in seconds for API requests on the RunIt response tab now


bryans99 (approved):
LGTM

PR #1238, opened 2023-01-24
chore: update release please

jkaster:
Our old release please version is now blocked because it was node 12. This updates it to the new version.


jeremytchang (approved):
LGTM. We should stick to versions and not specific commit hash in the future.

jkaster:
I think we had to use this specific commit at some point. Hopefully that's no longer the case.

jeremytchang:
Yeah iirc, Joel made a change to release please and pointed to it. Hopefully updating to new version doesn't break anything.

PR #1245, opened 2023-02-06
feat: keep all body parameters toggle in API Explorer requests

jkaster:
Adds a checkbox to the request form to keep all body parameters for an API request. By default, all "empty" parameters are stripped from the body for an API request


jkaster:
without any code or config changes to this branch, `bin/rebuild` now works without error. Now the actual test failures can be investigated

josephaxisa (approved):

PR #1246, opened 2023-02-09
chore: Refactor release-please config for versioning/release process changes

jeremytchang:
Updated release-please config to make use of latest features and improve our versioning/release process.

Requirements:
- Calendar versioned packages:
  - Must follow Looker’s major and minor calendar versioning of the API spec they’ve pulled in, even if the package’s changes are only dependencies with version bumps.  E.g. if Looker API spec version is at 23.12.x, then sdk is at 23.12.x, AND extension-sdk must be at 23.12.x even if it has no changes outside of its sdk dependency version bump.
    -  We make use of release-please manifest plugin “[linked-versions](https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md#linked-versions)” to sync all calendar versioned packages versions. Now instead of editing the config file, the sdk gen commit should have [“Release-As:X.X.X” in the commit body to force calendar versioning](https://github.com/googleapis/release-please#how-do-i-change-the-version-number). As long as the commit touches at least one file in at least one calendar versioned package it will force all calendar version packages to that version. There is a caveat to using linked-version. If we patch bump any of the linked packages, they will all match the highest version, however this is not too big od a deal, as it is the major/minor that matters, and the linked-version plugin saves us complexity in our release process and config file.
  - Update patch number for any changes not associated with a Looker release.
    - We use package scoped `"versioning": "always-bump-patch"` to keep any changes from turning into minor or major bumps. Now for any sdk changes in between Looker releases, you only need to make/merge the changes with a conventional commit title, and merge the resulting release PR.

- Semantically versioned packages:
  - Release-please naturally does semver on packages by default.
  - Should bump patch number if its dependency has a version bump.
    - Node workspace plugin automatically patch bumps any package if its dependencies are updated.

 Outside of all that, also cleaned up the extra `"release-as"`'s since we dont need them anymore.


jkaster (commented):
bin/sdk_gen writes a `release-as` top level key to the release please manifest. If that's not the appropriate place any more we'll need to update that too. Does this not need an update to our release-please version somewhere else? I was expecting more changes than just this one file, but I could be missing things.

bin/sdk_gen
```diff
@@ -71,31 +71,6 @@ const run = (command) => {
   return result
 }
 
-/**
- * Updates release-please-config.json to stamp the correct CalVer for the SDKs
- *
- * if `release-as` is already set, CalVer patch will be bumped if the current
```
jeremytchang, line 77: @jkaster I've removed this function becase we don't need to set release-as anymore with my changes. However i'm confused by this statement.`if `release-as` is already set, CalVer patch will be bumped if the current version is the same as the requested release version`. When will this happen? Afaict, when we are generating new bindings, the repo should not be on the new calendar version.

bin/sdk_gen
```diff
@@ -214,11 +189,10 @@ const regen = async (release) => {
     const branch = branchActive()
     if (await pullci(release)) {
       return batch([
-        (release) => releasePlease(release),
         'yarn gen',
         'bin/smoke typescript python kotlin',
         'git add -A',
-        `git commit -m "feat: generate SDKs for Looker ${release}"`,
+        `git commit -m "feat: generate SDKs for Looker ${release}" -m "Release-As: ${release}.0`,
```
jeremytchang, line 195: @jkaster This additional body will tell release-please to calendar version appropriately.

jeremytchang:
I missed the `sdk_gen` file because it is not part of our CI/CD manual. Updated it. Do we want sdk_codegen to be the standard way of generating new bindings? (If so we need to update the CI/CD manual)

bin/sdk_gen
jkaster: Sometimes we release, e.g., 23.0.0 then later 23.0.1. The `sdk_gen` script uses the current main `release-as` value to bump the patch version if `release-as` is not blank.

bin/sdk_gen
jkaster: Thanks for the reminder. You did tell me that before.

jkaster:
Yes, `sdk_gen` is the shortest path we currently have to generating a new SDK release, so it should be part of our process until we have something better. I'd intended to update the CI/CD manual but got bogged down with other issues.

bin/sdk_gen
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
 /**
- * Create sdk_version branch if on main,
+ * Creagit te sdk_version branch if on main,
```
jkaster, line 4: typos

jkaster (commented):
Got some typos in a comment. The code looks good, but I'll wait until the typos are fixed to approve

bin/sdk_gen
jkaster: TIL you can use `-m` twice in a git commit

jkaster (approved):
Looks good if `-m` can be used twice in a git commit command

PR #1247, opened 2023-02-13
chore: Update sdk Readme and kickoff Release 23.0.0

jeremytchang:
This is a change to the sdk readme. Since we link the versions of all our calendar versioned packages, this change plus the below "Release-As" git commit body footer will tell release-please to set all calendar versions packages to 23.0.0.

Release-As: 23.0.0


pfrisella (approved):
LGTM

PR #1254, opened 2023-02-27
chore: embed-services package

josephaxisa:
This is just a skeleton of the package that will contain the service layer for looker embed components.

### DIFF
--- a/packages/embed-services/package.json
+++ b/packages/embed-services/package.json
@@ -1,0 +1,45 @@
+{
+  "name": "@looker/embed-services",
+  "version": "0.0.1-alpha",
+  "description": "Looker Embed Services",
+  "main": "lib/index.js",
+  "module": "lib/esm/index.js",
+  "sideEffects": false,
+  "typings": "lib/index.d.ts",
+  "files": [
+    "lib"
+  ],
+  "private": false,
+  "publishConfig": {
+    "access": "private"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/looker-open-source/sdk-codegen.git",
+    "directory": "packages/embed-services"
+  },
+  "author": "Looker",
+  "license": "MIT",
+  "scripts": {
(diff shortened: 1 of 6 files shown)


packages/embed-services/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "main": "lib/index.js",
+  "module": "lib/esm/index.js",
+  "sideEffects": false,
+  "typings": "lib/index.d.ts",
+  "files": [
+    "lib"
+  ],
+  "private": false,
+  "publishConfig": {
+    "access": "public"
```
jkaster, line 14: I think you want this private for now

packages/embed-services/src/index.ts
```diff
@@ -0,0 +1,25 @@
+/*
+
+ MIT License
+
+ Copyright (c) 2022 Looker Data Sciences, Inc.
```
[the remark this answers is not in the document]
jkaster: Yes, just let the linter do its thing. It's based on the `LICENSE` from the main folder, which should be updated to 2023, then your local `LICENSE` file matches.

jkaster (changes requested):
You should update the main repo LICENSE file for the current copyright year

packages/embed-services/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "version": "0.0.1-alpha",
+  "description": "Looker Embed Services",
+  "main": "lib/index.js",
+  "module": "lib/esm/index.js",
+  "sideEffects": false,
+  "typings": "lib/index.d.ts",
+  "files": [
+    "lib"
+  ],
+  "private": false,
```
jkaster, line 12: should this be `true` or `false`?

jkaster (commented):
got a question before approval

packages/embed-services/package.json
josephaxisa: good catch, thanks.

jkaster (approved):

PR #1259, opened 2023-02-28
chore: fix deprecations

drstrangelooker:
Fix deprecations in github actions


jeremytchang (approved):
Nice! Thanks for adding the paths too LGTM

PR #1261, opened 2023-03-03
chore: Added workflows to label new/reopened issues

jeremytchang:
New issues get "p3" and "need triage". Reopened issues get "p3". Went with simpler approach of 2 workflows instead of figuring out how to do branching logic in a workflow.


drstrangelooker (approved):
LGTM

PR #1275, opened 2023-03-15
Set User-Agent header from ApiSettings AgentTag

### LINKED ISSUE #1274: Go SDK does not include User-Agent header
According to the [cookieless embedding documentation](https://cloud.google.com/looker/docs/cookieless-embed), the user agent of the calling user's browser has to match the user agent that is on the session. However, the Go SDK does _not_ pass a User-Agent Header on requests currently, so calling `AcquireEmbedCookielessSession` in the SDK errors.

The User-Agent Header should be set on AuthSession requests in a way that we can pass a browser user agent to it.

cgooding-va:
For cookieless embedding, the User-Agent Header on the request to acquire the session must match the embedded user's browser agent, which is not possible currently as there is no way to define that header.

This change allows a User-Agent header to be set when making requests with the AuthSession service.

Fixes #1274


jeremytchang:
Thanks for the PR. I've copied your commit/changes over to another PR so our CI tests will run correctly. https://github.com/looker-open-source/sdk-codegen/pull/1285 Closing this PR.

PR #1280, opened 2023-03-18
feat: embed playground

josephaxisa:
This PR delivers the embed playground harness, particularly to make it easier to test embed components. Follow up work which I will create the tickets for will include:
1. Refactor `OAuthConfigForm` to use `@looker/redux`
2. Replace the `ConfigForm` in `@looker/run-it` with the one in `@looker/extension-utils`
3. Correct the order of help messages in `OAuthConfigForm` 
4. Add adequate test coverage for `OAuthConfigForm` and its utiliies


packages/extension-utils/src/OAuthConfigForm/OAuthConfigForm.spec.tsx
```diff
@@ -0,0 +1,241 @@
...
+    }),
+    useHistory: jest
+      .fn()
+      .mockReturnValue({ push: jest.fn(), location: globalThis.location }),
+  }
+})
+
+const ConfigKey = 'ConfigKey'
+
+const initSdk = () => {
```
jkaster, line 59: You probably already have a test utils section planned where this could be shared, but we should in the future

packages/extension-utils/src/OAuthConfigForm/OAuthConfigForm.tsx
```diff
@@ -0,0 +1,400 @@
...
+    updateMessage(CRITICAL, message)
+  }
+
+  const saveConfig = (baseUrl: string, webUrl: string) => {
+    const data = {
+      base_url: baseUrl,
+      looker_url: webUrl,
+      client_id: clientId,
+      redirect_uri,
+    }
```
jkaster, line 176: we should create an interface for the configuration values we store. not required now, but will be much more useful for OAuth sharing across playgrounds

packages/extension-utils/src/OAuthConfigForm/OAuthConfigForm.tsx
```diff
@@ -0,0 +1,400 @@
...
+    updateFields({
+      [BASE_URL]: baseUrl,
+      [WEB_URL]: webUrl,
+    })
+    // TODO: replace when redux is introduced
```
jkaster, line 181: Even when we have redux, I think we'd still want to save OAuth configuration to local storage as well, so it can be used for the restart source

jkaster (approved):
Worked for me when I ran it. When I first entered the configuration, Login failed after verifying the config, but I restarted the server and got **Hello world** after that.

It would be handy to have this in the root package.json, or something similar

```json
    "play:ep": "yarn workspace @looker/embed-playground play",

```

packages/extension-utils/src/OAuthConfigForm/utils.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+  return versions
+}
+
+/**
+ * Validates URL and standardizes it
+ * @param url to validate
+ * @returns the standardized url.origin if it's valid, or an empty string if it's not
+ */
+export const validateUrl = (url: string): string => {
```
zeckertG, line 67: Is this referenced anywhere else in the code? Is there a url that you should be validating?

packages/extension-utils/src/OAuthConfigForm/utils.ts
```diff
@@ -0,0 +1,75 @@
...
+  try {
+    const result = new URL(url)
+    if (url.endsWith(':')) return url
```
zeckertG, line 70: What is the significance in checking if url ending is ending with a ':'?

zeckertG (commented):
Looks good just a couple of questions

packages/embed-playground/src/components/Loader/Loader.tsx
```diff
@@ -0,0 +1,54 @@
...
+  ProgressCircular,
+} from '@looker/components'
+import type { ThemeOverrides } from '@looker/extension-utils'
+
+export interface LoaderProps {
+  themeOverrides: ThemeOverrides
+  message: string
+}
+
+export const Loader = ({ themeOverrides, message }: LoaderProps) => (
```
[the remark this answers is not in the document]
josephaxisa: yeah, potentially. We can move things as we start needing them.

packages/extension-utils/src/OAuthConfigForm/OAuthConfigForm.spec.tsx
josephaxisa: yeah. Also `RunIt` also has its own `initRunItSdk`. No reason why we shouldn't use this for all dual mode extensions. It will require a bit more configurability.

packages/extension-utils/src/OAuthConfigForm/utils.ts
josephaxisa: it's used by the config form

packages/extension-utils/src/OAuthConfigForm/OAuthConfigForm.tsx
jkaster: Good deal!

packages/extension-utils/src/OAuthConfigForm/utils.ts
josephaxisa: you know what, I don't have a good answer for you. The reason is that these utilities were copied over from @looker/run-it, as well as a lot of the code for the `OAuthConfigForm`. Unfortunately we were lacking a lot of unit tests there. Another ticket to add to the list is for us to add unit tests to all of this.

jkaster: That's there because it was required to handle interactive url editing correctly

PR #1282, opened 2023-03-23
Fix: Oauth message on failed verify

zeckertG:
This PR fixes the OauthConfigForm's help message. This work was highlighted in https://github.com/looker-open-source/sdk-codegen/pull/1280

### DIFF
--- a/packages/extension-utils/src/OAuthConfigForm/OAuthConfigForm.tsx
+++ b/packages/extension-utils/src/OAuthConfigForm/OAuthConfigForm.tsx
@@ -317,35 +317,33 @@ export const OAuthConfigForm = ({
               />
             </Fieldset>
           </Form>
-          {fields.fetchedUrl &&
-            fields.fetchIntent === CRITICAL &&
-            fields.fetchResult !== '' && (
-              <>
-                <Paragraph fontSize="small">
-                  On {fields.fetchedUrl}, enable {clientLabel} as a{' '}
-                  <Link
-                    href="https://github.com/looker-open-source/sdk-codegen/blob/main/docs/cors.md#reference-implementation"
-                    target="_blank"
-                  >
-                    Looker OAuth client
-                  </Link>{' '}
-                  by adding "{(window as any).location.origin}" to the{' '}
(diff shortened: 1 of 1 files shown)


zeckertG:
> lgtm. Would it be possible to write a unit test for this change? If it's too much trouble with the current setup, it can wait for when you start working on the refactor.

I have a PR in the works that has some general refactoring before we make the move to redux. I will include some test coverage in that PR

josephaxisa (approved):

packages/extension-utils/src/OAuthConfigForm/OAuthConfigForm.tsx
```diff
@@ -312,33 +317,35 @@ export const OAuthConfigForm = ({
...
-                  register this API Explorer instance
-                </Link>{' '}
-                as an OAuth client.
-              </Paragraph>
-              <CodeCopy key="appConfig" language="json" code={appConfig} />
-            </>
-          )}
+          {fields.fetchedUrl &&
+            fields.fetchIntent === CRITICAL &&
+            fields.fetchResult !== '' && (
```
jkaster, line 322: We shouldn't be making these configuration steps harder to see. Showing the configuration setup is a feature, not a bug, even when a specific user might have a successful oauth config verification. We don't have this information clearly documented anywhere else in all of Looker.

jkaster (changes requested):
I'd like the configuration information to always be visible and update whenever the entered host name is updated.

packages/extension-utils/src/OAuthConfigForm/OAuthConfigForm.tsx
zeckertG: These steps are showing only when the verification / save is successful. Now it will always show if there are any issues. You are saying you would prefer the help message to update as the user types into the URL box? Even if URL is invalid?

jkaster:
I would want it to show with success or failure and only with a valid URL.

jkaster (approved):
Thanks for updating

josephaxisa (approved):
@zeckertG confirmed that unit tests are going to be in a following PR as it's too much trouble to mock the state without redux right now.

PR #1286, opened 2023-03-28
Chore: Cleanup Oauth Form state variables before moving to redux

zeckertG:
Updating the variables that control the state for the OauthConfigForm. If there are any naming changes that look off, now would be the best time to edit those.


packages/extension-utils/src/OAuthConfigForm/OAuthConfigForm.tsx
```diff
@@ -119,54 +96,32 @@ export const OAuthConfigForm = ({
...
-    setSaved(data)
-    updateFields({
-      [BASE_URL]: base_url,
-      [WEB_URL]: looker_url,
-      [FETCH_INTENT]:
-        base_url !== '' && looker_url !== '' ? POSITIVE : CRITICAL,
-    })
-  }, [])
-
+  const [apiServerUrlValue, setApiServerUrlValue] = useState('')
```
josephaxisa, line 99: could these be `apiServerUrl` and `setApiServerUrl` instead?

packages/extension-utils/src/OAuthConfigForm/OAuthConfigForm.tsx
```diff
@@ -119,54 +96,32 @@ export const OAuthConfigForm = ({
...
+  const [fetchedUrl, setFetchedUrl] = useState('')
+  const [webUrlValue, setWebUrlValue] = useState('')
```
josephaxisa, line 101: ditto

packages/extension-utils/src/OAuthConfigForm/OAuthConfigForm.tsx
```diff
@@ -176,107 +131,111 @@ export const OAuthConfigForm = ({
...
+  const handleSaveClick = async () => {
+    const versions = await verifyUrl()
+    if (versions) {
+      saveConfig(versions.api_server_url, versions.web_server_url)
+    }
   }
 
-  const handleVerify = async (e: BaseSyntheticEvent) => {
-    await updateForm(e, false)
+  const handleVerifyClick = () => {
```
josephaxisa, line 167: I prefer the previous handler names, I find them more concise. `handleVerify`, `handleSave`, `handleClear`, `handleLogin` etc.

josephaxisa (commented):
Left some comments. Also added @jkaster as well to review

packages/extension-utils/src/OAuthConfigForm/OAuthConfigForm.tsx
jkaster: I also think the `Value`suffix is redundant

packages/extension-utils/src/OAuthConfigForm/OAuthConfigForm.tsx
jkaster: agreed. The verb already makes the method name clear

jkaster (changes requested):
I agree with dropping the `Value` and `Click` suffixes

packages/extension-utils/src/OAuthConfigForm/OAuthConfigForm.tsx
```diff
@@ -119,54 +96,28 @@ export const OAuthConfigForm = ({
...
+    setMessageBarText(message)
   }
 
   const isConfigured = () => {
-    return (
-      saved !== EmptyConfig &&
-      fields[BASE_URL] === saved.base_url &&
-      fields[WEB_URL] === saved.looker_url
-    )
+    return apiServerUrl === saved.base_url && webUrl === saved.looker_url
```
zeckertG, line 115: comparing the EmptyConfig always results in false, as objects are passed / compared by reference in JS.

jkaster: I'm guessing empty urls can't be saved?

jkaster (changes requested):
Please drop the `Click` suffix

jkaster (approved):
LGTM

We may want to change the names of the stored config properties in the future, but not required right now.

josephaxisa (approved):
Looks good!

PR #1295, opened 2023-04-12
feat: base class for all services

josephaxisa:


zeckertG (dismissed):

packages/embed-services/src/ItemList.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+  })
+
+  describe('find', () => {
+    it('finds by key and value', () => {
+      const actual = ItemList.find('id', 'foo')
+      expect(actual).toEqual(items[0])
+    })
+
+    it('returns undefined if item not found', () => {
+      const actual = ItemList.find('name', 'bogus')
```
jkaster, line 124: can you add a success test for find on `name`?

packages/embed-services/src/ItemList.ts
```diff
@@ -0,0 +1,114 @@
...
+  readonly expiresAt: number
+  index(key?: keyof T): ItemList<T>
+  indexedItems: Record<string, T>
+  setExpiration(): void
+  clearIfExpired(): void
+  find(key: keyof T, value: any): T | undefined
+}
+
+export interface IEntityService<T> extends IItemList<T> {
+  get(id: string, cache?: boolean, ...options: any[]): Promise<T>
```
jkaster, line 45: based on expected usage patterns, I'm wondering if the cache toggle should just be one of the options you can pass in, since we could anticipate, e.g., wanting to specify `fields` more often than cache control.

packages/embed-services/src/ItemList.ts
```diff
@@ -0,0 +1,114 @@
...
+  private keyField = 'id'
+
+  constructor(sdk: IAPIMethods, timeToLive = DEFAULT_TTL) {
+    super(sdk, timeToLive)
+  }
+
+  /**
+   * Creates an indexed collection from the cached items
+   */
+  index(key: keyof T = this.keyField) {
```
jkaster, line 71: love that you can set a default like this in TS

jkaster (commented):
This is looking good. I have some feedback before approving, though

packages/embed-services/src/ItemList.spec.ts
josephaxisa: sure

jkaster (approved):

PR #1297, opened 2023-04-12
feat: embed url service

josephaxisa:


packages/embed-services/src/EmbedUrl.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+
+      expect(lookTargetUrl.isDashboard).toBe(false)
+      expect(lookTargetUrl.isExplore).toBe(false)
+      expect(lookTargetUrl.isLook).toBe(true)
+    })
+
+    it('should throw if target url is invalid', () => {
+      expect(() => new EmbedUrl(invalidUrl)).toThrowError(
+        'Unknown content type'
+      )
```
jkaster, line 71: Nice test pattern, but I'm wondering if we should throw an error for a utility class that tells us whether or not a given URL is an embed url

packages/embed-services/src/EmbedUrl.ts
```diff
@@ -0,0 +1,134 @@
...
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+
+ */
+
+export enum ContentType {
+  Look = 'LOOK',
+  Explore = 'EXPLORE',
+  Dashboard = 'DASHBOARD',
```
jkaster, line 30: I'm still not a fan of `SHOUTING` enums, but since it's an enum our code won't care whether it's upper or lowercased (which is more polite, of course).

packages/embed-services/src/EmbedUrl.ts
```diff
@@ -0,0 +1,134 @@
...
+}
+
+const THEMABLE_CONTENT: ContentType[] = [
```
jkaster, line 33: I presume this is from a standard we use somewhere on the frontend for constant names?

jkaster (commented):
This is looking great! I don't think we should throw an error on an invalid URL though. Perhaps have an `.isValid` property if it's not a recognized embed url

packages/embed-services/src/EmbedUrl.ts
josephaxisa: hmm, it's how I am used to doing it. I believe I saw @bryans99 doing it similarly as well. I can look through a style guide to make sure though

packages/embed-services/src/EmbedUrl.spec.ts
josephaxisa: Any ideas? I could set `contentType` to `invalid` and not error

packages/embed-services/src/EmbedUrl.ts
josephaxisa: See https://google.github.io/styleguide/tsguide.html#identifiers

packages/embed-services/src/EmbedUrl.ts
josephaxisa: I'll quiet them down when addressing feedback. However, see https://google.github.io/styleguide/tsguide.html#identifiers

packages/embed-services/src/EmbedUrl.ts
jkaster: I don't think Looker follows the Google style guide for TypeScript

packages/embed-services/src/EmbedUrl.ts
jkaster: Not critical

packages/embed-services/src/EmbedUrl.spec.ts
jkaster: I like the idea of Invalid

josephaxisa: Alright. I'll do that and make `EmbedUrl.embedUrl()` throw if called

jkaster: Good idea

jkaster (approved):

PR #1298, opened 2023-04-12
feat: singleton factory and theme service

josephaxisa:


packages/embed-services/src/ThemeService.ts
```diff
@@ -0,0 +1,147 @@
...
+      this.setExpiration()
+    }
+    return item
+  }
+
+  /**
+   * Get all themes
+   * @param fields to retrieve
+   */
+  async getAll(fields?: string) {
```
jkaster, line 80: this may need to be an `options?: GetOptions` param for interface consistency as we add other services with `all` calls that have different options. Doesn't need to happen now. Still need to think about smart ways to extend that interface for code hints.

jkaster (dismissed):
I have some suggestions but nothing blocking approval

jkaster (approved):
Quick work! Looks great.

PR #1300, opened 2023-04-14
feat: SelectTheme component

josephaxisa:


josephaxisa:
Apologies for the multiple updates. Had a minor issue when committing.

packages/embed-components/src/Theme/SelectTheme.tsx
```diff
@@ -0,0 +1,74 @@
...
+  }, [themes])
+
+  const handleChange = (id: string) => {
+    selectThemeAction({ id })
+  }
+
+  return (
+    <Select
+      disabled={themes.length === 1}
+      isLoading={!initialized}
```
zeckertG, line 68: should the isLoading also be true if themes.length === 0? The retrieval of themes can also take some time.

zeckertG (dismissed):
Looks good. Just a couple of questions

packages/embed-components/src/Theme/SelectTheme.spec.tsx
```diff
@@ -0,0 +1,153 @@
...
+    fireEvent.click(anotherTheme)
+
+    await waitFor(() => {
+      expect(selectThemeActionSpy).toHaveBeenCalledWith({
+        id: customTheme1.id,
+      })
+    })
+  })
+
+  it('is disabled when only one theme is available', () => {
```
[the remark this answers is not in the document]
josephaxisa: if `loadThemeDataSaga` fails for some reason, `themes` will have its default value, an empty array. Perhaps I can render a validation message below it.

packages/embed-components/src/Theme/SelectTheme.tsx
josephaxisa: This would only be the case if somehow the API call fails as there will always be at least 1 theme (the Looker theme, which is not deletable)

packages/embed-components/src/Theme/state/sagas.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+
+  describe('loadThemeDataSaga', () => {
+    const {
+      loadThemeDataAction,
+      loadThemeDataSuccessAction,
+      setFailureAction,
+    } = themeActions
+    const lookerTheme = { id: 1, name: 'Looker' } as unknown as ITheme
+    const defaultTheme = { id: 2, name: 'custom_theme' } as unknown as ITheme
+    const anotherTheme = { id: 3, name: 'custom_theme_1' } as unknown as ITheme
```
jkaster, line 100: are the ids returned by the API actually numeric?

jkaster: Could the `as unknown` be required because the IDs are numeric? What if you use `as Partial<ITheme>`?

jkaster (commented):
Looking good! I have minor questions.

packages/embed-components/src/Theme/SelectTheme.spec.tsx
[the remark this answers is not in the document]
josephaxisa: @zeckertG after thinking about this some more I realized I misunderstood the first time round and you made a valid point. I made some changes that should address this

packages/embed-components/src/Theme/SelectTheme.tsx
josephaxisa: My latest change should address this as well :)

packages/embed-components/src/Theme/SelectTheme.spec.tsx
```diff
@@ -0,0 +1,172 @@
...
+      ...overrides,
+    },
+  })
+
+  const initActionSpy = jest.fn()
+  const loadThemeDataActionSpy = jest.fn()
+  const selectThemeActionSpy = jest.fn()
+
+  beforeEach(() => {
+    ;(useThemeActions as jest.Mock).mockReturnValue({
```
bryans99, line 62: Why not mock useStoreState as well? Will this allow you to get rid of the provider?

packages/embed-components/src/Theme/SelectTheme.spec.tsx
```diff
@@ -0,0 +1,172 @@
...
+  })
+
+  afterEach(() => {
+    jest.clearAllMocks()
+  })
+
+  it('renders with the default theme selected', async () => {
+    const mockStore = createTestStore(getMockStoreState())
+    renderWithTheme(
+      <Provider store={mockStore}>
```
bryans99, line 76: Eliminate need for this by mocking useStoreState?

packages/embed-components/src/Theme/SelectTheme.tsx
```diff
@@ -0,0 +1,78 @@
...
+  useEffect(() => {
+    if (initialized) {
+      /** If theme service is initialized, load all theme data */
+      loadThemeDataAction()
+    }
+  }, [initialized])
+
+  useEffect(() => {
+    const themeOptions: SelectOptionObject[] = []
+    themes.forEach((theme) => {
```
bryans99, line 51: You could use map rather than forEach

packages/embed-components/src/Theme/SelectTheme.tsx
```diff
@@ -0,0 +1,78 @@
...
+      themeOptions.push({
+        value: theme.id!,
+        label: theme.name,
+      })
+    })
+    themeOptions.sort((x, y) => x.label!.localeCompare(y.label!))
```
bryans99, line 57: Then you could chain the sort.

Would it make sense to the sort the themes in the saga?

packages/embed-components/src/Theme/state/sagas.ts
```diff
@@ -0,0 +1,132 @@
...
+function* loadThemeDataSaga() {
+  const { loadThemeDataSuccessAction, setFailureAction } = themeActions
+  try {
+    const service = getThemeService()
+    yield* call([service, 'load'])
+    const searchParams = new EmbedUrl().searchParams
+    const urlThemeName = searchParams.theme
+    let urlTheme: ITheme | undefined
+    if (urlThemeName) {
+      urlTheme = service.find('name', urlThemeName)
```
bryans99, line 62: No use of call here? I assume its a synchronous function

packages/embed-components/src/Theme/state/sagas.ts
```diff
@@ -0,0 +1,132 @@
...
+}
+
+export function* saga() {
+  const {
+    initAction,
+    loadThemeDataAction,
+    getThemesAction,
+    getDefaultThemeAction,
+    selectThemeAction,
+  } = themeActions
```
bryans99, line 126: The actions can be initialized outside of the function

packages/embed-components/src/Theme/state/slice.ts
```diff
@@ -0,0 +1,122 @@
...
+      state.initialized = true
+    },
+    loadThemeDataAction() {
+      // noop
+    },
+    loadThemeDataSuccessAction(
+      state,
+      action: PayloadAction<LoadThemeDataSuccessAction>
+    ) {
+      state.defaultTheme = action.payload.defaultTheme
```
bryans99, line 82: you can access the contents of the payload using the spread operator

bryans99 (commented):
Looks reasonable. 

I've made some suggestions and perhaps an experiment

packages/embed-components/src/Theme/SelectTheme.spec.tsx
josephaxisa: I'll try. I think the provider is always required.

packages/embed-components/src/GlobalStore/sagas.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+import { takeEvery, put } from 'typed-redux-saga'
+import { createFactory } from '@looker/embed-services'
+import type { PayloadAction } from '@reduxjs/toolkit'
+
+import { factoryActions } from './slice'
+import type { InitFactoryAction } from './slice'
+
+function* initSaga(action: PayloadAction<InitFactoryAction>) {
+  const { initFactorySuccessAction, setFailureAction } = factoryActions
```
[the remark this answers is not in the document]
josephaxisa: hmm, if I remember correctly I tried this when writing this and it failed because `factoryActions` got destructured before the slice was built. I'll try again.

packages/embed-components/src/Theme/SelectTheme.tsx
josephaxisa: yeah it would. I'll do that

packages/embed-components/src/Theme/state/sagas.ts
josephaxisa: It is synchronous, no use for call.

packages/embed-components/src/Theme/state/slice.ts
josephaxisa: Okay, I'll replace these with `state = {...action.payload }`

packages/embed-components/src/GlobalStore/sagas.ts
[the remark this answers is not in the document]
josephaxisa: Yep, I get this:
```
    TypeError: Cannot destructure property 'initAction' of '_slice.themeActions' as it is undefined.

      36 |
      37 | const {
    > 38 |   initAction,
         |   ^
      39 |   loadThemeDataAction,
      40 |   getThemesAction,
      41 |   getDefaultThemeAction
```

packages/embed-components/src/Theme/state/sagas.ts
```diff
@@ -0,0 +1,136 @@
...
+ * @param action containing id of theme to select
+ */
+function* selectThemeSaga(action: PayloadAction<SelectThemeAction>) {
+  const { selectThemeSuccessAction, setFailureAction } = themeActions
+  try {
+    const service = getThemeService()
+    if (service.expired()) {
+      yield* call(getThemesSaga)
+      yield* call(getDefaultThemeSaga)
+    }
```
josephaxisa, line 115: @bryans99 are you ok with me reusing the other two sagas from this saga?

packages/embed-components/src/Theme/SelectTheme.tsx
```diff
@@ -0,0 +1,77 @@
...
+  }, [themes])
+
+  const handleChange = (id: string) => {
+    selectThemeAction({ id })
+  }
+
+  return (
+    <SpaceVertical gap="xxxsmall">
+      <Select
+        disabled={themes.length === 1}
```
jkaster, line 67: we know it shouldn't ever happen, but do you want to cover the scenario when themes.length = 0?

jkaster (approved):
LGTM

zeckertG (dismissed):

packages/embed-components/src/GlobalStore/sagas.ts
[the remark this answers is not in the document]
bryans99: I'm not seeing this when I try it in GlobalStore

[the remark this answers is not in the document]
bryans99: Not seeing it in theme sagas either

packages/embed-components/src/GlobalStore/sagas.ts
```diff
@@ -0,0 +1,46 @@
...
+  try {
+    createFactory(action.payload.sdk)
+    yield* put(initFactorySuccessAction())
+  } catch (error: any) {
+    yield* put(setFailureAction({ error: error.message }))
+  }
+}
+
+export function* saga() {
+  const { initFactoryAction } = factoryActions
```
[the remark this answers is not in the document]
bryans99: I'm not seeing your issue

packages/embed-components/src/Theme/SelectTheme.spec.tsx
bryans99: Did you try? Or will that be a separate PR

packages/embed-components/src/Theme/state/sagas.ts
bryans99: Why not put the action that triggers it. This will block

josephaxisa: sure, I could do that

josephaxisa: I changed it to call the actions (not put, as I want it to block).

zeckertG (approved):
lgtm

zeckertG (approved):

jkaster (approved):
LGTM

packages/embed-components/src/Theme/state/sagas.ts
josephaxisa: I couldn't get this to work. I suspect it has to do with the version of @looker/redux that we are using here.

PR #1319, opened 2023-05-26
Add onCopy prop for QuickEmbed

zeckertG:
Adding onCopy function to QuickEmbed component and added tests for both close and copy. 

document.execCommand = jest.fn(), was added to the test file because the copying to clipboard function must be mocked.


packages/embed-components/src/QuickEmbed/QuickEmbed.tsx
```diff
@@ -44,9 +44,10 @@ import { useThemesStoreState, SelectTheme, useThemeActions } from '../Theme'
 
 interface QuickEmbedProps {
   onClose: () => void
+  onCopy: () => void
```
josephaxisa, line 47: did you mean to make this optional? Also let's document the interface.

josephaxisa (dismissed):

jkaster (dismissed):

josephaxisa (approved):
tested locally and everything works as expected.

PR #1320, opened 2023-05-31
fix: use new CDN for errorcodes repository

jkaster:
using https://static-a.cdn.looker.app instead of the old CDN

### DIFF
--- a/packages/sdk-rtl/src/errorDoc.spec.ts
+++ b/packages/sdk-rtl/src/errorDoc.spec.ts
@@ -222,10 +222,7 @@ describe('ErrorDoc', () => {
       ['login_2_404.md', 'login_2'],
       ['and_another_404.md', 'and_another'],
       ['errorcodes/live/login_404.md', 'login'],
-      [
-        'https://marketplace-api.looker.com/errorcodes/live/login_404.md',
-        'login',
-      ],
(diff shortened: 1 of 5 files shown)


packages/sdk-rtl/src/errorDoc.spec.ts
```diff
@@ -222,10 +222,7 @@ describe('ErrorDoc', () => {
       ['login_2_404.md', 'login_2'],
       ['and_another_404.md', 'and_another'],
       ['errorcodes/live/login_404.md', 'login'],
-      [
-        'https://marketplace-api.looker.com/errorcodes/live/login_404.md',
-        'login',
-      ],
+      ['https://static-a.cdn.looker.app/errorcodes/live/login_404.md', 'login'],
```
zeckertG, line 225: Just remove /live from this path though. https://static-a.cdn.looker.app/errorcodes/login_404.md works!

zeckertG (changes requested):
Looks good. Just one quick change

zeckertG (approved):

packages/sdk-rtl/src/errorDoc.spec.ts
jkaster: FWIW the `/live` path could have been anything. I was testing to make sure localized error docs would also resolve to a method name.

PR #1327, opened 2023-06-15
Zeckert/create sagas for o auth form

zeckertG:
Updating OAuth Form to use sagas


packages/extension-utils/src/OAuthForm/src/state/sagas.ts
```diff
@@ -0,0 +1,201 @@
...
+    )
+  } catch (error: any) {
+    yield* put(handleVerifyActionFailure(error.message))
+  }
+}
+
+export function* saga() {
+  const {
+    initAction,
+    handleUrlChangeAction,
```
zeckertG, line 191: Should I drop the 'handle' from these actions?

packages/extension-utils/src/OAuthForm/src/components/OAuthForm.tsx
```diff
@@ -113,88 +105,31 @@ export const OAuthForm = ({
...
-        message: `'${value}' is not a valid url`,
-        type: 'error',
-      }
-    }
-
-    updateApiServerUrlAction({
-      apiServerUrl: newApiServerUrl,
-      webUrl: '',
-      validationMessages: newValidationMessages as ValidationMessages,
+    handleUrlChangeAction({
```
zeckertG, line 130: Is this something I could do inline the JSX, or is it better to leave it out of it?

packages/extension-utils/src/OAuthForm/src/state/sagas.ts
josephaxisa: you could change them to `setUrlAction`, `verifyAction`, `saveConfigAction`, `clearConfigAction` respectively. The handlers themselves should follow the naming convention `handle*`.

packages/extension-utils/src/OAuthForm/src/components/OAuthForm.tsx
josephaxisa: you could but I'd recommend against it as it makes debugging harder

packages/extension-utils/src/OAuthForm/src/state/sagas.ts
```diff
@@ -0,0 +1,201 @@
...
+    clearFormAction,
+    handleVerifyAction,
+    handleSaveConfigAction,
+  } = OAuthFormActions
+  yield* takeEvery(initAction, initSaga)
+  yield* takeEvery(handleUrlChangeAction, handleUrlChangeSaga)
+  yield* takeEvery(clearFormAction, clearFormSaga)
+  yield* takeEvery(handleVerifyAction, handleVerifySaga)
+  yield* takeEvery(handleSaveConfigAction, handleSaveConfigSaga)
+}
```
josephaxisa, line 201: Let's add full test coverage for all the sagas in a `sagas.spec.ts` file

PR #1329, opened 2023-06-20
feat: Add filteredQuery to tile host data

ryandunlavy:
This parameter has been added on the Looker side, so catching up the extension SDK to reflect this parameter in the tileHostData.


bryans99 (approved):
LGTM

PR #1338, opened 2023-07-15
Formatted README file to 80chars where appropriate

### LINKED ISSUE #1337: README file not legible without line wrap
Not a major issue, but a simple thing to fix, the current README requires line wrap to be on in order to read it. Not super ideal for us terminal-folk :)

prescottbreeden:
The current README file isn't readable in a text editor without turning linewrap on. I believe VScode and other IDEs under the hood are naught and turn autowrap on by default in .md files so most people are always unawares.

Fixes #1337 🦕


ryandunlavy:
LGTM

jkaster (approved):
this looks good but what we really need to do is update our prettier config to match our latest internal configuration

prescottbreeden:
Yea, definitely, this is actually an issue across all our docs, but how
could i resist a PR for #1337?? :)

PR #1348, opened 2023-08-14
feat: Send message on visConfig extension update

ryandunlavy:


packages/extension-sdk/src/connect/visualization/visualization_sdk.ts
```diff
@@ -126,6 +126,9 @@ export class VisualizationSDKImpl implements VisualizationSDK {
       this.visualizationData = visualizationData
       if (this._visConfig) {
         this._visConfig.update(this.visualizationData.visConfig)
+        this.hostApi.send(ExtensionRequestType.VIS_CONFIG_UPDATE, {
```
bryans99, line 129: Oops! Should have thought of this. Thanks

bryans99 (commented):
Test?

packages/extension-sdk/src/connect/visualization/visualization_sdk.ts
```diff
@@ -126,6 +126,9 @@ export class VisualizationSDKImpl implements VisualizationSDK {
...
+          updatedConfig: this.visualizationData.visConfig,
+        })
       }
       if (this._queryResponse) {
         this._queryResponse.update(this.visualizationData.queryResponse)
```
bryans99, line 134: Do we need to send an event for this?

ryandunlavy: My thinking was that the visConfig makes sense to be mutable in the popup context since it is manipulating the visualization state, but the queryResponse should not be able to be updatable since that is coming directly from the database and we cannot update the date Looker-side of a ran query result.

ryandunlavy:
Will add a test here

bryans99 (approved):

PR #1356, opened 2023-08-21
fix: sync up package versions

bryans99:


packages/embed-playground/package.json
```diff
@@ -42,8 +42,8 @@
     "@looker/embed-components": "*",
     "@looker/extension-utils": "^0.1.21",
     "@looker/redux": "^0.0.1",
-    "@looker/sdk": "^23.2.0",
-    "@looker/sdk-rtl": "^21.6.0",
+    "@looker/sdk": "^23.14.1",
+    "@looker/sdk-rtl": "^21.6.1",
```
jkaster, line 46: If we set these to `*` does release-please update them for publishing? Or is this an extra lint/package dependency step we're going to need to add to our generation/publishing step? @treshugart would like your thoughts here

jkaster (approved):

packages/embed-playground/package.json
bryans99: To be honest I dont know. I'm just syncing the packages.

jkaster: Understood. That's why I summoned Trey 😸

jkaster: This is *not* a blocker for your PR, which I've approved.

PR #1364, opened 2023-09-21
chore: create yarn.lock

drstrangelooker:
Resolve security issue


bin/looker-resources-index/yarn.lock
```diff
@@ -0,0 +1,13 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
treshugart, line 1: From the description:

> This folder contains data (in code) about Looker-related resources and scripts to maintain and repackage that data for publication.

It doesn't appear that this is production code, or even intended to be installed? @jkaster can confirm. If this is the case, there is no need for a lockfile.

If, indeed, this is meant to be installed _and_ is used in production, we should add it to this folder (`bin/looker-resources-index`) to the `workspaces` array in the root `package.json` file.

Either way, we should not add a lockfile here.

treshugart (changes requested):
Left a comment re `bin/looker-resources-index/yarn.lock`.

For `examples/access-token-server/...`, this doesn't appear to exist in `main`. Is this based off stale changes? Either way, anything in there probably falls into the same camp as `bin/looker-resources-index` and should, _at most_, be added to `workspaces` in the root `package.json`.

For `node_modules`, this is not committed and is generated by `npm i`, so these should not be committed.

Summary: the only change, if any, is to add `bin/looker-resources-index` to `workspaces` in the root `package.json`. However, this may not even be necessary.

bin/looker-resources-index/yarn.lock
drstrangelooker: I reverted that now

jkaster: The only things consumed from this folder are the json files

jkaster (approved):
If this resolves the security bot monitoring issue, let's get it in. We will need to switch from yarn to npm for our web stack build eventually anyway

treshugart:
Really appreciate you opening a PR to fix this, @drstrangelooker. The correct way (for our context) to fix this is #1366.

Issue #1369, opened 2023-09-27
Looker SDK auth.login_user() doesn't work as expected


dbuzolin-zencore:
Trying simple example with sudo (the user_id=5 does exists in the system)
looker-sdk==23.16.0
```
sdk.me()  # user for originally configured credentials
sdk.auth.login_user(5)
sdk.me()  # user with id 5
sdk.auth.logout()
sdk.me()  # user for originally configured credentials
```
Produces this output:

```
Traceback (most recent call last):
  File "/Users/db/repos/looker-utils/src/python/list_dashboards.py", line 274, in <module>
    main(args)
  File "/Users/db/repos/looker-utils/src/python/list_dashboards.py", line 219, in main
    sdk.auth.login_user(5)
  File "/Users/db/.pyenv/versions/3.9.13/lib/python3.9/site-packages/looker_sdk/rtl/auth_session.py", line 118, in login_user
    self._login_sudo(transport_options or {})
  File "/Users/db/.pyenv/versions/3.9.13/lib/python3.9/site-packages/looker_sdk/rtl/auth_session.py", line 172, in _login_sudo
    response = self._ok(
  File "/Users/db/.pyenv/versions/3.9.13/lib/python3.9/site-packages/looker_sdk/rtl/auth_session.py", line 240, in _ok
    raise error.SDKError(response.value.decode(encoding="utf-8"))
```

Sudo to this user works fine in UI

dbuzolin-zencore:
Please close this, this was due to wrong Looker instance used in initing the SDK.

PR #1376, opened 2023-10-12
fix(return type):Update render-task-result in methods.ts

### LINKED ISSUE #1375: render_task_results in typescript Looker40SDK returns string
render_task_results in typescript Looker40SDK returns string instead of blob.

Not able to convert blob string to URL.createObjectURL in typescript. This function accepts only Blob

yogendrasgautam:
Fix for #1369 - Update render-task-result in methods.ts to return Blob result

## 👋👋 Thank you for contributing to Looker sdk-codegen (⚡️🍣)

- 👉 While using render_task_result call to render the pdf result for dashboard. it returns string blob which is not accepted as parameter to URL.createObjectUrl(). URL.createObjectUrl() function accepts the value in Blob format

Fixes #1375  🦕


jeremytchang:
Based on the linked issue, it was not an issue with the code. Also our methods are auto generated and should not be manually changed. 

Closing for now, feel free to re-open if you want to update the PR.

PR #1381, opened 2023-10-13
Kotlin: allow pluggable HTTP clients

### LINKED ISSUE #1380: Kotlin: Allow pluggable HTTP clients
The Kotlin SDK currently uses `ktor` as a HTTP client. I am working in an environment where `ktor` is not an accepted client preventing me from using the Kotlin SDK.

I'd like to replace the `ktor` implementation with the [Google HTTP Java Client](https://github.com/googleapis/google-http-java-client). This client allows for pluggable low level HTTP transport: https://googleapis.github.io/google-http-java-client/http-transport.html. Three transports are provided. 

The existing `ktor` transport may be updated to implement the [HttpTransport abstract class](https://github.com/googleapis/google-http-java-client/blob/main/google-http-client/src/main/java/com/google/api/client/http/HttpTransport.java#L71) and related interfaces.

tjbanghart:
## 👋👋 Thank you for contributing to Looker sdk-codegen (⚡️🍣)

- 👆 Make sure your pull request title follows [Pull Request Title Guidelines](https://github.com/looker-open-source/sdk-codegen/blob/main/CONTRIBUTING.md#title-guidelines) from our [Contribution guide](https://github.com/looker-open-source/sdk-codegen/blob/main//CONTRIBUTING.md)
- 👉 Don't forget to replace these instructions with your _✨awesome✨_ description of what this change _actually does_. Additionally, it's great to include context on how it works and why the change was needed.
- 👇 Edit "Developer Checklist" to reflect items relevant to this PR (and try to make sure to check everything off before asking for review)

Fixes #1380 🦕


kotlin/build.gradle
```diff
@@ -32,10 +31,10 @@ dependencies {
     implementation "io.github.cdimascio:dotenv-kotlin:6.2.2"
     implementation "org.ini4j:ini4j:0.5.4"
 
-    implementation "io.ktor:ktor-client:$ktorVersion"
-    implementation "io.ktor:ktor-client-okhttp:$ktorVersion"
-    implementation "io.ktor:ktor-client-json:$ktorVersion"
-    implementation "io.ktor:ktor-client-gson:$ktorVersion"
```
tjbanghart, line 38: Removed for now but can add back if we decide to have a low-level ktor client to make the requests

kotlin/src/main/com/looker/rtl/Transport.kt
```diff
@@ -90,20 +94,26 @@ fun <T> ok(response: SDKResponse): T {
...
+    DELETE,
+    PATCH,
+    HEAD
+}
+
+enum class HttpTransports(val label: String) {
+    APACHE("Apache HTTP Client"),
+    JAVA_NET("Native Java HTTP Client"),
+    // URL_FETCH("Google App Engine HTTP Client"), TODO: App Engine support? Requires App Engine SDK.
+    // KTOR("Kotlin based HTTP Client") TODO: Add ktor transport wrapper.
```
tjbanghart, line 110: Is there a requirement to use `ktor`? If so I can make the wrapper class.

Apache and java.net are fairly standard.

jkaster:
Did you run the kotlin smoke tests with this change?

tjbanghart:
Ah, let me double check that. I ran the unit tests in intellij but should probably use the script. I'll mark as WIP and test it out again.

tjbanghart:
Okay, passing smoke tests but one issue with warnings about invalid cookies. I left a comment on the `TODO`

kotlin/src/main/com/looker/rtl/Transport.kt
```diff
@@ -267,100 +242,174 @@ class Transport(val options: TransportOptions) {
...
+     *
+     * Will disable SSL certificate verification iff [TransportOptions.verifySSL] is false.
+     */
+    fun initTransport(options: TransportOptions): HttpTransport {
+        return when (HttpTransports.valueOf(options.httpTransport.uppercase())) {
+            HttpTransports.APACHE -> {
+                // TODO: fix bug upstream that does not pass client context to requests.
+                //  The Netscape `expire` datetime format is not compatible with the "default" and
+                //  will log invalid warnings.
+                val clientBuilder =
```
tjbanghart, line 287: The response cookies we get back from Looker use the ` Wdy, DD-Mon-YY HH:MM:SS GMT` format for `expires`. We should be able to push down the config here to the low-level Apache request but the request does not reference the client context. I believe it is a bug in the Google Java Client.

Note, this doesn't stop requests, but rather logs a confusing error. We still get a correct response back. 

Alternatively, could we use `disableCookieManagement()` so the low-level client doesn't try to validate cookies. The SDK usually shouldn't care about cookies right? Wouldn't the required auth token header make cookies redundant?

kotlin/src/main/com/looker/rtl/Transport.kt
jkaster: There's not a requirement to use `ktor` although that appears to be the more "standard" HTTP client for Kotlin. Have you looked at the dependency size impact of the Apache and/or Java HTTP clients? Is Ktor a layer on top of Java HTTP? I haven't looked since we implemented the original SDK.

kotlin/src/main/com/looker/rtl/Transport.kt
jkaster: We definitely have responses that would set cookies and those would be needed (the ironically named cookieless embed is one example) but this could be a case of worrying about it when we discover someone needs it. The only concern I have is whether the Looker Android mobile client reads any cookies from Looker, but it probably doesn't? I've not really looked at how cookies are used in non-browser situations and someone could implement a custom payload parser if it really was required as long as we have the hooks available to get the original request. 

Do we have a test that verifies a payload with missing "required" fields still gets parsed correctly? Or are you still using GSon for parsing with whichever client? I'll look at the code now that I'm on a bigger screen.

kotlin/src/main/com/looker/rtl/Transport.kt
tjbanghart: Prior to this change we were using the `OkHttp` engine but `ktor` supports several types including Apache: https://ktor.io/docs/http-client-engines.html.

In my understanding `ktor` and the Google HTTP Client do similar things in that they abstract "low-level" HTTP request/responses handling. `ktor` provides more Kotlin specific features, like coroutines, but I am not sure what else is gained. Both have Android specific transports as well if that is a requirement.

jkaster (dismissed):
I have questions but nothing that blocks approval for this PR

kotlin/src/main/com/looker/rtl/Transport.kt
jkaster: Ah. I think @brianmadden  I talked about coroutines for `ktor` and that's why we wanted to use it, based on what he'd done in Krawler. That was a long time ago, though, so I'm not sure at this point.

jkaster (approved):

PR #1382, opened 2023-10-13
chore: fix build

treshugart:
Fix issue causing a broken build.

_We should probably run the build in CI (part of `npm test`) as it would have surfaced the error in the original PR._

The core problem was lerna wasn't filtering out workspaces we don't want to build.

Ideally we would use `npx -w packages`, but NPM doesn't topologically sort and we'd have to declare each workspace in order of dependence manually. See: https://github.com/npm/cli/issues/4139. NPM moves so slowly and has a general distaste - it seems - for adopting useful behavior (see pnpm), I wouldn't hold our breath. Lerna make sense for now. Hooooooowever, onto the next point

To move away from Lerna and Yarn, we'd need to do a combo of TypeScript project references (to remove the need for `--sort` when running `tsc`) and define `build` scripts in each package that needs to build, then use `npm run build --if-present --workspaces. We really shouldn't be defining root scripts that operate on workspaces.


package.json
```diff
@@ -35,9 +35,9 @@
     "view": "yarn api-explorer",
     "wipe": "rm -rf api spec",
     "build": "rm -Rf packages/*/lib && run-p -c build:*",
-    "build:cjs": "lerna exec --stream 'BABEL_ENV=build_cjs babel src --root-mode upward --out-dir lib --source-maps --extensions .ts,.tsx --no-comments'",
-    "build:es": "lerna exec --stream 'BABEL_ENV=build babel src --root-mode upward --out-dir lib/esm --source-maps --extensions .ts,.tsx --no-comments'",
-    "build:ts": "lerna exec --stream --sort 'tsc -b tsconfig.build.json'",
+    "build:cjs": "lerna exec --ignore @internal/* --stream 'BABEL_ENV=build_cjs babel src --root-mode upward --out-dir lib --source-maps --extensions .ts,.tsx --no-comments'",
```
treshugart, line 38: We have to have some sort of marker to filter unwanted workspaces. We can't use `private` because we have private workspaces in packages/ that need to be built. And we can't target just the packages/ folder because Lerna.

Moving away from root build scripts, and defining them in pacakges and using `--if-present` when running them would be the correct way to do this, I think.

josephaxisa (approved):

PR #1378, opened 2023-10-12
feat: generate SDKs for Looker 23.18

drstrangelooker:
Release-As: 23.18.0

PR #1388, opened 2023-10-20
chore: Revert "feat: generate SDKs for Looker 23.18"

drstrangelooker:
Reverts looker-open-source/sdk-codegen#1378

This allows other code to land while waiting for sdk_23_18 without 3.1 endpoints.


csharp/rtl/Constants.cs
```diff
@@ -61,7 +61,7 @@ public struct Constants
 
         public const string DefaultApiVersion = "4.0";
         public const string AgentPrefix = "CS-SDK";
-        public const string LookerVersion = "23.18";
+        public const string LookerVersion = "23.17";
```
jkaster, line 64: This isn't based on a CI image then? Should it be 23.16 still?

kotlin/src/main/com/looker/sdk/Constants.kt
```diff
@@ -28,7 +28,7 @@ package com.looker.sdk
 
 const val ENVIRONMENT_PREFIX = "LOOKERSDK"
 const val SDK_TAG = "KT-SDK"
-const val LOOKER_VERSION = "23.18"
+const val LOOKER_VERSION = "23.17"
```
jkaster, line 31: consistent with the bad version. If this isn't going to be published I'm not sure it matters.

packages/sdk/src/4.0/funcs.ts
```diff
@@ -6805,13 +6811,13 @@ export const search_groups = async (
...
  * @param sdk IAPIMethods implementation
- * @param request composed interface "IRequestSearchGroups" for complex method parameters
+ * @param request composed interface "IRequestSearchGroupsWithRoles" for complex method parameters
  * @param options one-time API call overrides
  *
  */
 export const search_groups_with_roles = async (
   sdk: IAPIMethods,
-  request: IRequestSearchGroups,
+  request: IRequestSearchGroupsWithRoles,
```
jkaster, line 6820: interesting. I wonder if we've introduced a breaking change in this SDK call.

jkaster (commented):
Not sure about the 23.17 version string

csharp/rtl/Constants.cs
drstrangelooker: No, somehow the version number got screwed up last time. But when we publish this again they'll be fixed.

packages/sdk/src/4.0/funcs.ts
drstrangelooker: Not sure.

drstrangelooker: This is just a revert so we will address when the sdk_23_18 actually rolls forward.

jkaster (approved):

PR #1402, opened 2023-11-27
feat: generate SDKs for Looker 23.20

deanlooker:
Release-As: 23.20.0


jeremytchang (dismissed):
Thanks a lot for taking care of this Dean! Looks like 23.21 version was generated.

csharp/sdk/4.0/methods.cs
```diff
@@ -3746,7 +3753,7 @@ namespace = SdkUtils.EncodeParam(namespace);
   /// You can use this function to change the string and integer properties of
   /// a dashboard. Nested objects such as filters, dashboard elements, or dashboard layout components
   /// cannot be modified by this function - use the update functions for the respective
-  /// nested object types (like [update_dashboard_filter()](#!/4.0/Dashboard/update_dashboard_filter) to change a filter)
+  /// nested object types (like [update_dashboard_filter()](#!/3.1/Dashboard/update_dashboard_filter) to change a filter)
```
jkaster, line 3756: need a ticket to fix this internally but it can wait until 24.0

kotlin/src/main/com/looker/sdk/4.0/methods.kt
```diff
@@ -48,12 +50,13 @@ class LookerSDK(authSession: AuthSession) : APIMethods(authSession) {
      * POST /alerts/{alert_id}/follow -> Void
      */
     fun follow_alert(
-        alert_id: String,
-    ): SDKResponse {
+        alert_id: String
+    ) : SDKResponse {
```
jkaster, line 54: we need to automate the new linter script. Perhaps @tjbanghart has created a reformatter script for kotlin already?

jkaster (commented):
We may want the Kotlin reformatted and apply the `models` alias for python as mentioned in other chats

drstrangelooker (approved):
LGTM

csharp/sdk/4.0/methods.cs
deanlooker: Should I bother manually changing these for 23.20 or just let it sit since it's just comments?

kotlin/src/main/com/looker/sdk/4.0/methods.kt
tjbanghart: I have not but will create an issue to do so. Thank you for flagging. I can follow up with the lint change when that is complete.

csharp/sdk/4.0/methods.cs
jkaster: no, we should change the source to correct this for 24.0. We just need to make sure we have a ticket to keep track.

PR #1417, opened 2023-12-26
doc: Update tile extension SDK api documentation

bryans99:
Also breaks out a couple of functions that are for internal use only into their own interfaces.

Also deprecates and replaces the addErrors and clearErrors functions.

Finally:
1. changes the tile extension theme to use google coloring.
2. changes the tile extensions to use the new addError and clearError functions.


packages/extension-sdk/src/connect/tile/types.ts
```diff
@@ -204,18 +204,63 @@ export interface Filters {
...
+   * Tile host data.
+   */
   tileHostData: TileHostData
-  tileHostDataChanged: (hostData: Partial<TileHostData>) => void
+  /**
+   * Display an error beneath the tile extension.
+   */
+  addError: (error: TileError) => void
+  /**
+   * Clear an error beneath the tile extension.
```
jkaster, line 227: I'm wondering when I'd use single errors vs. multiple errors as defined below

jkaster (approved):
I have a question but it doesn't block approval of this PR

PR #1434, opened 2024-02-21
chore: Fix NPM publishing workflow

jeremytchang:
Our publishing workflow was unable to execute `lerna publish` to publish our npm packages because its method of execution `npm bin` is not supported anymore. This stackoverflow article for more context: https://stackoverflow.com/questions/9679932/how-to-use-executables-from-a-package-installed-locally-in-node-modules

- Switching to using `npx`. I confirmed locally that `npx lerna -h` works which means other lerna commands should work. (unable to do publishing dry run because our version of lerna is too old, and doesn't have that option)


drstrangelooker (approved):
Looks LGTM To Me

PR #1435, opened 2024-02-27
chore: Changed imported package models to models40

renanbenedicto:
When tested with requirements below, `looker_sdk` no longer have `models` class. So the solution is to import the equivalent class `models40` which I chose the latest Looker API version (4.0).

Then it runs perfectly !

### Requirements:
looker-sdk==24.2.0

Fixes #1 🦕


deanlooker:
We've merged a change which includes this and addresses the same issue in other examples. 

https://github.com/looker-open-source/sdk-codegen/pull/1436

PR #1440, opened 2024-03-08
feat: error-doc handling in python SDK

drstrangelooker:
Lookups now take place against the CDN to find a more detailed error guide. The url to that error guide, and the error guide itself - a markdown document - are included in the SDKError class and the ErrorDetail class.

### DIFF
--- a/python/looker_sdk/error.py
+++ b/python/looker_sdk/error.py
@@ -21,33 +21,104 @@
 # THE SOFTWARE.
 
 import attr
-from typing  import Optional, Sequence
+from typing import cast, Dict, Optional, Sequence, Tuple
+import requests
+import json
+import re
 
 """API error class
 """
 
+
 @attr.s(auto_attribs=True, kw_only=True)
-class ErrorDetail():
+class ErrorDetail:
     """Error detail:
-        documentation_url: documentation link
-        field: field with error
-        code: error code
-        message: error info message
+    documentation_url: documentation link
+    field: field with error
+    code: error code
+    message: error info message
     """
+
     documentation_url: str
     field: Optional[str] = ""
     code: Optional[str] = ""
     message: Optional[str] = ""
+    error_doc_url: str = ""
+    error_doc: str = ""
 
 
 @attr.s(auto_attribs=True)
 class SDKError(Exception):
     """API error class:
-        message: main error info message
-        errors: array of error details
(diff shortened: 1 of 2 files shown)


drstrangelooker:
Updated the SDKError and ErrorDoc objects with a `__str__()` method so that they automagically display some useful data.
```
looker_sdk.error.SDKError: 
    message:           Not found
    documentation_url: https://cloud.google.com/looker/docs/r/err/4.0/404/get/connections/:connection_name
    error_doc_url:     https://static-a.cdn.looker.app/errorcodes/404.md
    error details:
```

python/looker_sdk/error.py
```diff
@@ -21,33 +21,128 @@
...
-        documentation_url: documentation link
-        field: field with error
-        code: error code
-        message: error info message
+    documentation_url: documentation link
+    field: field with error
+    code: error code
+    message: error info message
+    error_doc_url: URL that may point to additional useful information
+    error_doc: Markdown doc that may contain additional useful information
```
jkaster, line 41: This is a nice addition to the error class, but I'm wondering if `error_doc` retrieval could cause cascading "SDK" failures if access to the CDN is blocked for the SDK's running location

drstrangelooker: I catch all request errors at line 147-148 and jsut put empty strings in these fields if anything goes wrong.

python/looker_sdk/error.py
```diff
@@ -21,33 +21,128 @@
...
+    ) -> Tuple[str, str]:
+        m = re.search(self.RE_PATTERN, error_url)
+        if not m:
+            return ("", "")
+
+        code: str = cast(Tuple[str, str, str, str], m.groups())[2]
+        path: str = cast(Tuple[str, str, str, str], m.groups())[3]
+        try:
+            return self.lookup(url=url, code=code, path=path)
+        except requests.exceptions.RequestException:
```
drstrangelooker, line 147: @jkaster RequestException is the parent exception of everything in the requests library. If anything goes wrong it is caught here.

jkaster (approved):
This is a slightly different implementation than we have for other SDKs due to needing to keep SDKs like TS as lean as possible, but this feels like it could be a good approach for the Python SDK, and perhaps some of the others. Will consider it for other SDK implementations.

PR #1455, opened 2024-04-15
feat: generate SDKs for Looker 24.6

drstrangelooker:
There are a large number of changes to spec/Looker.4.0.json and spec/Looker.4.0.oas.json noting that 429 is a valid http error code. That is the code for Rate Limit Exceeded. This does not change the generated SDK code.

Release-As: 24.6.0


deanlooker (dismissed):
lgtm

PR #1475, opened 2024-06-20
feat: Add option to pass external contexts into Go SDK's HTTP calls

JCPistell:
This adds the option to include an external context in the Go SDK's HTTP calls. This is useful if the SDK is being used as part of a larger application that needs to share a context across multiple API boundaries.


go/rtl/auth_test.go
```diff
@@ -635,6 +635,65 @@ func TestAuthSession_Do_Timeout(t *testing.T) {
...
+		mux := http.NewServeMux()
+		setupApi40Login(mux, foreverValidTestToken, http.StatusOK)
+		server := httptest.NewServer(mux)
+		defer server.Close()
+
+		mux.HandleFunc("/api"+apiVersion+path, func(w http.ResponseWriter, r *http.Request) {
+			time.Sleep(4 * time.Second)
+		})
+
+		ctx, cncl := context.WithTimeout(context.Background(), 1*time.Second)
```
iyabchen, line 649: There is no guarantee that the test can always timeout when it runs session.Do. Better to sleep 1 second to trigger that and reduce flakiness.

iyabchen: missed the handle func sleep 4 seconds. all look good.

go/README.md
```diff
@@ -118,3 +118,46 @@ func main() {
     }
 }
 ```
+### Custom Context
+
+If you want even greater control over the lifecycle of the HTTP requests consider providing a [context](https://pkg.go.dev/context). Contexts can be set for all requests or per request.
+
+#### Custom Context for all requests
+
+Follow the example code snippet below if you want all requests to use the same context:
```
jeremytchang, line 127: Need to make it very clear that the `Timeout` in the api settings/config will be ignored if a custom context is set on all requests. The developer should use context.WithTimeout() to set a timeout. 

Also should note that the custom context will also apply on the automatic token fetching request.

go/rtl/auth_test.go
```diff
@@ -635,6 +635,65 @@ func TestAuthSession_Do_Timeout(t *testing.T) {
 			t.Errorf("Do() call did not error with context.DeadlineExceeded, got=%v", err)
 		}
 	})
+
+	t.Run("Do() follows Context set in AuthSession config", func(t *testing.T) {
```
jeremytchang, line 639: A bit of an ask, but could we test the overriding too?  (atleast partially)
So the first test case would only have config timeout set at 1 sec as it is now, then the next would have config timeout at 5 and option timeout 1, then next would config timeout 5, option timeout 5, and config context at 1 sec, and etc.

jeremytchang (commented):
Great addition!  
I left a few comments on updating documentation and comments to make clear what the behavior/side effects will be when setting context on per or all requests. Also had a small ask around increased testin.

go/README.md
```diff
@@ -118,3 +118,46 @@ func main() {
...
+    cfg.Context = ctx
+
+    session := rtl.NewAuthSession(cfg)
+    sdk := v4.NewLookerSDK(session)
+}
+```
+
+#### Custom Context per request
+
+Follow the example here to set a context for a specific request. **This will override any context set in the SDK config as outlined in the previous section.**
```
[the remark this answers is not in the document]
JCPistell: Done.

go/README.md
JCPistell: Done.

go/rtl/auth.go
```diff
@@ -67,8 +67,13 @@ func NewAuthSessionWithTransport(config ApiSettings, transport http.RoundTripper
 		AuthStyle:    oauth2.AuthStyleInParams,
 	}
 
+	bgCtx := context.Background()
```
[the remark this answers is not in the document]
JCPistell: I've added some notes to the README sections - let me know if you think they're sufficient

go/rtl/auth_test.go
JCPistell: Done.

go/rtl/auth.go
```diff
@@ -123,18 +128,25 @@ func (s *AuthSession) Do(result interface{}, method, ver, path string, reqPars m
...
-	}
-	if options != nil && options.Timeout != 0 {
-		timeoutInSeconds = options.Timeout
+	var ctx context.Context
+	if options != nil && options.Context != nil {
+		ctx = options.Context
+	} else if s.Config.Context != nil {
+		ctx = s.Config.Context
+	} else {
+		// create request context with timeout
```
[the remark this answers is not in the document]
JCPistell: Whoops - forgot this one in the last push... hold pls...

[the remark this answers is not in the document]
JCPistell: Done.

jeremytchang (approved):
Thanks for taking care of my comments! And your documentation looks great! LGTM

jeremytchang (changes requested):
Blocking for now as there's a proposal that would supercede this.

go/rtl/auth.go
```diff
@@ -123,16 +128,24 @@ func (s *AuthSession) Do(result interface{}, method, ver, path string, reqPars m
...
+	var timeoutInSeconds int32 = 120
 	if s.Config.Timeout != 0 {
 		timeoutInSeconds = s.Config.Timeout
 	}
 	if options != nil && options.Timeout != 0 {
 		timeoutInSeconds = options.Timeout
 	}
 
-	ctx, cncl := context.WithTimeout(context.Background(), time.Second*time.Duration(timeoutInSeconds))
+	ctx, cncl := context.WithTimeout(parent, time.Second*time.Duration(timeoutInSeconds))
```
jeremytchang, line 148: If the parent context has a different timeout, what's the expected behavior here?

jeremytchang, line 148: I think an easier to understand behavior would be:
If you use config context or options context, the timeout defined in options or config is ignored. 

Trying to document/understad the how timeout works with context/config/options is confusing and harder to test.

go/README.md
```diff
@@ -118,3 +118,57 @@ func main() {
...
+    session := rtl.NewAuthSession(cfg)
+    sdk := v4.NewLookerSDK(session)
+}
+```
+
+> Note: A context set here will become the parent context for all API calls as well as all requests to fetch/refresh oauth tokens, which are normally completely isolated from contexts set via the Timeout property. In this case the token refresh requests and each individual API call will share a common parent context.
+
+#### Custom Context per request
+
+Follow the example here to set a context for a specific request.
```
jeremytchang, line 153: + "The custom context per request will replace the custom context for all requests"

go/README.md
```diff
@@ -118,3 +118,57 @@ func main() {
...
+    
+    cfg, err := rtl.NewSettingsFromFile("path/to/looker.ini", nil)
+    cfg.Context = ctx
+
+    session := rtl.NewAuthSession(cfg)
+    sdk := v4.NewLookerSDK(session)
+}
+```
+
+> Note: A context set here will become the parent context for all API calls as well as all requests to fetch/refresh oauth tokens, which are normally completely isolated from contexts set via the Timeout property. In this case the token refresh requests and each individual API call will share a common parent context.
```
jeremytchang, line 149: "The SDK will ignore any timeout set in the Timeout section if you use custom context for all requests." You must set the timeout with the custom context.

go/README.md
```diff
@@ -118,3 +118,57 @@ func main() {
...
+    
+    cfg, err := rtl.NewSettingsFromFile("path/to/looker.ini", nil)
+    session := rtl.NewAuthSession(cfg)
+    sdk := v4.NewLookerSDK(session)
+
+    sdk.Me("", &ApiSettings{Context: ctx})
+}
+```
+
+> Note: Setting a context per request will NOT affect the context used for the background token fetching requests. If you have also set a context for all requests as mentioned above then that context
```
jeremytchang, line 173: Also add "If you set a custom context for a specific request, the SDK will ignore any timeout described in the previous Timeout section for the specific request. You must set the timeout with the custom context for the specific request."

go/README.md
```diff
@@ -118,3 +118,57 @@ func main() {
     }
 }
 ```
+### Custom Context
+
+If you want even greater control over the lifecycle of the HTTP requests consider providing a [context](https://pkg.go.dev/context). Contexts can be set for all requests or per request. They can be combined with the timeout options mentioned in the previous section to fine-tune the lifecycle of your requests.
+
+If you wish to include timeout functionality in your custom context then you should leverage [context.WithTimeout](https://pkg.go.dev/context#WithTimeout).
```
jeremytchang, line 125: Also add: ""The SDK will ignore any timeout described in the previous Timeout section for all requests if you use custom context for all requests, or per requests if you use custom context per request."

go/README.md
```diff
@@ -118,3 +118,57 @@ func main() {
...
+
+> Note: Custom contexts will be used as the parent context for any timeout you set as specified in the previous section. If the parent context gets cancelled it will propagate to the child context, but if the timeout context times out it does not propagate to the parent context.
```
jeremytchang, line 127: Not sure what this means? We can remove then if new behavior?

go/README.md
```diff
@@ -118,3 +118,57 @@ func main() {
     }
 }
 ```
+### Custom Context
+
+If you want even greater control over the lifecycle of the HTTP requests consider providing a [context](https://pkg.go.dev/context). Contexts can be set for all requests or per request. They can be combined with the timeout options mentioned in the previous section to fine-tune the lifecycle of your requests.
```
jeremytchang, line 123: Change to: "You can set a [context](https://pkg.go.dev/context) for all requests or a specific request for more control over the SDK's HTTP requests. "

go/README.md
jeremytchang: Change to "Note: The context set here will become the parent context for all requests including the SDK's automated oauth token fetching/refreshing requests. If you want to control the timeout of the oauth token fetching/refreshing requests, you must set a custom context for all requests."

go/README.md
```diff
@@ -118,3 +118,57 @@ func main() {
...
+}
+```
+
+> Note: A context set here will become the parent context for all API calls as well as all requests to fetch/refresh oauth tokens, which are normally completely isolated from contexts set via the Timeout property. In this case the token refresh requests and each individual API call will share a common parent context.
+
+#### Custom Context per request
+
+Follow the example here to set a context for a specific request.
+
+> Note: This will be used as the parent context for any timeout setting you've specified for API calls. If you've set contexts in both your API config and in the request options the request options context will be used instead. Background requests to fetch/refresh oauth tokens will NOT use a context set via request options - it will default to use a generic background context or, if you've also set a context in the API config it will still use that as specified in the previous section.
```
jeremytchang, line 155: Can be removed

go/README.md
jeremytchang, line 173: Replace with
`Note: If you set a custom context for a specific request, the SDK will ignore any timeout described in the previous Timeout section for the specific request. You must set the timeout with the request's custom context.

Note:
The custom context per request does NOT affect the background oauth token requests' context. You must use a custom context set on all requests to do so.`

go/rtl/auth.go
JCPistell: Not exactly. If the parent hits its cancellation condition (e.g. a timeout) first then the cancellation propagates to the child.  If the child cancels it depends on how that specific error is handled... perhaps the request retries, perhaps the error is passed to the calling function, in which case most of the time the function fails and the parent context then cancels, which propagates the cancellation to all other goroutines, etc etc.

Basically this approach means you can have an overall timeout set by a parent, but each individual request running in a goroutine can also have its own timeout... this avoids the scenario where you set a 60min overall timeout, but one bad request has to wait the entire 60m to time out and fail.

JCPistell: I disagree - see my previous comment. I actually think this is _easier_ to document and the tests are already in place.

go/README.md
JCPistell: This isn't true - request timeouts will still apply - any specified context simply becomes the parent of the timeout request.

go/README.md
JCPistell: See my previous comment - this is inaccurate.

drstrangelooker:
@JCPistell has this been superseded? Can we close it?

Issue #1439, opened 2024-03-05
Security Issue with request 2.88 in @looker/sdk and @looker/sdk-rtl


RobinShift4:
I have a project using @looker/sdk as a dependency and when I run:

`yarn run audit`

I get

`Severity: MODERATE
  Modules: @looker/sdk-rtl>request, @looker/filter-components>@looker/sdk-rtl>request, @looker/filter-components>@looker/sdk>@looker/sdk-rtl>request, @looker/filter-components>@looker/filter-expressions>@looker/sdk>@looker/sdk-rtl>request
  URL: https://github.com/advisories/GHSA-p8p7-x288-28g6`

four times.  I see that @looker/sdk and @looker/sdk-rtl are using

`request "^2.88.0"`

which has this security issue.  Are there plans to upgrade?

mnahkies:
`request` has been deprecated for years at this point - it really needs replacing IMO https://github.com/request/request/issues/3142

xiel:
I case someone needs a quick fix for this issue, the cypress team maintains a fork with a fix.

You can get it into your project by using npm overrides.

```
"overrides": {
  "request": "npm:@cypress/request@^3.0.0"
},
```

smartin88:
@xiel thanks for the solution! This works fine for me when I run my project locally, however as soon as it runs in a Docker container, it fails:

`Cannot find module 'request' Required stack: -/app/server/node_modules/@looker/sdk-node/lib/nodeTransport.js`

Really struggling to find a way to fix this, do you have any ideas?

xiel:
@smartin88 That difference between local and docker is weird. Which package manager do you use? How are the node_modules installed into the docker container? Do you install from inside the docker container or copy the modules from outside in? I am wondering if it might be related to symlinks...

Bit hard to debug from the outside...

smartin88:
Hey @xiel, we use pnpm as our package manager, and the modules are compiled and then copied from outside into the Docker container. Thanks for the thoughts regarding symlinks...appreciate this is hard to debug from outside, I will look at that as a possible issue

PR #1491, opened 2024-08-26
fix: Remove the requests package from the TypeScript SDK

jkaster:
Removes the request package from all TypeScript SDK packages

- Requires Node v22
- For browser, uses the standard `fetch` function
- For NodeJS, uses Node v22 global `fetch` function

Addresses #1439


drstrangelooker (approved):
Looks LGTM To Me

prescottbreeden (commented):
I would break this CL up, it is too large to critically review. I lost my place and on second review trying to use the viewed option is not tenable with this many files.

this is an excellent POC, I would now use this as your template to make the changes first to Prettier, and then perhaps node 22, and then your fetch updates.

I will leave this as non-blocking

jest.setup.js
```diff
@@ -32,6 +32,7 @@ const ResizeObserver = require('resize-observer-polyfill');
 require('@testing-library/jest-dom');
 require('jest-canvas-mock');
 require('jest-styled-components');
+// require('jest-environment-node');
```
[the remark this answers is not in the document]
jkaster: Yes. I will have a cleanup pass in another PR that removes stale commented-out code. The prettier changes had to be made to get this PR to pass CI after code was taken from internal changes.

packages/sdk-codegen-scripts/src/fetchSpec.spec.ts
```diff
@@ -35,18 +36,22 @@ import {
...
-const props = config.section as unknown as ISDKConfigProps;
+const config = TestConfig(specToModel);
+const props = ApiConfigSection(
+  fs.readFileSync(config.localIni, 'utf8')
+) as unknown as ISDKConfigProps;
 // api_version is no longer part of the INI, now set by sdkGen iterator
 props.api_version = '4.0';
 
-describe('fetch operations', () => {
+// TODO get real fetch working for command-line Jest like it does in IntelliJ and VS Code
```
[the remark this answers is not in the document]
jkaster: This will get changed as we adopt more Node test runner usage instead of jest.

packages/sdk-codegen-scripts/src/prettify.spec.ts
```diff
@@ -2,7 +2,7 @@
 
  MIT License
 
- Copyright (c) 2021 Looker Data Sciences, Inc.
+ Copyright (c) 2023 Looker Data Sciences, Inc.
```
jkaster, line 5: weird. This was **not** a renamed file but a new file.

packages/sdk-codegen-scripts/src/prettify.ts
```diff
@@ -35,18 +35,15 @@ const prettierTs: prettier.Options = {
   semi: true,
   singleQuote: true,
   trailingComma: 'es5',
-  arrowParens: 'always',
+  arrowParens: 'avoid',
```
jkaster, line 38: the primary cause of all the file reformatting in this PR. Required for compatibility with internal versions of the SDK.

packages/sdk-codegen-utils/src/utils.spec.ts
```diff
@@ -2,7 +2,7 @@
 
  MIT License
 
- Copyright (c) 2021 Looker Data Sciences, Inc.
+ Copyright (c) 2023 Looker Data Sciences, Inc.
```
jkaster, line 5: another file rename that wasn't

packages/sdk-node/jest.config.js
```diff
@@ -2,7 +2,7 @@
 
  MIT License
 
- Copyright (c) 2021 Looker Data Sciences, Inc.
+ Copyright (c) 2020 Looker Data Sciences, Inc.
```
jkaster, line 5: not a rename

PR #1506, opened 2024-09-16
feat: allow node 20 for the TS SDK

jkaster:
Allow Node 20 as a runtime environment for the Looker TypeScript SDK


prescottbreeden (approved):
The true work of art is but a shadow of the divine perfection.

PR #1521, opened 2024-10-16
fix: Fix bad python package exclude

jeremytchang:
The original exclude instead only included the `looker_sdk` folder. Afaict based off other examples, we should be using exclude.


drstrangelooker (approved):
Looks LGTM To Me

drstrangelooker:
I have an open bug for the Looker error that is causing the tests to fail. I'm going to force this without waiting for Looker to be fixed.

PR #1542, opened 2024-12-03
fix: Improve ok() helper function

### LINKED ISSUE #1539: The SDK does not handle error conditions properly
1. The [`SDKErrorResponse`](https://github.com/looker-open-source/sdk-codegen/blob/main/kotlin/src/main/com/looker/rtl/Transport.kt#L72-L79) type does not provide explicit access to the HTTP status code or other standard metadata
2. The [`ok()`](https://github.com/looker-open-source/sdk-codegen/blob/main/kotlin/src/main/com/looker/rtl/Transport.kt#L93-L100) function throws java.lang.Error, which should not be thrown by JVM application code
3. The [`ok()`](https://github.com/looker-open-source/sdk-codegen/blob/main/kotlin/src/main/com/looker/rtl/Transport.kt#L93-L100) function does not chain the cause of a client-side exception or include its error message

Source of the SDKErrorResponse class for convenience:
```
    /** An erroring SDK call. */
    data class SDKErrorResponse<T>(
        /** The error object returned by the SDK call. */
        val value: T,
    ) : SDKResponse() {
        /** Whether the SDK call was successful. */
        val ok: Boolean = false
    }
```

Source of the ok function for convenience
```
/**
 * Response handler that throws an error on error response, returns success result on success
 */
fun <T>fun <T> ok(response: SDKResponse): T {
    @Suppress("UNCHECKED_CAST")
    when (response) {
        is SDKResponse.SDKErrorResponse<*> -> throw Error(response.value.toString())
        is SDKResponse.SDKSuccessResponse<*> -> return response.value as T
        else -> throw Error("Fail!!")
    }
}
```

haywood:
I'm hesitant to change this function too much, since we don't know what kind of odd dependencies on its current behavior might exist. However, adding exception chaining and surfacing the original error message seems like a risk worth taking.

- Eliminate duplicated version in AuthSession by delegating to a shared implementation defined on the SDKResponse companion object.

- For the SDKError case:
  - Set Error.message from SDKError.message
  - Set Error.casue from SDKError.cause

- Also ran `ktlint -F` on the files I changed so that `gradle check` would pass the linting stage
- Also added kotlin-ci.yml to run `./gradlew jar` on kotlin changes

Fixes #1539 🦕

### DIFF
--- a/.github/workflows/kotlin-ci.yml
+++ b/.github/workflows/kotlin-ci.yml
@@ -1,0 +1,63 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+# This workflow will build a Java project with Gradle and cache/restore any dependencies to improve the workflow execution time
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-java-with-gradle
+
+name: Kotlin CI
+
+on:
+  pull_request:
+    paths:
+      - kotlin/**
+      - .github/workflows/gradle.yml
+
+  push:
+    branches: [ "main" ]
+    paths:
+      - kotlin/**
+      - .github/workflows/gradle.yml
+
+# TODO(#1544): Also run tests
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    permissions:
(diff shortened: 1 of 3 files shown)


drstrangelooker (approved):
Looks LGTM To Me

PR #1558, opened 2025-01-13
feat: generate SDKs for Looker 25.0

drstrangelooker:
The `content-validation` endpoint and associated method has a breaking change for at least some languages, notably TypeScript. The passed parameters have changed.

Release-As: 25.0.0


packages/sdk/src/4.0/methodsInterface.ts
```diff
@@ -3458,12 +3461,12 @@ export interface ILooker40SDK extends IAPIMethods {
...
    * GET /content_validation -> IContentValidation
    *
-   * @param fields Requested fields.
+   * @param request composed interface "IRequestContentValidation" for complex method parameters
    * @param options one-time API call overrides
    *
    */
   content_validation(
-    fields?: string,
+    request: IRequestContentValidation,
```
jkaster, line 3469: This is a breaking SDK change

jkaster (changes requested):
We should at least document that the content_validation function for the TypeScript SDK has a breaking change because of the newly created Request object for the endpoint

packages/sdk/src/4.0/methodsInterface.ts
drstrangelooker: sorry didn't see the top level comment.

How do you recommend we document?

drstrangelooker: I can add it to the commit message so that it goes into the changelog.

jkaster (approved):

PR #1574, opened 2025-04-15
feat: generate SDKs for Looker 25.6

drstrangelooker:
Release-As: 25.6.0


packages/sdk/src/4.0/models.ts
```diff
@@ -3247,6 +3415,17 @@ export interface IDialect {
   has_ssl_support?: boolean;
 }
 
+export interface IDialectDriverVersion {
```
jkaster, line 3418: FWIW in the next release this will be called `IDialectDriverNameVersion` but hopefully no one will start consuming it before it's renamed in 25.8

jkaster (approved):
I know CI is blocking but this all looks good.

PR #1582, opened 2025-05-06
fix(BaseTransport) Fix handling of AbortSignal availability detection in selected transport

p3drosola:
Updated the assignment of the signaller variable in the BaseTransport class to ensure proper usage of AbortSignal.timeout. The debug message for when AbortSignal.timeout is not defined has also been made to execute at the correct time.

### DIFF
--- a/packages/sdk-rtl/src/baseTransport.ts
+++ b/packages/sdk-rtl/src/baseTransport.ts
@@ -173,7 +173,7 @@ export abstract class BaseTransport implements ITransport {
     let signaller;
     if (AbortSignal.timeout) {
       const ms = sdkTimeout(options) * 1000;
-      let signaller = AbortSignal.timeout(ms);
+      signaller = AbortSignal.timeout(ms);
       if ('signal' in options && options.signal) {
         // AbortSignal.any may not be available, tolerate its absence
         // eslint-disable-next-line @typescript-eslint/ban-ts-comment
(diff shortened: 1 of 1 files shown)


packages/sdk-rtl/src/baseTransport.ts
```diff
@@ -173,7 +173,7 @@ export abstract class BaseTransport implements ITransport {
     let signaller;
     if (AbortSignal.timeout) {
       const ms = sdkTimeout(options) * 1000;
-      let signaller = AbortSignal.timeout(ms);
+      signaller = AbortSignal.timeout(ms);
```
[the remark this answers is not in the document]
p3drosola: This is incorrect. the assignment should be inside the `if(AbortSignal.timeout)` guard clause

packages/sdk-rtl/src/baseTransport.ts
```diff
@@ -189,11 +189,11 @@ export abstract class BaseTransport implements ITransport {
...
         }
-      } else {
-        console.debug(
-          'AbortSignal.timeout is not defined. Timeout will use default behavior'
-        );
       }
+    } else {
+      console.debug(
+        'AbortSignal.timeout is not defined. Timeout will use default behavior'
+      );
```
p3drosola, line 196: This log was being generated even if AbortSignal.timeout was available in the runtime, since it was attached to the wrong condition (options.signal)

packages/sdk-rtl/src/baseTransport.ts
p3drosola, line 176: Removed double let declaration, since it's block scoped, and would cause a new blocked scoped variable, which shadowed the top level `signaller` but the blocked scoped value was never included in props, since it's destroyed as soon as this block is executed.

gotenxds:
Can someone please merge this year old PR? it pollutes our server logs as well.

Issue #1454, opened 2024-04-12
Unhandled SDKError looker_sdk.error.SDKError: <exception str() failed>


bobwilkinson20:
Fresh install including a downgraded cattrs per the known issue with it.
attrs==23.2.0
cattrs==23.1.2
certifi==2024.2.2
charset-normalizer==3.3.2
idna==3.7
looker-sdk==24.4.0
requests==2.31.0
typing_extensions==4.11.0
urllib3==2.2.1 

Attempting to run example https://github.com/looker-open-source/sdk-codegen/blob/main/examples/python/simple_schedule_plan.py

Results in:
```
Traceback (most recent call last):
  File ".../schedule_updater.py", line 55, in <module>
    create_simple_schedule('1336',sdk.me().id,"This is an automated test", "assembled_pdf", "test@looker.com", "email", "Hi Looker User!", "0 1 * * *")
  File ".../schedule_updater.py", line 51, in create_simple_schedule2
    schedule = sdk.create_scheduled_plan(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../.pyenv/versions/3.12.2/lib/python3.12/site-packages/looker_sdk/sdk/api40/methods.py", line 10699, in create_scheduled_plan
    self.post(
  File ".../.pyenv/versions/3.12.2/lib/python3.12/site-packages/looker_sdk/rtl/api_methods.py", line 188, in post
    return self._return(response, structure)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../.pyenv/versions/3.12.2/lib/python3.12/site-packages/looker_sdk/rtl/api_methods.py", line 101, in _return
    raise sdk_error
looker_sdk.error.SDKError: <exception str() failed>
```

nathan-michel-devoteam:
I have got quite the same problem here when calling function sdk.create_group with Python sdk (but it works for other functions). In fact it creates the group as expected but still triggers this weird error
I have also looker-sdk version 24.4.0 and my looker instance is on 24.4.28 (Looker original)

kiran2706:
I've fixed this issue, but I can't create a new branch in this repo. Can somebody please help me fix the permission issue and assist me in making my first contribution to open source? Thank you. 

cc: @jkaster @joeldodge79 @josephaxisa

jlieow:
[...]

PR #1587, opened 2025-05-27
BugFix(SDKError): Correct error formatting in SDKError __str__ method

kiran2706:
#### Problem
Calling `str(SDKError)` raised the following runtime error:
`looker_sdk.error.SDKError: <exception str() failed>`

This happened when the SDK attempted to convert the error to a string, which failed due to a malformed generator expression or naming conflict during the formatting of `self.errors`.

#### Fix
Updated the loop variable in the generator expression from `error` to `error_details` to:
- Avoid potential name shadowing.
- Improve clarity and correctness of the `__str__` implementation.
- Resolve the runtime error so the error can now be printed correctly.

#### Notes
- The variable `error_details` represents a single instance; per PEP 8, we may later rename it to `error_detail` for readability.

### DIFF
--- a/python/looker_sdk/error.py
+++ b/python/looker_sdk/error.py
@@ -82,7 +82,7 @@ class SDKError(Exception):
     documentation_url: {self.documentation_url}
     error_doc_url:     {self.error_doc_url}
     error details:
-    {sep.join(str(error) for error in self.errors)}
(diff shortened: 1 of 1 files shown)


kiran2706:
This PR will fix this issue: https://github.com/looker-open-source/sdk-codegen/issues/1454

PR #1597, opened 2025-08-04
feat: support for oauth via PKCE in browser

drstrangelooker:
Support OAuth flow so that client_id and client_secret aren't required.


drstrangelooker:
@bryans99 are you a decent person to review this?

PR #1602, opened 2025-09-10
feat(go): Expose session object in Go SDK

drstrangelooker:
Currently you cannot access the auth session object passed into the Go SDK, because it's not passed in by reference. 
To align with other SDKs and also enable calling other APIs, we'll expose the session struct.

### DIFF
--- a/packages/sdk-codegen/src/go.gen.ts
+++ b/packages/sdk-codegen/src/go.gen.ts
@@ -261,7 +261,7 @@ export class GoGen extends CodeGen {
     return (
       `` +
       resultDef +
-      `${indent}err := l.session.Do(${resultPointer}, "${this.capitalize(
+      `${indent}err := l.AuthSession.Do(${resultPointer}, "${this.capitalize(
         method.httpMethod.toUpperCase()
       )}", "${this.apiPath}", ${this.httpPath(method.endpoint, request)}` +
       `${args ? ', ' + args : ''})` +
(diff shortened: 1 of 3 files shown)


packages/sdk-codegen/src/go.gen.ts
```diff
@@ -366,19 +366,23 @@ import (
...
-func NewLookerSDK(session authSessionDoer) *LookerSDK {
+func NewLookerSDK(session AuthSessionDoer) *LookerSDK {
   return &LookerSDK{
     session: session,
   }
 }
+
+func (l *LookerSDK) GetSession() AuthSessionDoer {
+  return l.session
+}
```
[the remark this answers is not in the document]
drstrangelooker: AuthSession is already used in the rtl, so not appropriate to rename it.

drstrangelooker:
This is to avoid dangerous hacks like https://github.com/googleapis/genai-toolbox/pull/1410#discussion_r2337509979 using reflect and unsafe packages to get access to private attributes.

packages/sdk-codegen/src/go.gen.ts
```diff
@@ -366,19 +366,25 @@ import (
...
 
-func NewLookerSDK(session authSessionDoer) *LookerSDK {
+func NewLookerSDK(session AuthSessionDoer) *LookerSDK {
   return &LookerSDK{
     session: session,
   }
 }
+
+// GetSession returns the underlying AuthSessionDoer object.
+// This is primarily for access to undocumented API endpoints.
```
jeremytchang: Remove comments please. We should not explicitly approve accessing undocumented API endpoints.

drstrangelooker: done

PR #1715, opened 2026-05-13
fix: fixes for python 3.14

drstrangelooker:
Stop supporting Python 3.9 since it is past end of life. Make some small changes to support Python 3.14, and make sure that tests cover Python 3.10 through Python 3.14.


kaumond:
LGTM

minhaz (approved):

